PR #685, opened 2014-06-10
Various improvements to uid management in ldapadmin

severo:
Validation of uid field:
* use regexp for Java validation
* use the same regexp for javascript validation in public "new account" page
* configure the regexp as a new parameter in config module
* new parameter in config module to specify the uid validation regexp

Auto generation of uid field:
* add a parameter to disable automatic uid generation in public "new account" page
* use this parameter to allow edition of uid field in privateui "new user" page
* accept uid as an input in POST /users/ API
* new parameter in config module to disable auto generation of uid field


fvanderbiest:
Bumping to 15.12, will be useful for Rennes.

PR #873, opened 2015-01-13
Jeonetwork jetty compatibility (icu4j-2.6.1.jar)

jusabatier:
Actually genonetwork doesn't start under Jetty 9.2

The problem is that the icu4j.jar seems to have compatibility problems with some JVMs

In my case, I corrected it by adding an exclusion in geonetwork/web/pom.xml near the line 266 : 

\<dependency> 
  \<groupId>jaxen\</groupId> 
  \<artifactId>jaxen\</artifactId> 
  \<exclusions> 
    \<exclusion> 
      \<groupId>com.ibm.icu\</groupId> 
      \<artifactId>icu4j\</artifactId> 
    \</exclusion> 
  \</exclusions> 
\</dependency> 

After that Geonetwork start again and everything looks OK

I read on stackoverflow that another way to solve it could be to get a more recent version of this jar but not tested yet.


fvanderbiest:
wrong target branch, closing.

PR #908, opened 2015-02-06
QuickSearch Addon

fvanderbiest:
This addon is brought by Le Puy en Velay. 
It allows one to query layer objects across different services using "typeahead" search.


fphg:
Hi
I understand the purpose, this is a nice addon. However it's a (more configurable) rewrite of the "referential search" feature. Is it a problem ?

fvanderbiest:
> However it's a (more configurable) rewrite of the "referential search" feature.

True.

> Is it a problem ?

You mean: we now have two ways in mapfishapp to do the same thing:
 - one, which is mandatory, which is targeted at people with a mouse
 - an other one, optional, which is targeted at people with a keyboard and a good knowledge of the tool.

I don't really see a problem.
Of course, we could make all tabs from the tabpanel in the bottom right corner of the screen optional, by "addonizing" them, but it would be another task.

fphg:
OK thanks

fvanderbiest:
No negative feedback got from Le Puy en Velay => merging

PR #913, opened 2015-02-12
Fix missing session's attributes in LDAPAdmin forms [TC7]

jusabatier:
When I try to submit new account creation form in LDAPadmin, I get this exception : 

org.springframework.web.HttpSessionRequiredException: Expected session attribute 'accountFormBean'

I found how to solve it in this thread : http://stackoverflow.com/questions/17737913/http-status-500-expected-session-attribute-userobject

The solution is to add a ModelAttribute method at the end of the NewAccountFormController class, cf. changes in file.

I don't know why I'm the only one to get this error and why this block of code correct it...

### DIFF
--- a/ldapadmin/src/main/java/org/georchestra/ldapadmin/ws/changepassword/ChangePasswordFormController.java
+++ b/ldapadmin/src/main/java/org/georchestra/ldapadmin/ws/changepassword/ChangePasswordFormController.java
@@ -135,7 +135,9 @@ public class ChangePasswordFormController {
 
 		}
 	}
-
-
-
+	
+	@ModelAttribute("changePasswordFormBean")
+	public ChangePasswordFormBean getChangePasswordFormBean() {
+		return new ChangePasswordFormBean();
+	}
 }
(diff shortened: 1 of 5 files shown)


fvanderbiest:
This issue seems specific to the use of tomcat7, and can be reproduced using the docker stack when submitting the "new user form".

fvanderbiest:
Thanks a lot for your contribution Julien !
(and thanks also Pierre for the review)

pmauduit:
In fact, I wonder if we don't have a deeper issue here ; in a vagrant with tomcat8, I stumbled upon an issue with this code: the beans are now recreated if Spring could not manage to find it back into the session store.

Maybe we encounter a session renewal at the security-proxy (reminding me an issue with tomcat and geoserver ...) where it should not happen. Anyway I could reproduce on my vagrant but not onto a freshly downloaded tomcat8 ...

pmauduit:
Ok, the code introduced by the PR should not be useful anymore with the following 2 commits:

https://github.com/pmauduit/georchestra/commit/3ad0ecb5529f3ff754f9e44b66e9572f95e6217d
https://github.com/pmauduit/georchestra/commit/ee0174fe50c38e3f7d726dac4ebc482b1dc4c558

The problem encountered by @jusabatier was that since the SP was not capable of correctly tracking sessions, the objects serialized by ldapadmin into the user's session were lost between the GET and the POST onto ldapadmin's interfaces.

Anyway, it is no harm to keep it that way.

PR #921, opened 2015-02-20
[Mapfish] Enable CQL Filter by JSON POST

jusabatier:
This patch enable to send POST JSON data to mapfihapp with a cql_filter parameter.

Tested with : 

```
{
    "layers": [{
        "layername": "geo_commune",
        "owstype": "WFS",
        "owsurl": "https://opendata.agglo-lepuyenvelay.fr/geoserver/cadastre/wfs",
		"cql_filter": "idu = 157"
    }, {
        "layername": "lepuy:secteurs_elus",
        "owstype": "WMS",
        "owsurl": "https://opendata.agglo-lepuyenvelay.fr/geoserver/wms",
		"cql_filter": "numsecteur = 8"
    }]
}
```

When I tested it, I can't load WMS and WFS servers simultaneously, so I added a condition over the layer Type in updateStoreFromWxSLayer().

This allow the javascipt to not crash. There still a lot of bugs with WFS protocol, but not related to this patch...

### DIFF
--- a/mapfishapp/src/main/webapp/app/js/GEOR_mapinit.js
+++ b/mapfishapp/src/main/webapp/app/js/GEOR_mapinit.js
@@ -22,6 +22,7 @@
  * @include OpenLayers/Projection.js
  * @include OpenLayers/Format/JSON.js
  * @include OpenLayers/Format/GeoJSON.js
+ * @include OpenLayers/Format/CQL.js
  * @include GeoExt/data/LayerRecord.js
  * @include GeoExt/data/LayerStore.js
  * @include GeoExt/data/WMSCapabilitiesReader.js
--- a/mapfishapp/src/main/webapp/app/js/GEOR_mapinit.js
+++ b/mapfishapp/src/main/webapp/app/js/GEOR_mapinit.js
@@ -318,6 +318,18 @@ GEOR.mapinit = (function() {
                     return (r.get('name') == item.name);
                 }).first();
                 if (record) {
+                    // handle cql_filter param in JSON POST
+                    if( type == "WFS" ) {
+                        if ( !(typeof item.cql_filter === 'undefined') ) {
+                            var format = new OpenLayers.Format.CQL();
+                            record.data.layer.filter = format.read(item.cql_filter);
+                        }
+                    } else {
+                        if ( !(typeof item.cql_filter === 'undefined') ) {
+                            record.data.layer.params.CQL_FILTER = item.cql_filter;
+                        }
+                    }
+                    
                     // set metadataURLs in record, data comes from GeoNetwork
                     if (item.metadataURL) {
                         record.set("metadataURLs", [item.metadataURL]);
--- a/mapfishapp/src/main/webapp/app/js/GEOR_mapinit.js
+++ b/mapfishapp/src/main/webapp/app/js/GEOR_mapinit.js
@@ -320,13 +320,13 @@ GEOR.mapinit = (function() {
                 if (record) {
                     // handle cql_filter param in JSON POST
                     if( type == "WFS" ) {
-                        if ( !(typeof item.cql_filter === 'undefined') ) {
+                        if ( item.hasOwnProperty("cql_filter") ) {
                             var format = new OpenLayers.Format.CQL();
-                            record.data.layer.filter = format.read(item.cql_filter);
+                            record.getLayer().filter = format.read(item.cql_filter);
                         }
                     } else {
-                        if ( !(typeof item.cql_filter === 'undefined') ) {
-                            record.data.layer.params.CQL_FILTER = item.cql_filter;
+                        if ( item.hasOwnProperty("cql_filter") ) {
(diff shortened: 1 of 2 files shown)


mapfishapp/src/main/webapp/app/js/GEOR_mapinit.js
```diff
@@ -313,11 +313,23 @@ GEOR.mapinit = (function() {
         var records = [], record;
         var errors = [], count = 0;
         Ext.each(initState, function(item) {
-            if (item.type == "WMSLayer" || item.type == "WFSLayer") {
+            if ( (item.type == "WMSLayer" || item.type == "WFSLayer") && item.type == type+'Layer' ) {
```
fvanderbiest: I don't understand why you had to change this

mapfishapp/src/main/webapp/app/js/GEOR_mapinit.js
```diff
@@ -313,11 +313,23 @@ GEOR.mapinit = (function() {
...
                 record = stores[item.url].queryBy(function(r) {
                     return (r.get('name') == item.name);
                 }).first();
                 if (record) {
+                    // handle cql_filter param in JSON POST
+                    if( type == "WFS" ) {
+                        if ( !(typeof item.cql_filter === 'undefined') ) {
```
fvanderbiest: I'd prefer:
```
if (item.hasOwnProperty("cql_filter")) {
```

mapfishapp/src/main/webapp/app/js/GEOR_mapinit.js
```diff
@@ -313,11 +313,23 @@ GEOR.mapinit = (function() {
...
+                            var format = new OpenLayers.Format.CQL();
+                            record.data.layer.filter = format.read(item.cql_filter);
```
fvanderbiest: ```record.data.layer``` -> ```record.getLayer()```

fvanderbiest:
Thanks for the contribution !

The mapfishapp readme should also be edited to explain the new abilities.

mapfishapp/src/main/webapp/app/js/GEOR_mapinit.js
fvanderbiest: OK, sorry, I did no see "When I tested it, I can't load WMS and WFS servers simultaneously, so I added a condition over the layer Type in updateStoreFromWxSLayer()."

fvanderbiest: Looks good then

mapfishapp/src/main/webapp/app/js/GEOR_mapinit.js
```diff
@@ -313,11 +313,23 @@ GEOR.mapinit = (function() {
...
-            if (item.type == "WMSLayer" || item.type == "WFSLayer") {
+            if ( (item.type == "WMSLayer" || item.type == "WFSLayer") && item.type == type+'Layer' ) {
                 record = stores[item.url].queryBy(function(r) {
                     return (r.get('name') == item.name);
                 }).first();
                 if (record) {
+                    // handle cql_filter param in JSON POST
+                    if( type == "WFS" ) {
+                        if ( item.hasOwnProperty("cql_filter") ) {
+                            var format = new OpenLayers.Format.CQL();
```
fvanderbiest: you probably miss the build dependency to OpenLayers/Format/CQL.js in the file header too

mapfishapp/src/main/webapp/app/js/GEOR_mapinit.js
```diff
@@ -313,11 +313,23 @@ GEOR.mapinit = (function() {
...
+                            record.getLayer().filter = format.read(item.cql_filter);
```
fvanderbiest: record.getLayer().filter = (new OpenLayers.Format.CQL()).read(item.cql_filter); would be more concise

fvanderbiest:
And I forgot to tell you that, since it is a new feature, it should be pushed to master, not stable 14.12.
But since this is a relatively harmless patch, OK for this time ...

fvanderbiest:
Question: does the filter persist after the map context is restored (either via permalink or WMC) ?

jusabatier:
Can't remember what name we have to set for the JSON variable in the POST request.

I'll test it when I get the name to use, and think it could be a good thing to add it in the doc.

fvanderbiest:
> Can't remember what name we have to set for the JSON variable in the POST request.

data

jusabatier:
It's appear that for WMS, the modifications don't persist with permalink (so wmc) because it use the filter attribute which is not "standard" of WMS protocol. This need an update and we can reuse the code of my PR for filter which create an altered SLD and load it (like the styler do). Use this method allow to persist the changes for WMS.

Can't test it for WFS because I can't load my layer, maybe du to the problems I metionned in my previous posts in this thread, don't remember. I'll look at that when I'll have some times. I think it could persist with WFS because FILTERS is a standard attribute of WFS protocol.

I'll try to work on those improvment next week.

Issue #40, opened 2013-01-30
mapfishapp: CONTEXTS array does not need to have title and description


fvanderbiest:
A WMC already has those fields (id & abstract)... we could just parse them.

fvanderbiest:
@pmauduit has an interesting proposal in a working branch for this one.

The keywords could also be inferred from the WMC file.

pmauduit:
some work can be integrated prematurely in master from this branch: https://github.com/pmauduit/georchestra/tree/mfapp-datadir

fphg:
thanks
How many WMC will it support ? Impact on performances ?
Note:
- thumbnails are costly, title&abstract aren't, we won't spare much
- if WMC are stored in the mapfishapp database, title & abstract can be easily retrieved serverside

pmauduit:
I tried on PIGMA WMC base (around ~30 WMCs) and performance / time was still not an issue

pmauduit:
@fpgh some metrics:

```
ab -n 500 -c 50 http://localhost:8080/mapfishapp/ws/contexts
This is ApacheBench, Version 2.3 <$Revision: 1604373 $>
Copyright 1996 Adam Twiss, Zeus Technology Ltd, http://www.zeustech.net/
Licensed to The Apache Software Foundation, http://www.apache.org/

Benchmarking localhost (be patient)
Completed 100 requests
Completed 200 requests
Completed 300 requests
Completed 400 requests
Completed 500 requests
Finished 500 requests


Server Software:        Jetty(6.1.26)
Server Hostname:        localhost
Server Port:            8080

Document Path:          /mapfishapp/ws/contexts
Document Length:        6236 bytes

Concurrency Level:      50
Time taken for tests:   16.749 seconds
Complete requests:      500
Failed requests:        0
Total transferred:      3176500 bytes
HTML transferred:       3118000 bytes
Requests per second:    29.85 [#/sec] (mean)
Time per request:       1674.944 [ms] (mean)
Time per request:       33.499 [ms] (mean, across all concurrent requests)
Transfer rate:          185.20 [Kbytes/sec] received

Connection Times (ms)
              min  mean[+/-sd] median   max
Connect:        0    0   0.4      0       2
Processing:   716 1641 342.0   1617    2647
Waiting:      716 1640 342.0   1617    2647
Total:        716 1642 342.2   1617    2648
```

fvanderbiest:
Closing, since it will be addressed by the so-called mapfishapp "data dir".

PR #943, opened 2015-03-12
mapfishapp - dynamic WMC controller

pmauduit:
This is kind of related to issue #40 : We set a large json array in
GEOR_Custom.js while this could be calculated dynamically by a
webservice.

This commit is a backport of a current work which aims to export
configuration of some aspects outside of the webapp. Since this can be
easily extracted and proposed onto master earlier than planned, let's go
for it.

Tests: unit tests backported, runtime tests run.

TODO: we might want to return the default array currently provided by
the configuration instead of an empty JSON array.

Note: since this is not currently wired up to the interface, merging it
into master is pretty harmless.


fvanderbiest:
Closing, this will be done after 15.06 with mapfishapp "data dir".

PR #957, opened 2015-04-13
basic-auth filter enhancement

pmauduit:
We need to avoid in certain cases to trigger the filter ; currently,
the géocatalogue is triggering this filter, because using a user-agent
activated by default (Jakarta, also used by GeoFence).

This commit provides a basic way of disabling the filter for certain
hosts. It makes use of the "x-forwarded-for" HTTP header.

Related issue:
https://github.com/camptocamp/georchestra-cigalsace-configuration/issues/482


fvanderbiest:
As discussed, the correct approach would rather be https://github.com/georchestra/georchestra/issues/960

Closing.

PR #962, opened 2015-04-21
security-proxy - Fix tomcat7 incompatibilities

pmauduit:
This is a backport of my current branch (georchestra_datadir), which
allows the use of tomcat > 6 for the security-proxy.

Main issues were the following:
- Redirect loops due to the Tuckeys' urlrewrite configuration ;
- useless default servlet mapping which was messing up somehow with
tomcat default servlet configuration

Also seems to fix a bug with jsp compilation.

Tests:
- Runtime tested on tomcat 6 and tomcat 7, no regression encountered
- Unit tests OK.

### DIFF
--- a/security-proxy/pom.xml
+++ b/security-proxy/pom.xml
@@ -135,7 +135,6 @@
 			<groupId>javax.servlet</groupId>
 			<artifactId>jstl</artifactId>
 			<version>1.1.2</version>
-			<scope>compile</scope>
 		</dependency>
 		<dependency>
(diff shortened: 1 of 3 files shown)


fvanderbiest:
Thanks Pierre.

In the proxy, @severo had setup customizable 403 and 404 pages, see eg ```config/defaults/security-proxy/403.jsp```. These pages use static resources available under ```/_static/```.

Are these pages still fully functional ?

pmauduit:
> Are these pages still fully functional ?

Yep

Issue #961, opened 2015-04-21
security-proxy - bad HTTP response in case of downstream "connection refused"


pmauduit:
The security-proxy currently returns a "200 - OK" status with a blank content in case of unavailable downstream tomcats.

I think a "502 - Bad gateway" or a "503 - Service unavailable" should be more relevant.

Step to "reproduce":

- Launch the proxycas tomcat
- shut down the georchestra tomcat
- attempt to connect onto `http[s]://host/mapfishapp/`

fvanderbiest:
+1, this would maximize the benefit of having the ErrorDocument setup as per https://github.com/georchestra/georchestra/blob/14.12/doc/setup/apache.md#configuration 

These error documents are currently shown (only) when the tomcat-proxycas is unavailable.

PR #963, opened 2015-04-21
security-proxy - Intercept "connection refused" errors

pmauduit:
This PR is related to the issue #961.

If underlying webapps are unavailable, an empty webpage is produced
(Content-length: 0), no log traces are produced (a printStackTrace() is
done anyway), and a 200 OK status code is returned.

This commit provides:
- A 503 "service unavailable" status code is returned instead
- A default status page for 503 (as it is currently done for 403 and 404
  statuses) is produced,
- A log trace is correctly (in the correct log file, not on stdout) is
  printed.

Note that the 503 error code can also be intercepted by the frontend
server, so that a custom error page is returned.

### DIFF
--- a/security-proxy/src/main/java/org/georchestra/security/Proxy.java
+++ b/security-proxy/src/main/java/org/georchestra/security/Proxy.java
@@ -612,7 +612,13 @@ public class Proxy {
             }
         } catch (IOException e) {
             // connection problem with the host
-            e.printStackTrace();
+            logger.error("Exception occured when trying to connect to the remote host: ", e);
+            try {
+                finalResponse.sendError(HttpServletResponse.SC_SERVICE_UNAVAILABLE);
+            } catch (IOException e2) {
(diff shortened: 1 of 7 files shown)


security-proxy/src/main/java/org/georchestra/security/Proxy.java
```diff
@@ -612,7 +612,13 @@ private void handleRequest(HttpServletRequest request, HttpServletResponse final
             }
         } catch (IOException e) {
             // connection problem with the host
-            e.printStackTrace();
+            logger.error("Exception occured when trying to connect onto the remote host: ", e);
```
fvanderbiest: "connect to" ?

fvanderbiest:
Looks good !

security-proxy/src/main/java/org/georchestra/security/Proxy.java
pmauduit: done

PR #967, opened 2015-04-30
Add filter button to querier

jusabatier:
Add a filter button to the querier in order to add CQL filters to the layer and refresh it.

Actually doesn't work with geometry because filter returned reference a null geometry attribute : 

```
var filterbuilder = this.findParentByType("gx_filterbuilder");
var filter = filterbuilder.getFilter();
```

### DIFF
--- a/mapfishapp/src/main/webapp/app/js/GEOR_querier.js
+++ b/mapfishapp/src/main/webapp/app/js/GEOR_querier.js
@@ -29,6 +29,7 @@
  * @requires OpenLayers/Filter/Comparison.js
  * @requires OpenLayers/Filter/Spatial.js
  * @include OpenLayers/Filter/Logical.js
+ * @include OpenLayers/Format/CQL.js
  * @include OpenLayers/Format/Filter.js
  * @include OpenLayers/Format/XML.js
  // see https://github.com/georchestra/georchestra/issues/482 :
@@ -258,10 +259,16 @@ GEOR.querier = (function() {
             deactivable: true,
             cookieProvider: cp,
             autoScroll: true,
-            buttons: [{
-                text: tr("Search"),
-                handler: search
-            }],
+            buttons: [
+                {
+                    text: tr("Search"),
+                    handler: search
+                },
+                {
+                    text: tr("Filter"),
+                    handler: filterLayer
+                }
+            ],
             map: map,
             attributes: attStore,
             allowSpatial: true,
@@ -271,6 +278,24 @@ GEOR.querier = (function() {
             })
         });
     };
+    
+    var filterLayer = function() {
+        var filterbuilder = this.findParentByType("gx_filterbuilder");
+        var filter = filterbuilder.getFilter();
+        if (!checkFilter(filter)) {
+            return;
+        }
+        
+        var layers = map.getLayersBy('id',record.id);
+        
+        if( layers.length > 0 ) {
+            if( filter.CLASS_NAME == "OpenLayers.Filter.Spatial" || filter.CLASS_NAME == "OpenLayers.Filter.Comparison" || ('filters' in filter && filter.filters.length > 0) ) {
+                layers[0].mergeNewParams({'CQL_FILTER': (new OpenLayers.Format.CQL()).write(filter)});
+            } else {
+                layers[0].mergeNewParams({'CQL_FILTER': null});
+            }
+        }
+    }
 
     return {
     
--- a/mapfishapp/src/main/webapp/app/js/GEOR_querier.js
+++ b/mapfishapp/src/main/webapp/app/js/GEOR_querier.js
@@ -29,7 +29,6 @@
  * @requires OpenLayers/Filter/Comparison.js
  * @requires OpenLayers/Filter/Spatial.js
  * @include OpenLayers/Filter/Logical.js
- * @include OpenLayers/Format/CQL.js
  * @include OpenLayers/Format/Filter.js
  * @include OpenLayers/Format/XML.js
  // see https://github.com/georchestra/georchestra/issues/482 :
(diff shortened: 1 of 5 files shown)


fvanderbiest:
CQL_FILTER is geoserver-specific.
IMO, this PR is not compatible with geOrchestra goals.

fphg:
CQL is a valid OGC language but a vendor option for WMS. fvanderbiest is right : 

* if a vendor is used, the viewer shall not let the user use it without assurance that it is supported by the entry point
* if there's another, non vendor way to do it, it has to be tested first

jusabatier:
I rewrote the function to use OGC's XML filters for WMS and WFS layers.

But for WMS there still some bug with the geom property which stay to null in the filters...

mapfishapp/src/main/webapp/app/js/GEOR_querier.js
```diff
@@ -271,6 +277,33 @@ GEOR.querier = (function() {
...
     };
+    
+    var filterLayer = function() {
+        var filterbuilder = this.findParentByType("gx_filterbuilder");
+        var filter = filterbuilder.getFilter();
+        if (!checkFilter(filter)) {
+            return;
+        }
+        
+        var layers = map.getLayersBy('id',record.id);
```
fvanderbiest: var layer = record.getLayer()

mapfishapp/src/main/webapp/app/js/GEOR_querier.js
```diff
@@ -271,6 +277,33 @@ GEOR.querier = (function() {
...
+        
+        if( layers.length > 0 ) {
+            if( layers[0].CLASS_NAME == "OpenLayers.Layer.WMS" ) {
+                if( filter.CLASS_NAME == "OpenLayers.Filter.Spatial" || filter.CLASS_NAME == "OpenLayers.Filter.Comparison" || ('filters' in filter && filter.filters.length > 0) ) {
+                    layers[0].params["FILTER"] = (new OpenLayers.Format.XML()).write((new OpenLayers.Format.Filter()).write(filter));
```
fvanderbiest: It should be simpler to serialize an OL filter into FE xml

mapfishapp/src/main/webapp/app/js/GEOR_querier.js
```diff
@@ -271,6 +277,33 @@ GEOR.querier = (function() {
...
+        var filter = filterbuilder.getFilter();
+        if (!checkFilter(filter)) {
+            return;
+        }
+        
+        var layers = map.getLayersBy('id',record.id);
+        
+        if( layers.length > 0 ) {
+            if( layers[0].CLASS_NAME == "OpenLayers.Layer.WMS" ) {
+                if( filter.CLASS_NAME == "OpenLayers.Filter.Spatial" || filter.CLASS_NAME == "OpenLayers.Filter.Comparison" || ('filters' in filter && filter.filters.length > 0) ) {
```
fvanderbiest: What are these checks for ?

jusabatier:
```
It should be simpler to serialize an OL filter into FE xml
```

What is "FE xml" ?

```
filter.CLASS_NAME == "OpenLayers.Filter.Spatial" || filter.CLASS_NAME == "OpenLayers.Filter.Comparison" || ('filters' in filter && filter.filters.length > 0)
```

Those tests are used to check if the filter list is not empty, if the user select just one logical or spatial filter we get the filter, if he select more than one, we get an array. If it's empty we reset the filter list and reload all features of the map.

jusabatier:
```
It should be simpler to serialize an OL filter into FE xml
```

I don't think, the OL example use exactly the same way to encode filter : 
http://dev.openlayers.org/examples/filter.html

```
(new OpenLayers.Format.XML()).write((new OpenLayers.Format.Filter()).write(filter));
```

Don't think there is a simpler way to do that...

jusabatier:
It have been reported that filters are not in WMS standars, so we need to add them in SLD for WMS layers.

I can do that, but It implies to set the non-matching features opacity to 0, so they still exist on the layer and are queriable... 

And for layers with huge set of features, it will take a longer time to render those transparent features, am I wrong ?

Or maybe there is an SLD option to say to the server to skip them during the render process ?

One of the interest of filtering features is to enable layers with huge amount of features (cadastre parcels for example) to load faster. If we filter by SLD, it's not faster because all features need to be proceced.

Please give me your opinions and suggestions, this option to filter features is a request of my users and I can work on it.

fphg:
Hello jusabatier

To filter features you should add your custom filters for each FE ruleset. The rendering engine draws features matching the ruleset and discards the other ones. Please build a custom, attribute dependent style with mapfishapp styler and download the style document, you'll notice the filters. Or have a look at http://docs.geoserver.org/stable/en/user/styling/sld-reference/rules.html, bottom of the page.

jusabatier:
OK, so if I understand correctly, if no rule are defined for a feature, it's not rendered.

No way to say "If not (condition) no render" ? Because adding a rule which do that to an existing SLD is possible, but preserve existing rules which match the filter conditions and remove others will be much more complex...

It will need some comparison between CQL filters and existing SLD rules to determine which ones correspond to the filter.

fphg:
not exactly :)
You have to define a Rule, but, depending on the optional rule filter, you will or won't render a particular feature.

Say you want to render only features associated to a specified zipcode. This :

```xml
   <PropertyName>zipcode</PropertyName>
   <Literal>44444</Literal>
</PropertyIsEqualTo>
```
Will render features with zipcode===44444 and discard the other ones.

Works exactly like CQL.

fphg:
Didn't read everything, sorry
You have to merge the current filter block and your specific ones into a <ogc:And> expression.

jusabatier:
Yes, that's a good solution :)

Thanks for the Idea I'll try it.

jusabatier:
New PR : https://github.com/georchestra/georchestra/pull/1003

You can close this one

PR #970, opened 2015-05-06
[Mapfish] Correct addon clear Tab2 bug

jusabatier:


fvanderbiest:
2 times ```this.fields = fields;``` ? 

afterrender changed into beforeshow ?

This deserves more explanations. Thanks

jusabatier:
I changed afterrender to beforeshow because beforeshow is called each time the TabPanel is show before show it and we need to reassign the fields each time we change the shown TabPan (switch between reference and owner search). Afterrender is only called at the TabPan creation, that's why actually the clean button doesn't work for owner tab.

We also need to define this handler for each of the two TabPan, that's why there is 2 time the assignation of this.fields

If you want I can add a function but I think it's not very useful for 2 lines of codes

fvanderbiest:
There's something weird in the PR: setting this.fields to fields in the callbacks

jusabatier:
Why weird ? Isn't this.fileds supposed to contain the current showed fileds ? That's why we need to update it when we change the active TabPan.

mapfishapp/src/main/webapp/app/addons/cadastre/js/main.js
```diff
@@ -543,8 +544,9 @@ GEOR.Addons.Cadastre = Ext.extend(GEOR.Addons.Base, {
             title: OpenLayers.i18n("tab2title"),
             items: fields,
             listeners: {
-                "afterrender": function(form) {
-                    fields[0].focus('', 50);
```
jusabatier: This is weird because actually you focus the Tab1 first field when Tab2 is created

fvanderbiest:
Old code, but AFAIK, this.fields only contains the first tab fields.
You might have to create and handle this.tab1Fields and this.tab2Fields in your PR.

jusabatier:
So tab1Fields will be a duplicate of this.fields ?

fvanderbiest:
Not a duplicate. 
this.tab1Fields should replace this.fields

jusabatier:
Ok, but a variable which contains current displayed fields allow to Clean only the displayed fields and preserve the hidden ones, isn't it interesting ?

Issue #975, opened 2015-05-12
extractor - WCS extraction requests could fail if layer name is prefixed 


pmauduit:
In some cases, the layer name in the extraction request is prefixed. We should be a little bit fuzzier in the way the layers are detected in the `GetCapabilities` response, so that the checkPermission() call does not fail when the layer is actually present (in its non-prefixed form).

fphg:
+1

pmauduit:
Merged

PR #976, opened 2015-05-12
extractorapp - Fuzziest check on layer name

pmauduit:
Depending on which GetCapabilities is made onto the GeoServer endpoint,
the layer names in the WCS GetCapabilities document can be or not
prefixed.

this patch proposes to be fuzzier on the layer name check, removing the
prefix.

Related Github issue:
https://github.com/georchestra/georchestra/issues/975

Tests: unit tests, not tested at runtime.

Bug: I consider checking the layer existence using regex onto XML
documents as a bug, it would be relevant imho to rewrite this part (not
to mention that this is kind of duplicated with the check present in the
WfsExtractor). The main reason it has been made this way is that some
implementations returned badly-formatted XML documents.


fvanderbiest:
LGTM, please merge.

PR #977, opened 2015-05-12
[Mapfish] Fix bug on editing feature

jusabatier:
When editing a feature (observed on close lines and points), sometimes the hovered feature is not the good one.

This problem is due to : 
```
maxFeatures: 1,
```

Because the wfs protocol requests a BBOX where more than one feature can be included, and as said in the OL doc, set maxFeatures to more than one allows wfs to return more than one feature. The returned set of features is then used to determine the best match client-side.

http://dev.openlayers.org/apidocs/files/OpenLayers/Control/GetFeature-js.html

I remove the line because by default, maxFeatures is set to 10, the good value to set there depends on the distance between features in the edited layer, so the default value seems to be a good average.

After doing that, the edit function is repaired.

### DIFF
--- a/mapfishapp/src/main/webapp/app/js/GEOR_edit.js
+++ b/mapfishapp/src/main/webapp/app/js/GEOR_edit.js
@@ -139,7 +139,6 @@ GEOR.edit = (function() {
                 hover: true,
                 click: false,
                 single: true,
-                maxFeatures: 1,
                 clickTolerance: 7,
                 eventListeners: {
                     "hoverfeature": function(e) {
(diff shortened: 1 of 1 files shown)


fvanderbiest:
Thanks a lot for spotting this !

I'm +1 on changing this value, but having a maximum of ten features returned by the service might hang the client in case of big, complex geometries. I would like this number to be as low as possible.

Is your issue gone with, eg maxFeatures = 2 or 3 ?

jusabatier:
As far as I tested, 3 should be suficcient for MY case, but as I said if some user have a layer with more close features, it could need more. I think a value of 5 could be a good compromise.

Moreover, if you set 5 and the server found only 3 in the BBOX, it will return only the 3 I think... So I don't think there is loss of perfs with a bigger value.

fvanderbiest:
ok, sounds good.

PR #980, opened 2015-05-20
Update postgresql.md

MaxiReglisse:
i don't know if libpostgis-java is useful or not...
but postgis and postgresql-postgis are required !

### DIFF
--- a/doc/setup/postgresql.md
+++ b/doc/setup/postgresql.md
@@ -1,3 +1,8 @@
+# Prerequisites
+```
+apt-get install postgis postgresql-9.1-postgis libpostgis-java
+```
+
 # Setting up the geOrchestra database
 
 The "georchestra" database hosts several schemas, which are specific to the deployed modules:


MaxiReglisse:
oh, i forgot "sudo" !!

pmauduit:
> i don't know if libpostgis-java is useful or not...

The package provides some jars in the debian own local maven-repo (`/usr/share/maven-repo`)  as well as the postgis jdbc driver into `usr/share/java`.

Since these jars are usually provided by the webapps I don't think it is useful to also install the package.

fvanderbiest:
-1 on this change since it is already mentioned in https://github.com/MaxiReglisse/georchestra/blob/14.12/doc/setup.md

MaxiReglisse:
sorry, i was focused on https://github.com/MaxiReglisse/georchestra/blob/14.12/doc/setup/postgresql.md...

PR #994, opened 2015-06-11
Tomcat7 fixes

pmauduit:
This is a cherry-pick of some commits from my georchestra-datadir branch which makes the security-proxy / georchestra compatible with newer versions of tomcat.

Not to be merged immediately, because it has not been validated on tomcat7 and 6 (regressions) yet.


pmauduit:
Tested in tomcat8: OK

pmauduit:
No regression on tomca6. It can be merged with no harm from my point of view.

PR #996, opened 2015-06-17
Upgrade Spring to 3.1.0 for catalogapp, dowloadform, extractorapp and header

jusabatier:
I can see with TC7 that the current version of Spring leads to some errors, for example on header : 

```
PageNotFound [WARN] No mapping found for HTTP request with URI [/header/ws/home] in DispatcherServlet with name 'ws'
```

So pmauduit explain me that I should upgrade to a newer version as he did on his branch : 

https://github.com/pmauduit/georchestra/commit/1e8c020e8750b747aa2a9bb429284ce82bf23bb4

What I did in this PR and after correct it everything is OK

### DIFF
--- a/catalogapp/pom.xml
+++ b/catalogapp/pom.xml
@@ -12,7 +12,7 @@
 	<url>http://www.georchestra.org</url>
 	<properties>
 		<finalName>catalogapp</finalName>
-		<java-version>1.6</java-version>
+		<spring.version>3.1.0.RELEASE</spring.version>
 		<org.slf4j-version>1.6.1</org.slf4j-version>
 		<force.sdk.version>22.0.3-BETA</force.sdk.version>
 	</properties>
@@ -31,7 +31,7 @@
 		<dependency>
 			<groupId>org.springframework</groupId>
 			<artifactId>spring-webmvc</artifactId>
-			<version>2.5.6.SEC01</version>
+			<version>${spring.version}</version>
 			<scope>compile</scope>
 		</dependency>
 		<dependency>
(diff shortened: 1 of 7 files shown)


pmauduit:
Note: there also is a spring.version property defined at root pom.xml level:
https://github.com/georchestra/georchestra/blob/14.12/pom.xml#L15

But I don't mind having a custom version in each module (we could imagine the need of a custom spring version for a given webapp)

pmauduit:
Runtime regression tests (jetty + tomcat6/java6):

- header: OK
- downloadform: NOK:

```
org.springframework.beans.factory.xml.XmlBeanDefinitionStoreException: Line 34 in XML document from ServletContext resource [/WEB-INF/ws-servlet.xml] is invalid; nested exception is org.xml.sax.SAXParseException; lineNumber: 34; columnNumber: 161; The prefix "p" for attribute "p:locations" associated with an element type "bean" is not bound.
```
Please add the following line in ws-servlet.xml for downloadform to fix it:
```
diff --git a/downloadform/src/main/webapp/WEB-INF/ws-servlet.xml b/downloadform/src/main/webapp/WEB-INF/ws-servlet.xml
index 1a02fc9..91a49ed 100644
--- a/downloadform/src/main/webapp/WEB-INF/ws-servlet.xml
+++ b/downloadform/src/main/webapp/WEB-INF/ws-servlet.xml
@@ -2,6 +2,7 @@
 <beans xmlns="http://www.springframework.org/schema/beans"
     xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
     xmlns:context="http://www.springframework.org/schema/context"
+    xmlns:p="http://www.springframework.org/schema/p"
     xsi:schemaLocation="http://www.springframework.org/schema/beans
     http://www.springframework.org/schema/beans/spring-beans-3.0.xsd
     http://www.springframework.org/schema/context

```
- extractorapp: OK

- catalogapp: OK

pmauduit:
Otherwise, looks good to me

pmauduit:
Closing because new one proposed here (rebased on master): https://github.com/georchestra/georchestra/pull/997

PR #997, opened 2015-06-18
Patch spring update

pmauduit:
following https://github.com/georchestra/georchestra/pull/996 (rebased on master)


pmauduit:
I rebased my georchestra-datadir branche so that the debian packages will automatically regenerated, would be way easier to test the PR at runtime

pmauduit:
Ok it works (untested for downloadform outside of jetty/java6), but the whole stack works with these modifications in my vm (branche georchestra-datadir).

I'm pretty confident in merging it. thanks @jusabatier

PR #1003, opened 2015-07-06
Mapfishapp filter in querier

jusabatier:
Adding a "Filter" button to hide features which don't match the rules.

Work with FILTERS attributes and CQL filters for WFS layers,
and for WMS ones by generate an customized SLD from the server availables styles with rules added to it.

So it matches the WMS standards.

In order to generate custom SLDs, I reused the styler code to save modified ones.

I commented some points that can be done better but actually this version works. I tested it only with the integrated GS, good thing if it can be tested with MS to see if the GetStyles return the same SLD format as GS.

Waiting your suggestion to optimize it ;)


jusabatier:
Any suggestions / corrections on this feature ?

Is the commit "Merge pull request #5 from georchestra/master" useless ?

If yes is there a way to remove this without remove the other ?

fvanderbiest:
This branch should be rebased onto master...

fvanderbiest:
mapfishapp EOL. Closing.

PR #1025, opened 2015-08-07
security-proxy/health-checker: make database host and port configurable

landryb:
Not tested yet in production, but i saw it as a //FIXME, and hardcoding localhost:5432 is so '90...
Not really critical since healthchecker is not enabled by default, and i still fail to see what's the usecase for it....

Based on what was done in https://github.com/georchestra/georchestra/commit/4c0ff764ad2a7c39477c9451eb85eadd0da710e4


fvanderbiest:
Looks good, thanks Landry !

PR #1030, opened 2015-08-13
Documentation for Monit

jusabatier:


fvanderbiest:
Looks good now, thanks a lot !

landryb:
Two minor ideas/nits:
- in the monit config, ```set mmonit http://monit:kp12trA0#xH74@<mmonit-server-IP>:8080/collector``` seems to contain a 'real' password - wouldnt it be better to have a simple password like ```mypassword``` to make it obvious one has to change it ?
- requesting the full /mapfishapp, /geonetwork and /geoserver full pages with all content/images/css/etc every minute might be 'heavy' on the service which is already loaded by regular users, and fills the logs with noise/false positive.. it might be simpler to request a page known for returning a very small amount of content, still ensuring the service works by calling an url actually handled by the webapp - like it's done for geoserver using describelayer...

for example, i use this for cas & geonetwork:
```
  if failed port 8080 protocol http and request /cas/images/services/true.gif then alert
  if failed port 8080 protocol http and request /cas/images/services/true.gif for 3 cycles then restart
  if failed port 8080 protocol http and request /geocat/srv/fre/xml.info then alert

```

jusabatier:
```in the monit config, set mmonit http://monit:kp12trA0#xH74@<mmonit-server-IP>:8080/collector seems to contain a 'real' password```

Effectively I forget to change it in my copy/paste, thanks to say that. I'll change it in my conf and maybe should change it in the doc. I'll do a PR for that.

```requesting the full /mapfishapp, /geonetwork and /geoserver```

Right, I use your rules for the geoserver conf, but I agree it's a good idea to reuse the same on the proxy conf.

PR #1035, opened 2015-08-31
Introducing a new module commons

pmauduit:
This is meant to share some common code across modules.

It also provides a new controller to get the GEOR_Custom.js file used in some modules (mapfishapp, extractorapp, ...) directly into a custom directory (defined by the `georchestra.datadir` env variable).

There are several concepts introduced by this module (which compiles into a jar, and that should be added across the modules as a new spring bean):

- georchestra.datadir is the configuration placeholder for georchestra
- the context.datadir is constructed as {georchestra.datadir}/{modulename}/
- A properties file is sourced considering the content as UTF-8 (instead of default Java behaviour which considers .properties file as ISO-8859-1)
- since the log4j configuration does not simply allow to be configured against a config file outside of the webapp, the class is responsible of reconfiguring log4j java-side at runtime

If the setup is well (needs the georchestra.datadir env variables, plus the files explained above), the boolean activated is set to true. Otherwise, the "old" geOrchestra behaviour is kept.

### DIFF
--- a/commons/src/main/java/org/georchestra/commons/configuration/GeorchestraConfiguration.java
+++ b/commons/src/main/java/org/georchestra/commons/configuration/GeorchestraConfiguration.java
@@ -1,0 +1,151 @@
+package org.georchestra.commons.configuration;
+
+import java.io.File;
+import java.io.FileInputStream;
+import java.io.IOException;
+import java.io.InputStream;
+import java.io.InputStreamReader;
+import java.util.Properties;
+
+import javax.servlet.ServletContext;
+import javax.servlet.http.HttpServletRequest;
+import javax.servlet.http.HttpServletResponse;
+
+import org.apache.commons.io.FileUtils;
+import org.apache.commons.io.IOUtils;
+import org.apache.log4j.PropertyConfigurator;
+import org.springframework.stereotype.Controller;
+import org.springframework.web.bind.annotation.RequestMapping;
+import org.springframework.web.context.ServletContextAware;
+
+@Controller
+public class GeorchestraConfiguration implements ServletContextAware {
+
+    protected String globalDatadir;
+    protected String contextDataDir;
+
+    protected String contextName;
+    protected ServletContext ctx;
+
+    protected Properties applicationSpecificProperties = new Properties();
+
+    public String getContextDataDir() {
+        return contextDataDir;
+    }
+
+    private boolean activated = false;
+
+    public void init() {}
+
+    public GeorchestraConfiguration(String context) {
+
+        this.contextName = context;
+        globalDatadir = System.getProperty("georchestra.datadir");
+        if (globalDatadir != null) {
+            contextDataDir = new File(String.format("%s%s%s%s", globalDatadir, File.separator, context, File.separator)).getAbsolutePath();
+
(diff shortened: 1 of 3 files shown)


landryb:
Looks small enough, i think it can quickly be merged without evil side-effects and could help CI :)
It's also nice that it prioritizes log4j.properties over log4j.xml......

landryb:
Hm, thinking about it a bit more, if both log4j.xml and log4j.properties exist, the xml one will take precedence.... do we want more xml in this world ?

pmauduit:
> Hm, thinking about it a bit more, if both log4j.xml and log4j.properties exist, the xml one will take precedence.... do we want more xml in this world ?

I don't know log4j enough to know if one format is better than the other, but it looked to me at first sight that the XML was allowing more features than the properties file format one. In any way, it is just configuration files, so the administrator should have to choose xml or properties, I cannot see why we can end up with a situation in both files are present

landryb:
It's good anyway, since previously iirc some apps were expecting/configured for .properties and others for .xml, and if i get it right now we can at least 'unify' this and use the same type of config for all modules.

commons/src/main/java/org/georchestra/commons/configuration/GeorchestraConfiguration.java
```diff
@@ -0,0 +1,151 @@
...
+    private boolean activated = false;
+
+    public void init() {}
+
+    public GeorchestraConfiguration(String context) {
+
+        this.contextName = context;
+        globalDatadir = System.getProperty("georchestra.datadir");
+        if (globalDatadir != null) {
+            contextDataDir = new File(String.format("%s%s%s%s", globalDatadir, File.separator, context, File.separator)).getAbsolutePath();
```
fvanderbiest: line too long (sorry)

fvanderbiest:
IMO, this introduces technical debt, because the best way to do would be to completely remove GEOR_custom.js in favor of database storage for the client-side config options.

Considering the fact that we now have to support 2 ways of deploying webapps, this is a "best effort" implementation.

Merging... but one day, we'll have to go over this again.

landryb:
What technical debt are you referring to ? Storing configuration files in a database is a nightmare admin-wise in terms of idempotence/historization...... just look at the settings table from a geonetwork instance :)

fvanderbiest:
The "settings table" in geonetwork was indeed a very very bad idea.

The technical debt I was referring to, lies in the fact that we are keeping the same file that we use to store client-side options, when we should use (IMO) a dedicated controller that passes these options from the server to the client instead.

Storing the config options in a properties/xml file, or in a db is indeed another question.

pmauduit:
> What technical debt are you referring to ? Storing configuration files in a database is a nightmare admin-wise in terms of idempotence/historization...... just look at the settings table from a geonetwork instance :)

After having discussed with Francois, having a js file might not be a good way to go ; from my PoV it is more of a Client-side consideration, so I was not chocked with my current approach (which had the benefit to be backward-compatible with the "old" GEOR_custom.js system). Maybe a file served as json could be a good compromise ? it would actually be a configuration file containing just plain data. I am not so fond of storing things in database either (do we really need the flexibility it offers ?).

PR #1038, opened 2015-08-31
mapfishapp - datadir

pmauduit:
This PR proposes the datadirization of the mapfishapp webapp:

- Exports the contexts, addons, JS configuration (including GEOR_custom.js) into the datadir
- Makes the rest of the webapp configurable outside of the webapp WAR (using the same mechanisms as previously)
- Allows mapfishapp to be easily customized by server-side addons (it allows automatic scanning of jars which classes begining with org.georchestra.mapfishapp.addons.*, as well as a applicationContext.xml to be defined. See https://github.com/pmauduit/osm2geor/tree/master/java as an example of use-case)

As the other PR, it requires the introduction of the georchestra-commons module. Travis-ci will fail until https://github.com/georchestra/georchestra/pull/1035 is merged.

Note for the reviewers: A lot of files are related to default addons inclusion, the majority of files added to src/deb/resources/etc are not necessary to be reviewed IMHO.

TODO:
* [x] Test compilation
* [x] testsuite
* [x] runtime test with `georchestra.datadir`
* [x] runtime test without `georchestra.datadir`


pmauduit:
Regression detected without the georchestra.datadir option: Images for the addons are broken.

pmauduit:
Ok fixed ; everything seems to go well now (without the `georchestra.datadir` variable)

pmauduit:
(FYI I pushed --force after rebasing / integration of the georchestra-commons module)

mapfishapp/pom.xml
```diff
@@ -405,5 +411,134 @@
...
+          <plugin>
+            <groupId>net.sf.debian-maven</groupId>
+            <artifactId>debian-maven-plugin</artifactId>
+            <version>1.0.6</version>
+            <configuration>
+              <packageName>georchestra-mapfishapp</packageName>
+              <packageDescription>Debian package for the geOrchestra's Viewer</packageDescription>
+              <projectOrganization>geOrchestra</projectOrganization>
+              <maintainerName>Pierre Mauduit</maintainerName>
+              <maintainerEmail>pierre.mauduit@gmail.com</maintainerEmail>
```
fvanderbiest: maintainer = PSC / psc@georchestra.org

mapfishapp/src/deb/resources/etc/georchestra/mapfishapp/addons/annotation/config.json
```diff
@@ -0,0 +1,17 @@
+[{
+        "id": "annotation_0", // unique & stable string identifier for this addon instance
```
fvanderbiest: Invalid json file (comments are not allowed)

mapfishapp/src/deb/resources/etc/georchestra/mapfishapp/addons/extractor/config.json
```diff
@@ -0,0 +1,28 @@
...
+        "roles": [
+            "ROLE_MOD_EXTRACTORAPP"
+        ],
+        "options": {
+            "srsData": [
+                ["EPSG:4326", "WGS84 (EPSG:4326)"],
+                ["EPSG:3857", "Spherical Mercator (EPSG:3857)"]
+            ],
+            "defaultSRS": "EPSG:4326",
+            "defaultVectorFormat": "shp", // must be one of shp, mif, tab, kml
```
fvanderbiest: same here, comments should be removed (or better: moved to the main README)

mapfishapp/src/deb/resources/etc/georchestra/mapfishapp/addons/fullscreen/config.json
```diff
@@ -0,0 +1,17 @@
+[{
```
fvanderbiest: The fullscreen addon is not yet an official one.
Where does this one come from ?

mapfishapp/src/deb/resources/etc/georchestra/mapfishapp/addons/magnifier/config.json
```diff
@@ -0,0 +1,17 @@
+[{
+        "id": "magnifier_0", // unique & stable string identifier for this addon instance
```
fvanderbiest: comments in json

mapfishapp/src/deb/resources/etc/georchestra/mapfishapp/addons/quicksearch/config.json
```diff
@@ -0,0 +1,52 @@
+[{
+        "id": "quicksearch_0",   // unique & stable string identifier for this addon instance
```
fvanderbiest: comments in this file should be mentioned in the README (if not already present)

mapfishapp/src/deb/resources/etc/georchestra/mapfishapp/addons/streetview/config.json
```diff
@@ -0,0 +1,21 @@
+[{
+        "id": "streetview_0", // unique & stable string identifier for this addon instance
```
fvanderbiest: comments

mapfishapp/src/deb/resources/etc/georchestra/mapfishapp/addons/streetview/config.json
```diff
@@ -0,0 +1,21 @@
...
+            "es": "StreetView",
+            "de": "StreetView"
+        },
+        "description": {
+            "en": "Get a detailed view of a place you choose on the map",
+            "es": "Obtener una vista detallada de un lugar que usted elija en el mapa",
+            "fr": "Obtenez une vue détaillée d'un endroit que vous choisissez sur la carte",
+            "de": "Sehen Sie eine Detailansicht eines Ortes, den Sie auf der Karte auswählen"
+        },
+        options: {
```
fvanderbiest: missing quotes for "options"

mapfishapp/src/deb/resources/etc/georchestra/mapfishapp/js/GEOR_custom.js
```diff
@@ -0,0 +1,80 @@
+Ext.namespace("GEOR");
+
+GEOR.custom = {
```
fvanderbiest: I think that lots of options are not mentioned in this file. Needs more work IMO.

mapfishapp/src/deb/resources/etc/georchestra/mapfishapp/js/GEOR_custom.js
```diff
@@ -0,0 +1,80 @@
...
+    HEADER_HEIGHT: 90,
```
fvanderbiest: Here we obviously miss the previous behavior where the header height was obtained from shared maven filters.
We need to talk about this one.

mapfishapp/src/deb/resources/etc/georchestra/mapfishapp/js/wfs.servers.json
```diff
@@ -0,0 +1,12 @@
+{
```
fvanderbiest: Are you sure these files belong to /etc/georchestra/mapfishapp/js ?
I would have said /etc/georchestra/mapfishapp

mapfishapp/src/deb/resources/etc/georchestra/mapfishapp/print/README.md
```diff
@@ -0,0 +1,7 @@
+Setup
+=======
+
+To take into account this folder (and ignore the parameter set in web.xml), it
+is needed to declare a `mapfish-print-config` at JVM launch, i.e. adding a
+`-Dmapfish-print-config=/path/to/config.yaml` in `CATALINA_OPTS`.
```
fvanderbiest: Good thing you told me :-)

mapfishapp/src/deb/resources/etc/georchestra/mapfishapp/print/config.yaml
```diff
@@ -0,0 +1,1026 @@
...
+  - 8735660
+  - 17471321
+  - 34942642
+  - 69885283
+  - 139770566
+  - 279541132
+  - 559082264
+
+# see https://github.com/georchestra/georchestra/issues/362
+brokenUrlPlaceholder: http://upload.wikimedia.org/wikipedia/commons/c/ce/Transparent.gif
```
fvanderbiest: Was replaced by https://github.com/georchestra/template/blob/15.06/mapfishapp/WEB-INF/print/config.yaml#L40

I hope we won't forget other things like that ...

mapfishapp/src/main/filtered-resources/WEB-INF/jsp/index.jsp
```diff
@@ -4,19 +4,50 @@
...
 
 String lang = request.getParameter("lang");
 if (lang == null || (!lang.equals("en") && !lang.equals("es") && !lang.equals("ru") && !lang.equals("fr") && !lang.equals("de"))) {
+  if (defaultLanguage != null) {
+    lang = defaultLanguage;
+  }
+  else {
     lang = "${language}";
+  }
+}
```
fvanderbiest: I'm not sure this logic enables the user to override the defaultLanguage by passing ?lang=XX to the webapp

mapfishapp/src/main/filtered-resources/WEB-INF/jsp/index.jsp
```diff
@@ -4,19 +4,50 @@
...
+
 <%@ page contentType="text/html; charset=UTF-8" %>
 <%@ page pageEncoding="UTF-8" %>
 <%@ page isELIgnored="false" %>
 <%
 Boolean anonymous = true;
 Boolean admin = false;
+Boolean georDatadirActivated = false;
+
+String instanceName = null;
```
fvanderbiest: By default: instanceName = "geOrchestra" ?

mapfishapp/pom.xml
pmauduit: As said, I fear that I had to regenerate a GPG key so that I can publish the signed packages on build.georchestra.org ; it will deserve a try in my georchestra-datadir branch, and see if the CI is not broken by the change.

Anyway, I've no objection for such a change

mapfishapp/src/main/java/org/georchestra/mapfishapp/ws/AddonController.java
```diff
@@ -0,0 +1,163 @@
...
+            response.setStatus(HttpServletResponse.SC_BAD_REQUEST);
+            return;
+        }
+        File actualFile = new File(String.format("%s%s%s", datadir, File.separator, pathInfo));
+        if (! actualFile.exists()) {
+            LOG.error("requested file does not exist: " + actualFile);
+            response.setStatus(HttpServletResponse.SC_NOT_FOUND);
+            return;
+        }
+        String ext = FilenameUtils.getExtension(actualFile.getAbsolutePath());
```
fvanderbiest: Isn't this dangerous ?

mapfishapp/src/main/java/org/georchestra/mapfishapp/ws/AddonController.java
```diff
@@ -0,0 +1,163 @@
...
+
+        if ("css".equalsIgnoreCase(ext)) {
+            response.setContentType("text/css; charset: UTF-8");
+        } else if ("js".equalsIgnoreCase(ext)) {
+            response.setContentType("application/javascript; charset: UTF-8");
+        } else if ("png".equalsIgnoreCase(ext)) {
+            response.setContentType("image/png");
+        } else {
+            response.setContentType("text/plain");
```
fvanderbiest: ouch, what about jpg/jpeg files and other potential types ...

mapfishapp/src/main/java/org/georchestra/mapfishapp/ws/ContextController.java
```diff
@@ -0,0 +1,208 @@
...
+        XPathExpression xpTitle = xpath.compile("//ViewContext/General/Title/text()");
+        Object oTitle = xpTitle.evaluate(doc, XPathConstants.STRING);
+
+        // Parsing tip (abstract)
+        XPathExpression xpAbstract = xpath.compile("//ViewContext/General/Abstract/text()");
+        Object oAbstract = xpAbstract.evaluate(doc, XPathConstants.STRING);
+
+        // Parsing keywords
+        XPathExpression xpKeyword = xpath.compile("//ViewContext/General/KeywordList/Keyword");
+        Object oKeyword = xpKeyword.evaluate(doc, XPathConstants.NODESET);
```
fvanderbiest: :+1:

mapfishapp/src/main/java/org/georchestra/mapfishapp/ws/DocController.java
```diff
@@ -41,36 +44,65 @@
...
-	/** the postgresql connection pool used by the document services*/
+    /** the connection pool used by the document services*/
+    @Autowired
     private ConnectionPool connectionPool;
 
-    public ConnectionPool getConnectionPool() {return connectionPool;}
-	public void setConnectionPool(ConnectionPool connectionPool) {	this.connectionPool = connectionPool; }
+    public void init() throws IOException {
+        if (georchestraConfiguration.activated()) {
+            docTempDir = georchestraConfiguration.getProperty("docTempDir");
```
fvanderbiest: docTempDir could probably be removed, now that a db stores these documents

fvanderbiest: hmm, maybe not, since it seems to be still used for uploads cache (unless we can store them elsewhere)

mapfishapp/src/main/java/org/georchestra/mapfishapp/ws/WxsJsonController.java
```diff
@@ -0,0 +1,75 @@
...
+                ((! "wms".equalsIgnoreCase(proto)) &&
+                 (! "wmts".equalsIgnoreCase(proto)) &&
+                 (! "wfs".equalsIgnoreCase(proto))))
+        {
+            response.setStatus(HttpServletResponse.SC_BAD_REQUEST);
+            response.getOutputStream().write(String.format("Bad parameter value: \"%s\"", proto).getBytes());
+            return;
+        }
+
+        String wxsJsonPath = String.format("%s/js/%s.servers.json", georConfig.getContextDataDir(),
```
fvanderbiest: ok, now I get it.
I think that the js dir should not host these files.
Either webapp root or another dir, such as /mapfishapp/config/wxs.servers.json
What do others think about it ?

mapfishapp/src/test/resources/js/wms.servers.json
```diff
@@ -0,0 +1 @@
+[{'sample': 'json to test WxsJsonController'}]
```
fvanderbiest: double quotes

mapfishapp/src/test/resources/org/georchestra/mapfishapp/ws/addons/annotation/config.json
```diff
@@ -0,0 +1,17 @@
+[{
+  "id": "annotation_0", // unique & stable string identifier for this addon instance
```
fvanderbiest: comments

mapfishapp/src/test/resources/org/georchestra/mapfishapp/ws/addons/magnifier/config.json
```diff
@@ -0,0 +1,17 @@
+[{
+  "id": "magnifier_0", // unique & stable string identifier for this addon instance
```
fvanderbiest: comments

fvanderbiest:
I'm done.
Mostly concerned by the missing config options in GEOR_custom.js.

mapfishapp/pom.xml
pmauduit: Let's see: https://build.georchestra.org/ci/job/georchestra-generic-wars/55/ / https://github.com/pmauduit/georchestra/commit/6f372fc3e276f10635d80dbd814248de23fff3a3

mapfishapp/src/deb/resources/etc/georchestra/mapfishapp/addons/fullscreen/config.json
pmauduit: I wanted to try it out in my VM

mapfishapp/src/main/java/org/georchestra/mapfishapp/ws/WxsJsonController.java
landryb: Why webapp root or inside the webapp ? Doesnt that defeat the purpose of having all the config *outside* the webapp and centralized in /etc/georchestra/mapfishapp ?

mapfishapp/src/deb/resources/etc/georchestra/mapfishapp/js/GEOR_custom.js
pmauduit: I'll synchronize the GEOR_custom.js from template into the default debian package provided file. (Anyway, we will encounter the same issue as the one we just discussed with the addons ...)

mapfishapp/src/deb/resources/etc/georchestra/mapfishapp/js/GEOR_custom.js
pmauduit: If `georchestra.datadir`:

It is the administrator's responsability to ensure his datadir is correctly configured, the main added value using the datadir is that the administrator is able to update the parameter without rebuilding the whole suite/webapp.

if `! georchestra.datadir`:
Then one will use the GEOR_custom from the build, so it would have the expected HEADER_HEIGHT.

So, I cannot see the point

mapfishapp/src/deb/resources/etc/georchestra/mapfishapp/js/wfs.servers.json
pmauduit: Implementation detail, I've no objection on storing them into `mapfishapp` or `mapfishapp/js` if you think it is more relevant

fvanderbiest:
As said with @pmauduit I'm also concerned by addon code duplication: in the webapp and in the data dir provided by the package.

We agreed to say that:
 * the default addons should stay where they are (webapp/add/addons)
 * if the datadir is activated and addons are found in /etc/georchestra/mapfishapp/addons, they override those from the webapp (based on addon id)

mapfishapp/src/main/filtered-resources/WEB-INF/jsp/index.jsp
pmauduit: The previous behaviour is not altered:
```
If 
  a parameter lang is present (line 17 / 40)
then
  use it
else
  if
    default language is not null (line 42)
  then
    use it (note: it comes from the mapfishapp.properties from the datadir) (line 43)
  else
    use the one from the shared.maven.filter which would be replaced at compilation time (line 46)
  end
end
```

mapfishapp/src/deb/resources/etc/georchestra/mapfishapp/addons/fullscreen/config.json
fvanderbiest: Yes, thanks.

mapfishapp/src/deb/resources/etc/georchestra/mapfishapp/js/GEOR_custom.js
fvanderbiest: You're absolutely right that there is code duplication here.
That's why I was talking about "technical debt" yesterday: maybe we should move all the config options from GEOR_custom.js to a server-side properties file, and have a controller dedicated to passing these config options from the server to the client-side in both cases (I mean with and without the data dir activated). Maybe we could also insert these options into the ModelAndView that mapfishapp's HomeController handles.

mapfishapp/src/main/filtered-resources/WEB-INF/jsp/index.jsp
pmauduit: instanceName takes the value from the datadir if available (line 27), or the value specified at compilation level in the shared.maven... (line 50).

Do not forget that the current jsp file is situated in filtered-resources/ which would be copied at compilation time into the right path (src/main/webapp/...), allowing to remain compatible with the "old" system (no datadir).

mapfishapp/src/deb/resources/etc/georchestra/mapfishapp/js/GEOR_custom.js
fvanderbiest: ok

mapfishapp/src/deb/resources/etc/georchestra/mapfishapp/js/wfs.servers.json
fvanderbiest: Yes, the ```js``` dir should be reserved for code, not config.

I know GEOR_custom.js is stored in the js folder ... it should probably not be.

mapfishapp/src/main/java/org/georchestra/mapfishapp/ws/AddonController.java
pmauduit: I've a basic check that ensures we actually are in the datadir (line 80), and I hardly remember having coded a unit test that "penetration tests" this piece of code. Of course, having some code that dumps file to the client without basic checks on what is requested is actually dangerous, hence my check on line 80. I think this should address the LFI (local file inclusion) risk, by jailing the possible files served into georchestra.datadir/mapfishapp/addons/

mapfishapp/src/main/filtered-resources/WEB-INF/jsp/index.jsp
fvanderbiest: ok, good.

mapfishapp/src/main/java/org/georchestra/mapfishapp/ws/AddonController.java
pmauduit: I looked for a library that could guess the file type, but did not find any relevant one to be used here. I think that limiting the file formats here is an acceptable compromise

mapfishapp/src/main/filtered-resources/WEB-INF/jsp/index.jsp
fvanderbiest: ok

mapfishapp/src/main/java/org/georchestra/mapfishapp/ws/AddonController.java
fvanderbiest: OK, but please add jpg/jpeg support.

mapfishapp/src/main/java/org/georchestra/mapfishapp/ws/DocController.java
fvanderbiest: OK, let's keep it this way, we won't break older permalinks.

pmauduit:
To sum the remarks up:
* [ ] maintainer PSC / psc@georchestra.org
* [ ] removing addons from src/deb/resources/
* [ ] rework on the addons controller so that it scans the webapp addons/ directory, and overrides if present by the ones present in datadir
* [ ] GEOR_custom.js synchronized with the one from the template (so that comments are kept)
* [ ] moving files from datadir/mapfishapp/js/ to datadir/mapfishapp/config/wxs.servers.json, s/simple quotes/double quotes/
* [ ] updating config.yaml for the print provided by the default debian package
* [ ] adding JPEG as file types served into the addons subdirectory

Am I missing something else ?

mapfishapp/src/main/java/org/georchestra/mapfishapp/ws/WxsJsonController.java
fvanderbiest: I just meant that these files in the data dir should probably be stored in /etc/georchestra/mapfishapp/ rather than /etc/georchestra/mapfishapp/js

pmauduit:
I will do extra tests by merging back the branch into my own one.

pmauduit:
Runtime tests: Ok (basic tests, mainly on the addons management, with and without `georchestra.datadir`).

landryb:
I'm testing this, with georchestra.datadir set and some customizations done in /etc/georchestra/mapfishapp/js/GEOR_custom.js and mapfishapp/w?s.servers.json, but it seems the values returned by the w?s.servers.json webservice come from the json files in the webapp, and GEOR_custom.js is fetched from the webapp dir too. Interestingly, i see no addon at all listed in 'tools' window, even if some are listed in the default GEOR_custom.js found/used within the webapp. Guess i'll need to dig a bit more....

the mapfishapp/log4j/log4j.properties and mapfishapp/mapfishapp.properties files are properly read/applied from the datadir.

/mapfishapp/addons/ and /mapfishapp/contexts/ yield 404's..

maybe relevant info from mapfishapp.log :
```
2015-09-14 09:08:23 BaseMapServlet [INFO] app is null, setting it as default configPath: WEB-INF/print/config.yaml
2015-09-14 09:08:23 BaseMapServlet [INFO] app is null, setting it as default configPath: WEB-INF/print/config.yaml
2015-09-14 09:08:23 BaseMapServlet [INFO] Attempting to locate app config file: 'WEB-INF/print/config.yaml in the webapplication.
2015-09-14 09:08:23 BaseMapServlet [INFO] Attempting to locate app config file: 'WEB-INF/print/config.yaml in the webapplication.
2015-09-14 09:08:23 BaseMapServlet [INFO] Unable to find config file in web application using getRealPath.  Adding a / because that is often dropped
2015-09-14 09:08:23 BaseMapServlet [INFO] Unable to find config file in web application using getRealPath.  Adding a / because that is often dropped
2015-09-14 09:08:23 BaseMapServlet [INFO] Loading app from: /data/tomcat/georchestra/webapps/mapfishapp/WEB-INF/print/config.yaml
2015-09-14 09:08:23 BaseMapServlet [INFO] Loading app from: /data/tomcat/georchestra/webapps/mapfishapp/WEB-INF/print/config.yaml
2015-09-14 09:08:23 BaseMapServlet [INFO] Loading configuration file: /data/tomcat/georchestra/webapps/mapfishapp/WEB-INF/print/config.yaml
2015-09-14 09:08:23 BaseMapServlet [INFO] Loading configuration file: /data/tomcat/georchestra/webapps/mapfishapp/WEB-INF/print/config.yaml
```

landryb:
Hm, i was mistaken, at mapfishapp loading /ws/app/js/GEOR_custom.js is called and it effectively returns the content of the file found in the datadir.

landryb:
/ws/addons/ returns
```
{"success":false, "error":"fileupload_error_ioError", "msg": "null"} 
```
java.lang.NullPointerException at org.georchestra.mapfishapp.ws.AddonController.buildAddonSpecs(AddonController.java:188)

/ws/contexts/ returns the default geobretagne context, which seems expected, although it isnt loaded at initial mapfishapp display. It seems nothing calls /ws/contexts/ nor /ws/addons/ at startup....

landryb:
As for the w?s servers, by default mapfishapp/wms.servers.json is loaded/queried (which returns the default list from the webapp) while i think it should be mapfishapp/ws/wms.servers.json (which contains my custom list..)

pmauduit:
TODO:
- [ ] Remove contexts and addons from the provided `GEOR_custom.js`
- [ ] hardenning contexts / addons to avoid NPE in case of non existing directory in the datadir

landryb:
For the w?s server list, i think GEOR_LayerBrowser.js just lacks /ws/ being prepended to GEOR.config.OGC_SERVERS_URL[this.id] in line 59, otherwise the list comes from the webapp.

Tested this, works..
```
--- a/mapfishapp/src/main/webapp/app/js/GEOR_LayerBrowser.js
+++ b/mapfishapp/src/main/webapp/app/js/GEOR_LayerBrowser.js
@@ -56,7 +56,7 @@ GEOR.LayerBrowser = Ext.extend(Ext.Panel, {
         this.serverStore = new Ext.data.Store({
             proxy: new Ext.data.HttpProxy({
                 url: GEOR.util.getValidURI(
-                    GEOR.config.OGC_SERVERS_URL[this.id]
+                    "ws/" + GEOR.config.OGC_SERVERS_URL[this.id]
                 ),
                 method: 'GET',

```

landryb:
Maybe only local to my setup, but there are some encoding issues in various places:

- in the addons list, their description (coming from  webapp/app/addons/*/*.json) is double-encoded, yielding ?? instead of accentued letters. This is not the case with default contexts where the accents coming from datadir/contexts/*.wmc are properly rendered - and the addons description is fine in http://sdi.georchestra.org/mapfishapp/?lang=fr
- saving a context also double-encodes accentued letters (looking bad in mapfishapp db too, although the db is UTF8), badly displayed when re-sending the wmc. Tested the same, works on sdi.georchestra.org.

landryb:
More details on the encodings:

- if i browse to /mapfishapp/ws/contexts/ in firefox, an accentued in displayed as a single ? but they're properly displayed in the wmc browser. curl -i on the url says that content-type is app/js;charset=UTF-8
- if i browse to /mapfishapp/ws/addons/ in firefox, an accentued letter is displayed as two ??, and they're also displayed this way (ie badly) in the addons browser. curl -i also says that content-type is app/js;charset=UTF-8
- if i browse to /mapfishapp/ws/wmts.servers.json in firefox, an accentued letter is displayed as utf vomit (ie GÃ©oPortail) and is displayed 'correctly' in the server list dropdown menu. curl -i doesnt specify a content-type, and the accent is properly rendered in curl output in my term

pmauduit:
> if i browse to /mapfishapp/ws/contexts/ in firefox, an accentued in displayed as a single ? but they're properly displayed in the wmc browser. curl -i on the url says that content-type is app/js;charset=UTF-8

Have you set the URIencoding option onto your tomcat connectors ?
<Connector URIEncoding="UTF-8"  ...>

landryb:
I had URIEncoding="UTF-8" set on all the tomcat connectors, but the issue was somewhere else. Setting LANG=C.UTF-8 (or any of the other UTF-8 locales available, check locale -a output) in /etc/default/tomcat-* worked around the issue. Guess that's what i get for not having it set in /etc/default/locale......

landryb:
Small buglet in addons handling.. added openls,magnifier,cadastrapp,streetview & annotation in /etc/georchestra/mapfishapp/addons, loaded /ws/addons -> extractorapp was in the list (and enabled) although not present in the datadir - i think this was caused by cadastrapp not having a config.json file (oversight from me) - i had the error message, but somehow some addon was randomly taken from the webapp..
```
2015-09-15 16:50:54 ws [ERROR] Addon cadastrapp does not have a config.json configuration file, skipping it.
```

landryb:
okay, after more explanation from @pmauduit it makes sense to have extractor in the list of app since it's in the webapp, and enabled by default. To disable it, one just needs to make a copy of it in the datadir, and set enable to false in its config.json.

landryb:
And with the current logic in https://github.com/georchestra/georchestra/blob/mapfishapp-datadir/mapfishapp/src/main/java/org/georchestra/mapfishapp/ws/AddonController.java#L200 (and also the two loops over addons arrays in constructAddonsSpec()) it doesnt seem possible to 'disable' an addon that is enabled in the webapp. If i copy extractor to my datadir and change enabled to false in config.json, it will still be listed in the available & enabled addons.

pmauduit:
Done today:
- Removing remaining comments in the standard mapfishapp addons
- Fixing bug with disabled addons in the overriden georchestra.datadir path
- Rework to harden context / addon controller a little bit (avoids NPE)
- fixing bug with the WxS access

landryb:
Retested tip of the PR, all the previous issues i had are now fixed, i've been able to disable extractor addon and the wxs server lists are fetched from the datadir. Thanks!

fvanderbiest:
Good, please merge.

PR #1040, opened 2015-09-01
Use groupOfMembers instead of groupOfNames

landryb:
groupOfNames doesn't allow to use empty groups, which leads to hacks like fakeuser... so let's use it instead, and get rid of fakeuser. 

Note that this PR requires https://github.com/georchestra/LDAP/pull/5.


fvanderbiest:
LGTM. 
Waiting to hear about georchestra/LDAP#5

PR #1053, opened 2015-09-04
Allow to cluster ldapadmin

landryb:
This way, ldapadmin can be clusterized between several tomcat instances, adding
<distributable/> to web.xml. Otherwise, account/userdetails was complaining
that EditUserDetailsFormBean wasn't serializable in a session, which is a
requirement for tomcat clustering.

```
java.lang.IllegalArgumentException: setAttribute: Non-serializable attribute editUserDetailsFormBean
```


fvanderbiest:
LGTM, thanks Landry :-)

PR #1055, opened 2015-09-05
Mapfishapp - Fix several potential XEE vulns

pmauduit:
Some times ago, between the 14.12 and the 15.06 version, I introduced some code that allows to pretty-print provided XML before saving it into the database, this to let people XML-aware the possiblity to be able to read the saved documents with more ease.

This code actually introduced a vulnerability that I was able to exploit with a PoC (Yes, XEE attacks are trendy these days). Since I used a XML parser that was External-entity-resolution-aware, it was possible to disclose any files from the webserver in every saved documents in Mapfishapp.

This PR addresses the flaw and limits the exploitation on the WMC controller (the vulnerability on the modified code on WMCDocService is unproved, but it should not harm to be patched).


fvanderbiest:
Thanks a lot Pierre, but the PR should target 14.12 rather than 15.06.

pmauduit:
As a contributor, all these version issues are really driving me crazy ...

Anyway that is weird, when I checked I could not manage to find my code back in the 14.12 version, hence the current PR onto 15.06

PR #1063, opened 2015-09-09
Fix group filter not applying correctly on selections

pgiraud:
Fixes camptocamp/georchestra-cigalsace-configuration#549 (comment)


fvanderbiest:
not 15.06 ... 14.06 is the one.

PR #1064, opened 2015-09-09
Fix group filter not applying correctly on selections

pgiraud:
Fixes camptocamp/georchestra-cigalsace-configuration#549

Sorry for the noise.

PR #1065, opened 2015-09-09
Fix group filter not applying correctly on selections

pgiraud:
Same fix as in https://github.com/georchestra/georchestra/pull/1064 but for 15.06


fvanderbiest:
Thanks

Issue #981, opened 2015-05-20
mapfishapp: fullscreen addon


fvanderbiest:
See http://geofao.org.bo/mapfishapp/app/addons/fullscreen/manifest.json or (mirror) https://gist.github.com/fvanderbiest/b004818d8bd3f77cce02

Would be good to see it more widely deployed => review + integration in geOrchestra sources

MaelREBOUX:
Would be great to add it to the 15.12 milestone.

fvanderbiest:
IIRC CIGALsace is very interested by this one too for 15.12

fvanderbiest:
Expected compatibility with IE >= 10

fvanderbiest:
http://caniuse.com/#feat=fullscreen says the API is supported by IE >= 11
As a result, the addon will work in a degraded fashion with IE10: sidebars will be collapsed.

fvanderbiest:
http://sdi.georchestra.org/mapfishapp/

Tiles are not always loading at the edges, when fullscreened ... I'm wondering if the map div has the correct size...

Please report other potential issues.

fvanderbiest:
All done, with documentation: https://github.com/georchestra/georchestra/blob/15.06/mapfishapp/src/main/webapp/app/addons/fullscreen/README.md

PR #1087, opened 2015-10-13
fullscreen addon

fvanderbiest:
This is a proposal for #981


fvanderbiest:
Deployed to http://sdi.georchestra.org/mapfishapp/ and available for the 14.12 & 15.06 releases

Issue #1058, opened 2015-09-07
ogc-server-statistics logs requests to /proxy/ too


fvanderbiest:
ogc-server-statistics currently logs all requests going through the security proxy, including those targeting external servers (matching the /proxy/ endpoint, used to circumvent browser's "same origin policy").

These requests appear in analytics, but the view is very incomplete, since we lack the external server URL.

So, IMO, we have two choices:
 - [solution 1] either we drop support for requests going through /proxy/?url=
 - [solution 2] or we add support for the service endpoint

fvanderbiest:
Voting for solution 1.

pmauduit:
also :+1: for solution 1

fphg:
+1 for 1
however requests still shall be file logged for security reasons

pmauduit:
> however requests still shall be file logged for security reasons

Since they still will be intercepted by the frontend webserver, this would end up in the webserver logfiles logically, independantly of what is done by the Security-proxy afterwards.

landryb:
While here, see also #1027 ?

fvanderbiest:
> While here, see also #1027 ?

right.

PR #1090, opened 2015-10-16
ogc server statistics - do not log if request made onto the /proxy endpoint

vdorut:
See #1058 
Also addresses an useless extra call to handleRequest()

Tests: runtime tested


fvanderbiest:
LGTM, thanks Vincent :-)

PR #1100, opened 2015-10-22
Ldapadmin remove fields - bugfix for #1097

### LINKED ISSUE #1097: LDAPAdmin - removing blank fields introduces another bug
When blanking a field in LDAPAdmin, the userPassword field is removed. There might be a bug in the Account class which does not load the userPassword when mapping from the LDAP.

This is related to recent work on attribute removal (see https://github.com/georchestra/georchestra/pull/1093)

vdorut:
Remove the userPassword of the maping + add test with real LDAP
fixes #1097

Tests:
- runtime: ok
- Integration test (against real LDAP with default geOr base): ok


pmauduit:
Looks good, thanks ;-)

PR #1103, opened 2015-10-27
Distributable webapps

landryb:
Realized i forgot to create a PR for this one.. other webapps (analytics, extractor, etc...) untested yet.

geoserver web ui and geonetwork 3 known not compatible with this (yet?).


fvanderbiest:
LGTM, thanks Landry !

Issue #1111, opened 2015-11-09
Geowebcache Version


jusabatier:
I had a NullPointer Exception on some tiles/zoomlevel of an Ortho, I tried to just update my standalone GWC to latest version (1.8.0). And with the new version the error is solved. Maybe could be a good thing to update the GWC used in Georchestra to latest version ?

pmauduit:
It should be as easy as to modify the following line:
https://github.com/georchestra/georchestra/blob/master/geowebcache-webapp/pom.xml#L16

But honestly I don't have the time to test it for now, do you mind giving it a try ?

jusabatier:
Yes I'll try it :)

PR #1114, opened 2015-11-09
Upgrade to GWC v1.8.0

jusabatier:
Since I had some bug with current version, I tried to upgrade GWC to latest version (1.8.0) and it seems to be OK.

cf https://github.com/georchestra/georchestra/issues/1111


fvanderbiest:
Thanks Julien !

Issue #1115, opened 2015-11-09
ldapadmin API should return json content-type


landryb:
Right now the API sets content-type header to text/html when being called from the web ui, while it'd be more 'correct' for it to be 'application/json' .. since the actual value is json.

Also, in https://github.com/georchestra/georchestra/blob/master/ldapadmin/CLIENT_SERVER_INTERFACE.md#use-of-curl i dont really understand why the Content-Type header is set to text/xml since we POST json content...

fvanderbiest:
agreed, +1

pmauduit:
Reproduced:

```
$ curl -i 'http://sdi.georchestra.org/ldapadmin/private/users' ...

HTTP/1.1 200 OK
Date: Mon, 09 Nov 2015 13:28:37 GMT
Server: Apache-Coyote/1.1
Content-Type: text/html;charset=UTF-8
Vary: Accept-Encoding
Content-Encoding: gzip
Content-Length: 1889
Keep-Alive: timeout=5, max=100
Connection: Keep-Alive
```
Should be easily fixed using the `@Produces` annotation, see https://docs.oracle.com/cd/E19798-01/821-1841/gipxf/index.html ; (should also work on a method)

pmauduit:
Tested into generated vm on current master ldapadmin package

PR #1117, opened 2015-11-10
Ldapadmin return json 1115

vdorut:
https://github.com/georchestra/georchestra/issues/1115

### DIFF
--- a/ldapadmin/src/main/java/org/georchestra/ldapadmin/ws/backoffice/users/UserListResponse.java
+++ b/ldapadmin/src/main/java/org/georchestra/ldapadmin/ws/backoffice/users/UserListResponse.java
@@ -25,7 +25,7 @@ import org.springframework.web.bind.annotation.RequestMethod;
  * @author Mauricio Pazos
  *
  */
-@RequestMapping(produces = "application/json")
+@RequestMapping(produces = "application/json; charset=utf-8")
 final class UserListResponse {
 	
 	private List<Account> accountList;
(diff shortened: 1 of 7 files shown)


vdorut:
runtime tested

landryb:
Do we even need to set the Content-Type header in the curl GET example ? Since by default json is returned.. i think Content-Type is needed only for POST. Other that that, reads good to me, and thanks for working on it :)

fvanderbiest:
LGTM, please merge.

ldapadmin/src/main/java/org/georchestra/ldapadmin/ws/backoffice/users/UserListResponse.java
```diff
@@ -23,6 +25,7 @@
  * @author Mauricio Pazos
  *
  */
+@RequestMapping(produces = "application/json")
```
pmauduit: charset ?

ldapadmin/src/main/java/org/georchestra/ldapadmin/ws/backoffice/users/UsersController.java
```diff
@@ -140,7 +140,7 @@ public void findAll( HttpServletRequest request, HttpServletResponse response )
 	 * @param response Returns the detailed information of the user as json
 	 * @throws IOException
 	 */
-	@RequestMapping(value=REQUEST_MAPPING+"/*", method=RequestMethod.GET)
+	@RequestMapping(value=REQUEST_MAPPING+"/*", method=RequestMethod.GET, produces = "application/json")
```
pmauduit: charset ?

pmauduit:
Does not work after more thoroughly checks, needs rework

PR #1124, opened 2015-11-17
[Mapfish] Correct line return in context list

jusabatier:
When the context title is too long, there is no line return and the context is larger than others and it shift all next contexts. That's ugly.

With this css rule, the width of each context is fixed so it prevent that shift.


fvanderbiest:
LGTM, thanks !

PR #1127, opened 2015-11-19
Add emails

Vampouille:
Work in progress !


Vampouille:
Please don't merge this branch !

fvanderbiest:
@Vampouille says it's good to go now. 
So, we need @pmauduit or @fgravin for a review.
I'll have a quick look too :-)

pmauduit:
Looks good (and thanks for having repaired the testsuite ;))

pmauduit:
Just a side-note: it will break people using older versions of tomcat (servlet-3.x required, tomcat >=7 needed, c.f. http://tomcat.apache.org/whichversion.html)

fvanderbiest:
> Just a side-note: it will break people using older versions of tomcat (servlet-3.x required, tomcat >=7 needed

Then the documentation should be updated accordingly in this PR. 
Thanks !

This PR should probably also update https://github.com/georchestra/georchestra/blob/master/doc/setup/postgresql.md#ldapadmin-schema

fvanderbiest:
As a side note, I would have prefered a separate PR with the remaining updates to Jetty.

pmauduit:
> This PR should probably also update https://github.com/georchestra/georchestra/blob/master/doc/setup/postgresql.md#ldapadmin-schema

Done in the first commit of the next branche https://github.com/georchestra/georchestra/commit/a71a1e336aacf25b86adcd6782a835d31d8e078f

> Then the documentation should be updated accordingly in this PR. 
> Thanks !

The UserToken is still needed / not managed by JPA, so the doc does not need to be updated at this point.

PR #1132, opened 2015-11-24
[Mapfish] [WIP] Addon for search places on OSM

jusabatier:
This addon use the overpass API for retrieve the places name, because it allow wildcards in search terms. Once the completes search terms associated to the query are returned, it use nominatim to get the full entity name (allow to get the City name and more if needed) and geometry.

Hope it could help, I created it because my users want to search in same place town, ways, places , ...

And the best data source I could have for this purpose is OSM.


fvanderbiest:
Thanks for the contribution Julien.
Please fix the indents according to our coding style and wrap longer lines if you expect a review.

jusabatier:
Code fix done :) I always have problems with indent when I import from my Notepad++ :/

mapfishapp/src/main/webapp/app/addons/osmsearchplaces/manifest.json
```diff
@@ -0,0 +1,27 @@
+{
+    "js": [
+        "js/main.js"
+    ],
+    "default_options": {
+		"comboTemplate": "{display}",
```
fvanderbiest: indents here

jusabatier:
Example here : https://opendata.agglo-lepuyenvelay.fr/mapfishapp/

MaelREBOUX:
Nice add-on.
We also have a similar add-on for our RVA based ou our API  : https://github.com/sigrennesmetropole/addon_rva

Would be nice to merge into one add-on wich aloow to switch on RVA, BAN, OAPI.
To keep in mind.

jusabatier:
There still are some problems with this addon : 
- I sometimes get a too many request reponse from OSM service
- Some places are missing from the OSM DB (lieux-dits, rues, ...)

I developed another addon for search which fit better to our user's needs.
It allow to configure the referential(s) your want to search in and to filter on a city/region/bounds...
You can test it on http://opendata.agglo-lepuyenvelay.fr/mapfishapp
The bad point is that doesn't allow to suggest/auto-correct user input and is sensitive to accents as it's based on WFS 1.1.0
Tell me if some of you are interested by this addon, I'll make a PR.

fvanderbiest:
> Tell me if some of you are interested by this addon, I'll make a PR.

You say your new addon is better than the one you propose in the current PR ?
If so, yes, please close this PR and submit another one, I'll have a look.

jusabatier:
The new addon have different purpose : 
It allow to define WFS layers as referentials for the searches like cities, ways, shops, etc...
And to define a WFS layer for define a BBOX to search into.

So it doesn't use the OSM datasource but any WFS one.

fvanderbiest:
mapfishapp EOL. Closing.

PR #1133, opened 2015-11-25
ldapadmin,mapfishapp,extractorapp - commons-fileupload bump

pmauduit:
the current version of commons-fileupload is vulnerable to a known vulnerability, which might cause a deny-of-service[1].

Tests: Compilation time, would require runtime tests (once merged and the VM is regenerated on the CI)

Note: I am wondering why this dependency is necessary in extractorapp / ldapadmin ? From what I'm aware of, sending file from the browser is only possible from mapfishapp.

[1] https://www.cvedetails.com/cve/CVE-2014-0050/

### DIFF
--- a/extractorapp/pom.xml
+++ b/extractorapp/pom.xml
@@ -59,7 +59,7 @@
     <dependency>
       <groupId>commons-fileupload</groupId>
       <artifactId>commons-fileupload</artifactId>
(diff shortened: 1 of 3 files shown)


fvanderbiest:
Indeed, there is no file upload from ldapadmin or extractorapp.

Issue #1136, opened 2015-11-27
broken ldapadmin links in header


landryb:
A bit different from #1096 and fallout of #1010 i think - with this config:

```
cas/cas.properties:ldapadmin.contextpath=/ldapadmin/
header/header.properties:ldapadminPublicContextPath=/ldapadmin/
ldapadmin/js/config.js:    publicContextPath: "//ldapadmin/"
ldapadmin/ldapadmin.properties:publicContextPath=/ldapadmin/
```
In my header, the ldapadmin link looks like this:
```
<a href="${ldapadminPublicContextPath}/privateui/">
```

which resolves to http://georchestra.craig.fr/header/$%7BldapadminPublicContextPath%7D/privateui/

And the 'user account details' link looks like this:
```
<a href="${ldapadminPublicContextPath}/account/userdetails">
```
which resolves to
http://georchestra.craig.fr/header/$%7BldapadminPublicContextPath%7D/account/userdetails

I think this is due to the removal of the double // prefixing ldapadminPublicContextPath, and on top of this the value doesnt seem correctly read from the config.

My header index.jsp comes from the my own build template (ie georchestra/config/configurations/craig-template/header/WEB-INF/jsp/index.jsp, copied/adapted from georchestra/header/src/main/filtered-resources/WEB-INF/jsp/index.jsp) and is properly propagated/deployed to the webapp.

landryb:
Maybe due to this commit bdbbab01600ce8d2b418adcf705431191c5540b7 ?

landryb:
Hm probably not since this should only be for jetty...

landryb:
As i expected reverting bdbbab0 locally doesnt fix the issue :) .. it's like the config file in `${georchestra.datadir}/header/header.properties` was ignored.

landryb:
It's probably the case, because if i go to `view-source:http://georchestra.craig.fr/header/` -> default language, the fr value is not taken from the properties file.

landryb:
Okay, added some debugging gunk to the jsp here https://github.com/georchestra/georchestra/blob/master/header/src/main/filtered-resources/WEB-INF/jsp/index.jsp#L30..

```
exception: No bean named 'class org.georchestra.commons.configuration.GeorchestraConfiguration' is defined
```
but there's the <%@page import pointing at it, and the class should be here...
```
[...]

PR #1138, opened 2015-11-27
Fix loading properties by using the same construct as in other module…

landryb:
…s (#1136)

Otherwise an exception is trigerred:
No bean named 'class org.georchestra.commons.configuration.GeorchestraConfiguration' is defined


fvanderbiest:
LGTM, thanks Landry !

Issue #132, opened 2013-06-11
mapfishapp: better addons


fvanderbiest:
- handle more accurately interactions between activated tools
- preloading of an addon for a group of users, ~~or via GET parameters~~
- search in addon list
- group addons into submenus
- ~~possibility to insert the addon-generated component at a given place in the UI~~

fvanderbiest:
Related PR for "possibility to insert the addon-generated component at a given place in the UI" is https://github.com/georchestra/georchestra/pull/894

fvanderbiest:
Currently implementing "forced loading of addons via GET parameters" (thanks CIGALsace)

fvanderbiest:
Example URL to load 2 addons: http://localhost:8287/mapfishapp/?addons=magnifier_0,annotation_0

landryb:
Nice - how will it fit with the 'have addons depending on the profile or the user or the context' ? Still depending on OWSContext support ?

fvanderbiest:
IMO, URL parameters override everything else.
For the rest, we'll see later ;-)

PR #1143, opened 2015-11-30
mapfishapp - force addon loading via GET parameter

fvanderbiest:
See #132


fvanderbiest:
Please note that http://sdi.georchestra.org/mapfishapp/?addons=fullscreen_0&fullscreen=true cannot work (unless in "toolbars", aka "fake fullscreen" mode) since requestFullScreen can only be initiated by a user gesture.

PR #1145, opened 2015-12-02
[Mapfishapp] Enable to POST a search request

jusabatier:
In order to link Georchestra with an other solution (openADS), I need to allow MF to load parcels or directory from some layers, determined by their ID.

In order to do that I suggest this implementation of a "search" parameter in the POST arguments.
I treated this parameter using the logic of the querier already present in MF.

So if you provide a params like this one in POST data : 

```
"search":{
    "owsurl":"https://georchestra.mydomain.org/geoserver/ows",
    "cql_filter": "id = '2015430157000BO0224'",
    "featurens":"https://georchestra.mydomain.org/geoserver/cadastre",
    "typename":"geo_parcelle"
}
```

It will display at startup the search toolbar with dataset corresponding to the CQL filters and recenter the map on the resultset.

I think it can be very useful for a lot of externals applications which want to show a particular data in Georchestra.


jusabatier:
Example here : [Link](https://opendata.agglo-lepuyenvelay.fr/soap/elyxmapper?KEY=cadastre:parcelle_view*/430157000BC0046&KEY=cadastre:parcelle_view*/430157000BC0227&MAX_SCALE=500)

This link use a wrapper which convert params in JSON and POST them to MFapp. (Used by OpenADS to parse the Elyx SIG URLs because Georchestra is not yet implemented)

pmauduit:
Looks good server-side, but I'll leave it to another reviewer concerning client-side / js code modifications.

jusabatier:
@fvanderbiest  : Is it possible to validate and merge this functionality with the 16.12/17.01 version of georchestra ?

It would be great as it is a dependance for the connector of our ADS solution (OpenADS)

fvanderbiest:
OK, I'll have a second look tomorrow.

mapfishapp/README.md
```diff
@@ -56,10 +56,21 @@ It is also possible to POST a JSON string to the home controller, for instance :
             "layername": "ign:ign_bdtopo_region",
             "owstype": "WMS",
             "owsurl": "http://ids.pigma.org/geoserver/ign/wms"
-        }]
+        }],
+        "search":{
+    		"owsurl":"http://ids.pigma.org/geoserver/ows",
+    		"cql_filter": "id_dept = 48",
+    		"featurens":"http://ids.pigma.org/geoserver/ign",
+    		"typename":"ign_bdtopo_departement"
```
fvanderbiest: I'm not sure you need to ask for `featurens` here if the external app provides you with the fully qualified layer name (`nsalias:typename`).
That would also be more in line with the other "layer" inputs above.

mapfishapp/src/main/webapp/app/js/GEOR_mapinit.js
```diff
@@ -340,6 +350,67 @@ GEOR.mapinit = (function() {
...
+            && initSearch.hasOwnProperty('owsurl') && initSearch.hasOwnProperty('cql_filter') ) {
+            var record = {
+                typeName: initSearch.typename,
+                featureNS: initSearch.featurens,
+                owsURL: initSearch.owsurl,
+                WFSversion : "1.1.0"
+            }
+            var filter = (new OpenLayers.Format.CQL()).read(initSearch.cql_filter);
+            
+            var attStore = GEOR.ows.WFSDescribeFeatureType(record, {
```
fvanderbiest: All this code might deserve a bit of factorization, since it looks like it is borrowed from the querier.

mapfishapp/src/main/webapp/app/js/GEOR.js
```diff
@@ -410,7 +410,28 @@ Ext.namespace("GEOR");
...
+                    });
+                    tab.populate({
+                        features: options.features,
+                        // here we do have a valid model (got from describeFeatureType)
+                        model: options.model
+                    });
+                    southPanel.insert(southPanel.items.length-1, tab);
+                    southPanel.setActiveTab(tab);
+                    southPanel.expand();
+                    tab._zoomToFeatures(options.features);
```
jusabatier: What do you think I should use instead ?

mapfishapp/src/main/webapp/app/js/GEOR_mapinit.js
```diff
@@ -340,6 +350,67 @@ GEOR.mapinit = (function() {
...
         });
+        
+        // support for JSON search parameter
+        if( initSearch.hasOwnProperty('typename') && initSearch.hasOwnProperty('featurens') 
+            && initSearch.hasOwnProperty('owsurl') && initSearch.hasOwnProperty('cql_filter') ) {
+            var record = {
+                typeName: initSearch.typename,
+                featureNS: initSearch.featurens,
+                owsURL: initSearch.owsurl,
+                WFSversion : "1.1.0"
```
jusabatier: No idea, should I replace it with 2.0.0 or just remove the parameter ?

fvanderbiest: I don't know if there are some default values enforced somewhere ...

mapfishapp/README.md
jusabatier: I don't think it's good to set prefix in typeName, as there is a field for setting this (featurePrefix).

When I developed an addon for search accross different zones/referentials, I have to set the featurePrefix param, because if I use only the featureType, in the raquest I have "feature:referentiels:communes", because it append featureNS with default prefix value.

It happen with the OpenLayers.Protocol.WFS_1_1_0 read method, but I haven't tested with 2_0_0

But effectively the featureNS can be removed if we have a featurePrefix provided

mapfishapp/src/main/webapp/app/js/GEOR_mapinit.js
jusabatier: I think default value is 1.0.0 : https://github.com/openlayers/ol2/blob/master/lib/OpenLayers/Protocol/WFS.js#L85

So I will replace with 2.0.0 for use the latest version is it OK ?

mapfishapp/src/main/webapp/app/js/GEOR_mapinit.js
jusabatier: I borrowed only a little part (3 lines) from the querier for get the geometryName do you think it really need to define a function ?

The callbacks I use are not the same as the querier.

mapfishapp/src/main/webapp/app/js/GEOR.js
jusabatier: In order to zoom on a set of features, the method is the same a the one in _zoomToFeatures : calculate total bounds of features or get dataExtent and zoom on it.

Maybe we can set this method as a public one ?

mapfishapp/src/main/webapp/app/js/GEOR_mapinit.js
fvanderbiest: Reading again the code behind GEOR.ows.WFSDescribeFeatureType, you should not have to provide a version number, it we be automatically negociated, since extractFeatureNS is true.

You also do not need featureNS if you provide a fully qualified typename (ie with nsalias included)

PR #1158, opened 2015-12-21
Bumping to new version of GeoServer (2.8.1) and GeoFence (3.0)

pmauduit:
The purpose of this PR is to integrate evolutions proposed in GeoServer / GeoFence upstream into geOrchestra.

This PR is mainly related to pom / doc / submodule modifications and does not contain so much code, since everything is already published into the GeoServer / GeoFence fork of geOrchestra (see https://github.com/georchestra/geoserver/tree/2.8.1-layergroups-georchestra for geoserver, https://github.com/georchestra/geofence/tree/3.0-georchestra for geofence standalone).

Note: there is still a buggy page in the GS interface (the "about geoserver" page is broken).

PR #1159, opened 2015-12-21
GeoServer 2.8.1 - GeoFence 3.0 integration

pmauduit:
Same as https://github.com/georchestra/georchestra/pull/1158 with remarks taken into account and branches correctly named ...

### DIFF
--- a/geoserver/README.md
+++ b/geoserver/README.md
@@ -1,0 +1,47 @@
+# geOrchestra specific GeoServer build documentations
+
+## GeoServer
+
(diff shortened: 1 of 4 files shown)


fvanderbiest:
LGTM !

fvanderbiest:
Do you think you could update sdi.georchestra.org with the new GeoServer ?

pmauduit:
I cannot see any reason why we could not, but we may wait for the CI to pass on build.g.o

PR #1168, opened 2016-01-05
Change temporary group name (remove underscore) and change ui message

Vampouille:


pmauduit:
The only remark is that the message in the `alert()` cannot be translated (easily) ? IIRC there are some other places in the UI where the strings are not translated.

Looks good !

Vampouille:
If we want to provide translation for messages in js files, I think we need to move all messages to a jsp file.

ldapadmin/src/main/filtered-resources/WEB-INF/views/privateUi.jsp
```diff
@@ -110,6 +111,7 @@ try {
     var GEOR_config = {
         publicContextPath: "<%= contextPath %>"
     };
+    var VIRTUAL_TEMPORARY_GROUP_NAME = "<%= GroupsController.VIRTUAL_TEMPORARY_GROUP_NAME %>";
```
fvanderbiest: Better use the above config object instead:
```js
GEOR_config = {
          publicContextPath: "<%= contextPath %>",
          virtualTempGroupName: <%= GroupsController.VIRTUAL_TEMPORARY_GROUP_NAME %>
};
```

ldapadmin/src/main/java/org/georchestra/ldapadmin/ws/backoffice/groups/GroupsController.java
```diff
@@ -55,7 +55,7 @@
 
 	private static final String USER_NOT_FOUND = "user_not_found";
 
-	private static final String VIRTUAL_TEMPORARY_GROUP_NAME = "TEMPORARY_USER";
+	public static final String VIRTUAL_TEMPORARY_GROUP_NAME = "TEMPORARY-USER";
```
fvanderbiest: Or just "TEMPORARY" ?

fvanderbiest:
> If we want to provide translation for messages in js files, I think we need to move all messages to a jsp file.

ldapadmin is not i18n right now. This will be adressed in a second time. 
So, for me, your contribution is fine regarding the error message.

fvanderbiest:
You merged too soon !

Vampouille:
I just commit on master (5b178320d758073dbca768a1f429c685de15545c) :
  * move VIRTUAL_TEMPORARY_GROUP_NAME var in GEOR_config object
  * rename group to : Temporary

PR #1172, opened 2016-01-08
Remove virtual group from group list in user edit page

Vampouille:


fvanderbiest:
Looks good, but our coding style requires that you add parentheses for a if statement.

fvanderbiest:
To be fixed, thanks.

Issue #1123, opened 2015-11-16
Migrate to full HTTPS by default


fvanderbiest:
With the advent of https://letsencrypt.org/ I see no reason to keep unsecured transfers to the browser by default.
Proposal: migrate everything to https by default, including the documentation.

landryb:
so much +1, even if i hate the complexity nightmare of TLS :)

fphg:
You mean : sdi.georchestra.org ? Or geOrchestra instances ?

fvanderbiest:
I mean: the geOrchestra install documentation + template configuration should target an HTTPS setup by default for 15.12. Having a mixed HTTP/HTTPS setup will still be possible, but will require more changes.

Side-effect: sdi.georchestra.org should also switch to https.

fphg:
Le 19/11/2015 09:39, François Van Der Biest a écrit :
> I mean: the geOrchestra install documentation + template configuration
> should target an HTTPS setup by default for 15.12. Having a mixed
> HTTP/HTTPS setup will still be possible, but will require more
> changes.
> 
> Side-effect: sdi.georchestra.org should also switch to https.

While I'm convinced of https interest, I still find it amazingly 
difficult to (correctly) set up. The IDS may move to full encrypted 
streams but what about the third party clients ? Do referers have to 
switch ?


> 
> ---
> Reply to this email directly or view it on GitHub:
> https://github.com/georchestra/georchestra/issues/1123#issuecomment-157989841

-- 
Fabrice Phung

fvanderbiest:
I'd say, it depends on the clients. 
Some of them will follow the 302 redirection to https without raising an eyebrow, some probably won't...

fvanderbiest:
See also https://github.com/georchestra/geoserver_minimal_datadir/commit/ee204706947ff134e5da28c921e4cfc79b4a320b

fvanderbiest:
Done. There are probably minor squirks here and there, but we'll fix them as we go.

landryb:
So i'm all for HTTPS & letsencrypt, but if i try to harvest https://sdi.georchestra.org/geonetwork/ from a standalone GN3 instance, when fetching sources it immediately fails with (request from the proxy):
```
Some unexpected error occurred. Error text was: Received fatal alert: handshake_failure
```
and the traceback in catalina.out
```
javax.net.ssl.SSLHandshakeException: Received fatal alert: handshake_failure
[...]

PR #1175, opened 2016-01-11
Switch to SSL by default

fvanderbiest:
Proposal for georchestra/georchestra#1123

Updated documentation comes with.

Work in Progress, do not merge for now

### DIFF
--- a/doc/setup/apache.md
+++ b/doc/setup/apache.md
@@ -1,16 +1,8 @@
 # Setting up Apache
 
-Basically, there are two options here:
- 1. either your whole SDI is all-HTTPS
- 2. or it is mostly HTTP, but some parts of it use HTTPS (typically CAS & LDAPadmin webapps)
+We assume here that SSL is used for the whole SDI.  
+Having a partial SSL/non-SSL setup should work, but will require several changes in your configuration.
 
-Option 1 is achieved through the following apache configuration:
-```
-RewriteCond %{HTTPS} off 
-RewriteRule (.*) https://%{HTTP_HOST}%{REQUEST_URI} 
-```
-
-Option 2 is described in detail below.
 
 ## Modules setup
 
@@ -61,6 +53,8 @@ In ```/etc/apache2/sites-available/georchestra.conf```:
     CustomLog /var/www/georchestra/logs/access.log "combined"
     Include /var/www/georchestra/conf/*.conf
     ServerSignature Off
+    RewriteCond %{HTTPS} off 
+    RewriteRule (.*) https://%{HTTP_HOST}%{REQUEST_URI} 
 </VirtualHost>
 <VirtualHost *:443>
     ServerName georchestra.mydomain.org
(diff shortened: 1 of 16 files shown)


doc/setup/apache.md
```diff
@@ -377,63 +305,8 @@ ProxyPass /mapfishapp/ http://localhost:8180/mapfishapp/
...
-    -in /var/www/georchestra/ssl/georchestra.csr \
-    -signkey /var/www/georchestra/ssl/georchestra.key \
-    -out /var/www/georchestra/ssl/georchestra.crt
-```
-
-* Restart the web server:
-```
-sudo service apache2 restart
-``` 
+An SSL certificate is absolutely required, at least for the CAS module.
```
pmauduit: Actually, it seems that new version of CAS server allows login / ticket validation in plain HTTP (which I'm pretty sure was not the case with older versions). In the current docker composition, I don't use SSL at all and it works (Le Puy-en-Velay also confirmed they don't use HTTPS).

fvanderbiest: I would not recommend this kind of setup, since credentials would transit in plain text...

mapfishapp/src/deb/resources/etc/georchestra/mapfishapp/print/config.yaml
```diff
@@ -37,7 +37,7 @@ scales:
   - 559082264
 
 # see https://github.com/georchestra/georchestra/issues/362
-brokenUrlPlaceholder: http://sdi.georchestra.org/public/transparent.gif
+brokenUrlPlaceholder: https://sdi.georchestra.org/public/transparent.gif
```
pmauduit: Just as a reminder:

- This can cause problems with old JVM (the one used in pigma is too old for certain HTTPS website - https://github.com/camptocamp/georchestra-pigma-configuration/issues/215 (private))

- More relevant point (since the doc mentionned letsencrypt above): The letsencrypt certificates might not be trusted by the JVM: http://stackoverflow.com/questions/34110426/does-java-support-lets-encrypt-certificates

pmauduit:
Looks good anyway

mapfishapp/src/deb/resources/etc/georchestra/mapfishapp/print/config.yaml
fvanderbiest: Good point.
Maybe we shouold document how to import their certificate into the JVM truststore ?

pmauduit: Something funny about this: In my docker composition, since I use the latest available official image from jetty (which derives from official openjdk one), the Java version seems recent enough :) (But I have some problems with the JVM used on my development environment).

doc/setup/apache.md
landryb: I think what pierre meant was that the 'internal' communications between webapps and cas (only on localhost/internal networks) didnt *require* a self-generated https cert like it's done now. Of course we want the frontend communications with the clients to be https.

mapfishapp/src/deb/resources/etc/georchestra/mapfishapp/print/config.yaml
landryb: > Good point.
> Maybe we should document how to import their certificate into the JVM truststore ?

```
openssl s_client -connect site.fqdn:443 </dev/null 2>/dev/null|openssl x509 -out /tmp/site.fqdn.crt
keytool -importcert -alias site.fqdn -file /tmp/site.fqdn.crt -keystore /path/to/keystore -storepass keystore.pass -noprompt

```
The first line is not needed if you already have the crt, but i used this to import remote certs for geonetwork harvesting over https

landryb: > More relevant point (since the doc mentionned letsencrypt above): The letsencrypt certificates might not be trusted by the JVM: http://stackoverflow.com/questions/34110426/does-java-support-lets-encrypt-certificates

Argh. That means if we want everyone to be able to communicate (ie harvest, use wxs services..) with https-only-with-letsencrypt-cert instances, we have to document/explain how to import LE's root cert in their tomcat's keystore, since we cant expect/rely on having java8 ?

PR #1178, opened 2016-01-13
Docker

fvanderbiest:
Work in progress creating an easy-to-setup docker environment with maven.

Compared to https://github.com/georchestra/docker-ci this new setup has several advantages:
 * no need to wait for the CI to produce artifacts to use them, since the maven build process takes care of  creating the docker images (thanks to spotify's docker-maven-plugin).
 * it is directly integrated into the main source code repository

Progress tracker:
- [x] sample ldap
- [x] sample database
- [x] analytics
- [x] cas
- [x] catalogapp
- [x] downloadform
- [x] extractorapp
- [x] geonetwork
- [x] geoserver
- [x] geofence
- [x] geowebcache
- [x] header
- [x] ldapadmin
- [x] mapfishapp
- [x] security-proxy

<p align="center"></img></p>


fvanderbiest:
Works like a charm !

pmauduit:
- [ ] analytics
- [ ] cas
- [ ] catalogapp
- [ ] downloadform
- [ ] extractorapp
- [ ] geonetwork
- [ ] geoserver
- [ ] geowebcache
- [ ] header
- [ ] ldapadmin
- [ ] mapfishapp
- [ ] security-proxy

fvanderbiest:
Pierre, do you think we could live with a ```docker``` branch in the main repo, in which we would merge the changes from master from time to time, in order to trigger image build ? Maybe we could even delete the georchestra/config repository, which is a bit redundant, thus a pain to maintain.

pmauduit:
Why not keeping a "copy" of the datadir into docker/config ? We don't touch the parameter that much and I'll need such a "docker-ready" datadir somewhere in the source tree so that I can generate image for a remote docker instance. My PoV: ok to deprecate / remove the config repository, but having both */src/deb/resources and docker/config sounds reasonable to me.

fvanderbiest:
@Vampouille also suggests we could:
 * keep georchestra/config 
 * get rid of moduleX/src/deb/resources but dynamically generate them during the debian package build cycle

This would address the config duplication issue, but would complicate the docker setup (https://github.com/georchestra/georchestra/commit/be5dbb3f78bdbc9ba2eba38430158ea1392a7abd really sounded nice to me)

pmauduit:
As discussed this morning I am :+1: with maven-copy for auto-generating src/deb/resources (and keeping a georchestra/config repository)

fvanderbiest:
At camptocamp, we were all sold for this last proposal.

pmauduit:
Note: you can now generate the docker images with customized `/etc/georchestra/<module>` for each modules:

1. check out your datadir into `./docker/<myDatadir>`
2. `mvn clean package docker:build -Pdocker --pl ... -Ddatadir=<myDatadir>`

fvanderbiest:
Then we should probably remove the volumes mapped to /etc/georchestra in docker-compose.yml

fvanderbiest:
There's an issue with the database:
```
org.apache.commons.dbcp.SQLNestedException: Cannot create PoolableConnectionFactory (Connection refused. Check that the hostname and port are correct and that the postmaster is accepting TCP/IP connections.)
```

Vampouille:
which webapp is affected by database issue ?

fvanderbiest:
mapfishapp at first sight

fvanderbiest:
"Use new version of ldap image" but I do not see any new image (except maybe on docker hub).
Missing file in your latest commit @Vampouille ?

fvanderbiest:
With 8ebddd7 I reverted the following ```georchestra/base``` dockerfile update
```
ADD https://github.com/georchestra/datadir/archive/docker-official.zip /tmp/georchestra.zip
RUN unzip /tmp/georchestra.zip -d /etc && mv /etc/datadir-docker-official /etc/georchestra
```
because this was already taken care of by the docker maven plugin.

Thinking about it twice, this may ease our task, since this removes the need to update the datadir submodule every time it is updated. What do others think about it ?

Vampouille:
Ldap image is not finished. It missing ldap users and groups. We need to find a way to populate ldap directory just after ldap image start.

About datadir, removing submodule is nice for me ! :+1:

Vampouille:
I just commit last modification for ldap image. Now modules and overlay are loaded before slapd start and ldap database is populated just after slapd start.

Vampouille:
Composition should work with : 
```bash
sudo apt-get install docker
wget https://raw.githubusercontent.com/georchestra/georchestra/dockerize/docker-compose.yml
docker-compose up
```
you also need to install docker-compose : https://github.com/docker/compose/releases

fvanderbiest:
```
$ docker stats georchestra_proxy_1 georchestra_ldapadmin_1 georchestra_geonetwork_1 georchestrcas_1 georchestra_ldap_1 georchestra_mapfishapp_1 georchestra_header_1 georchestra_extractorapp_1 georchestra_geoserver_1 georchestra_database_1 georchestra_geoserver_geodata_1
```

screenshot from 2016-01-31 15-33-42

I'm just surprised by mapfishapp's container memory usage.

fvanderbiest:
Exit codes on CTRL+C
```
georchestra_header_1 exited with code 143
georchestra_cas_1 exited with code 143
georchestra_proxy_1 exited with code 143
georchestra_ldapadmin_1 exited with code 143
georchestra_mapfishapp_1 exited with code 143
georchestra_extractorapp_1 exited with code 143
georchestra_geonetwork_1 exited with code 143
georchestra_database_1 exited with code 0
georchestra_smtp_1 exited with code 137
georchestra_courier-imap_1 exited with code 137
georchestra_webmail_1 exited with code 137
georchestra_ldap_1 exited with code 137
georchestra_geoserver_geodata_1 exited with code 137
georchestra_geoserver_1 exited with code 137
```

Exit-codes above 128 mean that the process died because of a received signal (exitCode = 128 + signalNumber):
 * 137 means the signal was 9 (SIGKILL)
 * 143 means the signal was 15 (SIGTERM)

Vampouille:
Images that exists with code 137 are also slow to stop ~10sec. In those
images main process does not have PID 1, I think docker is trying to kill
pid 1 but pid 1 is waiting for another process to stop (main process), so
docker wait for 10 sec (configurable) and then send a SIGKILL to PID 1.
This can be fixed with some modifications on entrypoint script.

2016-01-31 16:12 GMT+01:00 François Van Der Biest <notifications@github.com>
:

fvanderbiest:
> This can be fixed with some modifications on entrypoint script.

OK, let's do these modifications. Thanks :-)

Vampouille:
About db connection, with composition, webapp should wait for DB. There is a dedicated java class WaitForDB (and WaitForLdap). We can replace WaitForDB class with datasource that use `ComboPooledDataSource` from `com.mchange.v2.c3p0`
This class manage db reconnection : http://www.mchange.com/projects/c3p0/#configuring_recovery

fvanderbiest:
> About db connection, with composition, webapp should wait for DB

Sounds good too.

fvanderbiest:
Docker engine 1.10 removes the need for links in the composition, since containers know each others through DNS rather than /etc/hosts, see https://github.com/docker/libnetwork/pull/841

fvanderbiest:
> Docker engine 1.10 removes the need for links in the composition, since containers know each others through DNS rather than /etc/hosts

But @raphink says we should keep them anyway

raphink:
@fvanderbiest yes, they are useful for backward compatibility, and rancher uses them to generate the stack graph and order container startup (building a dependency graph from the links).

geonetwork
```diff
@@ -1 +1 @@
-Subproject commit 5e375faa1257eb97f2435d501760bd58a4120c7b
+Subproject commit dd1d9d5047cd30d1a449610a72761da3a0841d72
```
fvanderbiest: to be set to 4dac01826

Issue #1177, opened 2016-01-12
proxy - session timeout is too short


fvanderbiest:
User sessions expire after 2 hours.
A better compromise would be 24 hours IMO.

landryb:
+1 - i've been bitten by this several times recently :)

fvanderbiest:
I think it's merged, so closing.

PR #1181, opened 2016-01-15
 proxy - session timeout is too short #1177

vdorut:
Please do not merge . Currently testing on sdi


fvanderbiest:
24 hours instead ?

Issue #1186, opened 2016-01-20
reset userAgents list if no entry was found in user-agents.properties


landryb:
Separate issue, extracted from https://github.com/georchestra/georchestra/issues/1120#issuecomment-159266756

When one doesnt want to use the 'force basic auth when a specific user-agent is used', the natural thing to do would be to comment out all entries in `${georchestra.datadir}/security-proxy/user-agents.properties`.

But if you do this, the code in https://github.com/georchestra/georchestra/blob/master/security-proxy/src/main/java/org/georchestra/security/BasicAuthChallengeByUserAgent.java#L66 will fallback to the user-agent list in https://github.com/georchestra/georchestra/blob/master/security-proxy/src/main/webapp/WEB-INF/applicationContext-security.xml#L79-L87 because all values in ${georchestra.datadir}/security-proxy/user-agents.properties are commented out.

If i read the code that confirms it - my file is not empty but contains no 'valid entries' so `_userAgents.clear()` isnt called - so the default UA list comes from `applicationContext-security.xml`. Adding a "dummy" entry `useragent0.value=.*nonexistentUA.*` to `${georchestra.datadir}/security-proxy/user-agents.properties` workarounds the issue and i can load that damn getcapabilities in Gn3 default viewer. Phew.

So maybe, for this specific issue.... in `BasicAuthChallengeByUserAgent.java:init()` the case when `user-agents.properties` is found but empty should be handled as 'i dont want any UA filtering to happen' and `_userAgents.clear()` should be called ?

landryb:
That's a regression from 92f7b704149df01ba3613d7c5ab9e31b7df16360 - i think the code added during the original datarisation (https://github.com/georchestra/georchestra/commit/c108cd8d787b9a9972a0b5f08ca1f3e666bb2638?w=1#diff-1f6a00d520c7d3d080cda5252e7b05dbR46) was fine wrt this.

PR #1182, opened 2016-01-15
Set UserAgent for extractor requests

Vampouille:

PR #1187, opened 2016-01-20
Reset the _userAgents list if an user-agents.properties file is found…

landryb:
… in the datadir (#1186)

The list wasnt emptied if the file was found but contained no properties
(ie all commented out) and thus the default UA list from applicationContext-security.xml
was still in use.

Untested yet, but the logic should be better. Hope this wont conflict/get in the way of  testing #1182


pmauduit:
Looks good, thanks

Issue #1184, opened 2016-01-20
[security-proxy] basic auth regression ?


landryb:
Using a sec-proxy from 201512221520 (43234ccad1e3bc33276159e4b515f9582d025fe3) i can't use basic auth against geoserver through the sec-proxy. Same issue with sdi.georchestra.org

```
curl 'https://sdi.georchestra.org/geoserver/osm/wms?service=WMS&request=getcapabilities'
-> code 200, OK

curl -u validuser:validpassword 'https://sdi.georchestra.org/geoserver/osm/wms?service=WMS&request=getcapabilities'
-> code 401, NOK

```
The same issue can be tested trying to access `/geoserver/rest` which should require basic auth.

Geoserver returns an error message because it tries to do the basic auth himself and has no configuration to find user/passwords in ldap
```
HTTP Status 401 - No AuthenticationProvider found for org.springframework.security.authentication.UsernamePasswordAuthenticationToken
```

On my instance, i have no user-agents listed in sec-proxy config because i dont want to force a basic auth, but that should be unrelated to the issue we're seeing here:
```
$cat user-agents.properties
useragent0.value=.*nonexistentUA.*
#useragent0.value=.*ArcGIS.*
#useragent1.value=.*uDig.*
#useragent2.value=.*QGIS.*
#useragent3.value=.*Jakarta.*
#useragent4.value=.*Apache-HttpClient.*
```
The first line is here because there's a bug if no uncommented/valid line is found in the file, it's already tracked in a ticket comment but i cant find which one right now.

Looking at sp logs, the received request contains the Authorization cookie, but it isnt transformed into a proper sec-username header to send to geoserver -> the user isn't authenticated against LDAP.

A georchestra instance deployed in the beginning of july (so something around 15.06) doesnt exhibit this issue, and i see this in the log with an authenticated curl request (this log line isnt seen in the 'broken' version)
```
security [INFO] Storing attributes into session for user :testadmin' 
```

landryb:
Looking at the codepath, i'm afraid it might be related to c108cd8d787b9a9972a0b5f08ca1f3e666bb2638 .. digging more.

landryb:
> The first line is here because there's a bug if no uncommented/valid line is found in the file, it's already tracked in a ticket comment but i cant find which one right now.

This was in #1120 and i've extracted it on #1186

landryb:
@pmauduit says that the curl -u test against CIGAL running 14.12 works

landryb:
[...]

PR #1188, opened 2016-01-20
Fix UA matching/Authorization checking flow (#1184)

landryb:
Instead of only looking at the Authorization header if the UA
matched, check it inconditionally and do the UA matching only if
it wasnt sent. Should fix clients which preemptively send the
Authorization header (curl, for example, or if you've disabled
UA matching)

Should fix the logic issue - pending testing here!

### DIFF
--- a/security-proxy/src/main/java/org/georchestra/security/BasicAuthChallengeByUserAgent.java
+++ b/security-proxy/src/main/java/org/georchestra/security/BasicAuthChallengeByUserAgent.java
@@ -84,19 +84,22 @@ public class BasicAuthChallengeByUserAgent extends BasicAuthenticationFilter {
         }
 
         final HttpServletRequest request = (HttpServletRequest) req;
-        final String userAgent = request.getHeader("User-Agent");
-        if (userAgentMatch(userAgent)) {
-            String auth = request.getHeader("Authorization");
+        String auth = request.getHeader("Authorization");
 
-            if ((auth == null) || !auth.startsWith("Basic ")) {
+        /* no valid Authorization header sent preemptively */
+        if ((auth == null) || !auth.startsWith("Basic ")) {
+            final String userAgent = request.getHeader("User-Agent");
+            if (userAgentMatch(userAgent)) {
+                /* UA matched, return a 401 directly to the client */
+                LOGGER.debug("the user-agent matched and no Authorization header was sent, returning a 401.");
                 getAuthenticationEntryPoint().commence(request, (HttpServletResponse) res, _exception);
             } else {
-                LOGGER.debug("Activating filter ...");
-                super.doFilter(req, res, chain);
+                LOGGER.debug("the user-agent does not match, skipping filter.");
+                chain.doFilter(req, res);
             }
         } else {
-            LOGGER.debug("the user-agent does not match, skipping filter.");
-            chain.doFilter(req, res);
+            LOGGER.debug("Authorization header sent in the request, activating filter ...");
+            super.doFilter(req, res, chain);
         }
     }
 


landryb:
Okay, think i've tested all 6 cases and the behaviour looks sane to me:
  * with/without a matching UA
  * with/without sending a preempt auth
  * with/without sending a valid auth

using this UA matching config:
```
$grep -v ^# /etc/georchestra/security-proxy/user-agents.properties 
useragent0.value=.*nonexistentUA.*
useragent1.value=.*curl.*
useragent2.value=.*QGIS.*
```
with the following testcase, on an url (`$req`) doing a `getMap` on a layer in a protected workspace:
```
echo testing with matching UA without auth - should trigger 401 from georchestra
curl -A 'QGIS' -o /dev/null -v $req 2>&1 |egrep '(< HTTP/1.1|< WWW-Authenticate|> User-Agent)'
sleep 10
echo testing with matching UA with valid preempt auth - should trigger a 200 and a valid content
curl -A 'QGIS' -o /dev/null -v -u breuil:myvalidpass $req 2>&1 |egrep '(< HTTP/1.1|< WWW-Authenticate|> User-Agent)'
sleep 10
echo testing with matching UA with invalid preempt auth - should trigger a 401 from georchestra
curl -A 'QGIS' -o /dev/null -v -u nonono:nonono $req 2>&1 |egrep '(< HTTP/1.1|< WWW-Authenticate|> User-Agent)'
sleep 10
echo testing with non matching UA without auth - should trigger a 401 from geoserver
curl -A 'lalala' -o /dev/null -v $req 2>&1 |egrep '(< HTTP/1.1|< WWW-Authenticate|> User-Agent)'
sleep 10
echo testing with non matching UA valid with preempt auth - should trigger a 200 and a valid content
curl -A 'lalala' -o /dev/null -v -u breuil:myvalidpass $req 2>&1 |egrep '(< HTTP/1.1|< WWW-Authenticate|> User-Agent)'
sleep 10
echo testing with non matching UA with invalid preempt auth - should trigger a 401 from georchestra
curl -A 'lalala' -o /dev/null -v -u nonono:nonono $req 2>&1 |egrep '(< HTTP/1.1|< WWW-Authenticate|> User-Agent)'

```
I get this output, which seems to match what we expect, and looking at the debug log, the 'correct' branch is taken in the ifs. Sending an invalid auth correctly triggers org.springframework.security.authentication.BadCredentialsException too.

```
$sh /tmp/testUA.sh  
testing with matching UA without auth - should trigger 401 from georchestra
> User-Agent: QGIS
< HTTP/1.1 401 Unauthorized
< WWW-Authenticate: Basic realm="georchestra"
testing with matching UA with valid preempt auth - should trigger a 200 and a valid content
> User-Agent: QGIS
< HTTP/1.1 200 OK
testing with matching UA with invalid preempt auth - should trigger a 401 from georchestra
> User-Agent: QGIS
< HTTP/1.1 401 Unauthorized
< WWW-Authenticate: Basic realm="georchestra"
testing with non matching UA without auth - should trigger a 401 from geoserver
> User-Agent: lalala
< HTTP/1.1 401 Unauthorized
< WWW-Authenticate: Basic realm="GeoServer Realm"
< WWW-Authenticate: Basic realm="GeoServer Realm"
testing with non matching UA valid with preempt auth - should trigger a 200 and a valid content
> User-Agent: lalala
< HTTP/1.1 200 OK
testing with non matching UA with invalid preempt auth - should trigger a 401 from georchestra
> User-Agent: lalala
< HTTP/1.1 401 Unauthorized
< WWW-Authenticate: Basic realm="georchestra"
```

So i think we should be good to merge. Testing the same script (for 'regressions')against https://sdi.georchestra.org/geoserver/rest (with a valid l/p) i only get 401's but im not sure it's expected or the current redeploys 'broke' this :)

pmauduit:
Also tested from my side (using a different but kindof similar way), looks good enough to go !

landryb:
Fwiw, also tested a getMap request on an unprotected geoserver workspace:
  * not sending any credentials -> valid data, 200 OK
  * sending valid credentials -> valid data, 200 OK
  * sending invalid credentials -> 401 Unauthorized

I think that's the expected behaviour, and it matches what happens 'before' this fix/PR.

PR #1194, opened 2016-01-25
mapfishapp - same behaviour for contexts

### LINKED ISSUE #1193: mfapp - datadir regression on direct linking to WMC
In 15.06, contexts could be stored in the webapp, in contexts/ subdir. Hotlinking to a context was done with `/mapfishapp/?wmc=contexts/my.wmc`.

Since the move to datadir and the `/ws` service handling contexts, they can be stored in `datadir/mapfishapp/context/`, and direct linking is done via `/mapfishapp/?wmc=context/my.wmc` - would be nice to have backwards-compatible url-rewriting to support the previous url parameters, probably with a 302 redirect to `context/` removing the extra s.

Dunno if it has to be fixed serverside in https://github.com/georchestra/georchestra/blob/master/mapfishapp/src/main/webapp/WEB-INF/urlrewrite.xml or in the client-side js parsing the wmc parameter (in https://github.com/georchestra/georchestra/blob/master/mapfishapp/src/main/webapp/app/js/GEOR_mapinit.js#L523 ?).

pmauduit:
Between datadirs and non-datadirs mode, both urls (/context/blah.wmc and
/contexts/blah.wmc) should redirect onto the context controller.

Should fix #1193

Tests: untested at runtime yet

### DIFF
--- a/mapfishapp/src/main/webapp/WEB-INF/urlrewrite.xml
+++ b/mapfishapp/src/main/webapp/WEB-INF/urlrewrite.xml
@@ -5,8 +5,8 @@
 <urlrewrite>
 
     <rule>
-        <from>^/context/(.*)$</from>
-        <to type="forward" last="true">/ws/context/$1</to>
+        <from>^/context(s?)/(.*)$</from>
+        <to type="forward" last="true">/ws/context/$2</to>
     </rule>
     <rule>
         <from>^/map/(\d{19})/?$</from>


pmauduit:
runtime tested (@landryb tested in datadir mode, I tested in ! datadir), both return the same WMC:

- http://localhost:8287/mapfishapp/context/default.wmc
- http://localhost:8287/mapfishapp/contexts/default.wmc

It should not harm to merge

fvanderbiest:
LGTM, thanks !

PR #1203, opened 2016-02-01
enable geopkg, colormap and importer extensions via profiles. fix sld…

landryb:
…service name

sldservice and importer could really be useful.


fvanderbiest:
No objection from my side. 
@pmauduit maybe ?

landryb:
Of course it doesn't "enable" them by default, it just makes them available via -P switches at packaging time :)

fvanderbiest:
Thanks Landry.

Issue #1206, opened 2016-02-02
copyright notice in every file


fvanderbiest:
We should add the copyright notice in every source file, eg:
```
# Copyright (C) 2009-2016 by XXX
#
# This file is part of geOrchestra.
#
# geOrchestra is free software: you can redistribute it and/or modify it under
# the terms of the GNU General Public License as published by the Free
# Software Foundation, either version 3 of the License, or (at your option)
# any later version.
#
# geOrchestra is distributed in the hope that it will be useful, but WITHOUT
# ANY WARRANTY; without even the implied warranty of MERCHANTABILITY or
# FITNESS FOR A PARTICULAR PURPOSE.  See the GNU General Public License for
# more details.
#
# You should have received a copy of the GNU General Public License along with
# geOrchestra.  If not, see <http://www.gnu.org/licenses/>.
```

landryb:
Quick poll - so far only `{analytics,catalogapp,extractorapp,mapfishapp}/src/main/webapp/app/js/GEOR*` have this section, im not even sure all of them have it, and it only mentions the GPL without specifying the version - the Copyright lines dont have the year.

fvanderbiest:
Yes, we are currently missing several items.
In addition, the server-side code misses the copyright notice in every source file.

fvanderbiest:
I think i'm done with it.

landryb:
A full build with all your PRs merged into a master twig went fine here - but i really think you should commit your script to the repo :)

fvanderbiest:
All done !

PR #1221, opened 2016-02-08
gwc copyright notice

fvanderbiest:
see #1206


landryb:
:+1:

Thinking out loud.. no copyright header on the test files too ?

```
geowebcache-webapp/src/test/java/org/geowebcache/security/PreAuthFilterTest.java
geowebcache-webapp/src/test/java/org/geowebcache/util/GeorchestraURLManglerTest.java
```

fvanderbiest:
> Thinking out loud.. no copyright header on the test files too ?

Maybe we should, yes ...
This is less critical IMO.

PR #1227, opened 2016-02-08
security proxy copyright notice

fvanderbiest:
see #1206


landryb:
:+1: 

As a general note for all the PRs.. maybe add the script you used to add the copyright header to the git repo topdir, so that it can be reused periodically when one adds a file and forgets the copyright, and/or to bump every year ? I supppose it checks for files under src/ subdirs, and looks at the extension to check for which type of comment to add in the header ?

fvanderbiest:
> maybe add the script you used to add the copyright header

No scripting here, since I wanted to double check every file.
Migrating the copyright year should be easier...

fvanderbiest:
> when one adds a file and forgets the copyright

This should not happen, we'll be vigilant ;-)

PR #1238, opened 2016-02-10
ldapadmin - publicContextPath same as shared.ldapadmin.contextpath

fvanderbiest:
publicContextPath should be the same as shared.ldapadmin.contextpath, see https://github.com/georchestra/georchestra/blob/15.06/config/shared.maven.filters#L61


fvanderbiest:
To be merged at the same time: https://github.com/georchestra/datadir/pull/9

Issue #1239, opened 2016-02-10
issue with default webapp naming


pmauduit:
Since the docker branch merge, something is broken with the CI on build.g.o:

Every webapps are now generated with {webapp}.war name, instead of {webapp}-{server}.war as it was done before.

expected behaviour:

General case: {webapp}-{server}.war
RPM / deb: {webapp}-generic.war
docker: {webapp}.war

Only the general case is broken.

pmauduit:
Commit which introduced the modification: https://github.com/georchestra/georchestra/commit/19444f94d42f1eb1cb3aad6324e534f776f8592d

pmauduit:
Ok, I found a fix to propose,  waiting for a PR behind: https://github.com/pmauduit/georchestra/tree/issue-1211

Issue #1211, opened 2016-02-08
Get rid of moduleX/src/deb/resources


fvanderbiest:
As we said in https://github.com/georchestra/georchestra/pull/1178#issuecomment-171581789, we should get rid of moduleX/src/deb/resources but dynamically generate these resources during the debian package build cycle.

pmauduit:
WIP: https://github.com/pmauduit/georchestra/tree/issue-1211

- [ ] analytics
- [ ] cas
- [ ] catalogapp
- [ ] downloadform
- [ ] extractorapp
- [ ] geonetwork
- [ ] geoserver
- [ ] geowebcache
- [ ] header
- [ ] ldapadmin
- [ ] mapfishapp
- [ ] security-proxy

landryb:
Fwiw, i like this idea :) one less copy of every configurable file !

pmauduit:
GN done in https://github.com/pmauduit/geonetwork/tree/georchestra-gn3-3.0.x-issue-1211

PR #1240, opened 2016-02-11
Removing src/deb, harmonizing conventions on webapp naming

pmauduit:
This PR addresses several issues:

- #1211 We need to avoid duplication of resources when generating deb|rpm packages, this PR allows to git clone a remote repository (see georchestra/datadir) and use it as default configuration files, so that we do not have duplicated configurations.

- #1239 Since the docker merge, we lost the naming conventions on the webapp, which somehow breaks the CI / autodeploy process on sdi.g.o and also makes the documentation on webapp generation obsolete, this PR restores the previous accepted conventions:

if mvn clean package without server variable provided => webapp-template.war
if mvn clean package -Dserver=sdi => webapp-sdi.war
if mvn clean package -PdebianPackage => webapp-generic.war
if mvn clean package -PrpmPackage => webapp-generic.war
if mvn clean package -Pdocker => webapp.war

We might remove the -generic suffix in case of packaging, but I think we could accept the current convention for the release, and maybe improve this later on (the "-generic" convention is present since the begining of packaging introduction in geOrchestra).

Tests:
Manual tests since this hits the build system and we do not have (yet ?) unit testing system for the compilation process. For each affected webapps, the following have been checked:

webapp naming on mvn clean package
webapp naming on mvn clean package -Dserver=sdi
webapp naming on mvn clean package -PdebianPackage
webapp naming on mvn clean package -PrpmPackage
webapp naming on mvn clean package -Pdocker
Check generated debian package (ensures all the config files as well as the webapp-generic.war are included)
Check generated rpm package (same)
Check generated docker image by launching it

TODO after merge: realigning on georchestra/geonetwork submodule (once the other PR related to this one has been merged there).


fvanderbiest:
Thanks a bunch Pierre, this is excellent.

pmauduit:
ok done

PR #1250, opened 2016-02-13
postgresql docker composition - adding geofence

pmauduit:
Even if not used, it might be interesting to prepare the database to be geofence-compatible.

Tests: recreating georchestra/database docker image by hand, using it in a compossition: schema is correctly created.

### DIFF
--- a/postgresql/06-geofence.sql
+++ b/postgresql/06-geofence.sql
@@ -1,0 +1,183 @@
+begin;
+
+create schema geofence;
+
+set search_path to geofence;
+
+create table gf_gfuser (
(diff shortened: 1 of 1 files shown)


fvanderbiest:
LGTM, thanks Pierre.

For the record, here's what I had used for GF3 during my tests: https://gist.github.com/fvanderbiest/fb0a7ee4c3971be4c2fa

pmauduit:
since geofence in the current version creates the tables if they don't exist, I wonder if this PR is actually relevant.

Maybe we should let geofence create its own tables ?

PR #1022, opened 2015-07-24
[Mapfishapp] Add filter by roles for contexts

jusabatier:

PR #1251, opened 2016-02-14
Add missing config.json files

landryb:
measure/config.json shamelessly stolen from https://github.com/GFI-Informatique/georchestra/blob/measurement/mapfishapp/src/main/webapp/app/addons/measure/config.json but defaulting to enabled: false. Fixes mfapp warnings/errors in datadir mode.

Since @pmauduit commited 9daa1827dd46e03339c96f73c0eeb68bce1cf4bf i'm a bit lost wrt addons/contexts configuration in non-datadir mode, is it still valid to use ADDONS and CONTEXTS in GEOR_custom.js or those should be dropped ? (i can do it in this PR). https://github.com/georchestra/georchestra/pull/1022 also got broken in datadir mode, we have to find a new way to configure it.

I'd also amend all the addons README.md because so far they only mention editing GEOR_custom.js, which wont work (at least for datadir mode)

Shouldnt we move addons to georchestra-datadir repo, since that's something end-users will often configure (okay, only the config.json files maybe) ? We'd lose history though..

I need to test if an addon works when its code/img/css is kept in the webapp but the config.json file is found in the datadir, if it overrides the one in the webapp.

### DIFF
--- a/mapfishapp/src/main/webapp/app/addons/fullscreen/config.json
+++ b/mapfishapp/src/main/webapp/app/addons/fullscreen/config.json
@@ -1,0 +1,20 @@
+[{
+    "id": "fullscreen_0",
+    "enabled": false,
+    "name": "Fullscreen",
+    "title": {
+        "en": "Fullscreen",
+        "es": "Pantalla Completa",
+        "fr": "Plein écran",
+        "de": "Vollbild"
+    },
+    "options": {
+        "toolbars": true
+    },
+    "description": {
+        "en": "Make the map go fullscreen",
+        "es": "Mostrar el mapa en pantalla completa",
+        "fr": "Voir la carte en plein écran",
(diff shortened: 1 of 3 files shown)


pmauduit:
> Since @pmauduit commited 9daa182 i'm a bit lost wrt addons/contexts configuration in non-datadir mode, is it still valid to use ADDONS and CONTEXTS in GEOR_custom.js or those should be dropped ? (i can do it in this PR).

The blocks in GEOR_custom are unused now, everything goes through the addons / contexts controllers in both cases (datadir and non-datadir mode). The release notes have been updated about it: https://github.com/georchestra/georchestra/blob/master/RELEASE_NOTES.md#upgrading

fvanderbiest:
LGTM, thanks !

PR #1264, opened 2016-02-18
Fix missing sequence

Vampouille:
Fix mismatch between sequence name in java code and sequence created by
postgres bigserial field


fvanderbiest:
LGTM

Issue #1265, opened 2016-02-18
Date mismatch between debian:package and debian:publish target on jenkins


Vampouille:
When publishing deb package, jenkins tries to publish deb file with current date and time. It is not working when building at 13:05:59 and publishing at 13:06:02

pmauduit:
We should find a file that is modified when the repository is updated, and set the following option in each debianPackage profile:

snapshotRevisionFile	deb.package.snapshotRevFile	When building snapshot packages, append the modification time of this file to the version number.	File	The default is to use the system time.

see http://debian-maven.sourceforge.net/usage.html.

pmauduit:
Tried locally: it works well using:

```
+              <snapshotRevisionFile>${project.build.directory}</snapshotRevisionFile>
```

in the debianPackage configuration profile. PR to come ...

pmauduit:
Should be solved now, closing.

pmauduit:
It reappeared:

```
[WARNING] dpkg-deb: building package `georchestra-downloadform' in `/home/jenkins/jobs/georchestra-pipelines/jobs/georchestra-sandbox-pipeline/workspace/downloadform/target/georchestra-downloadform_16.12+201702031428+sandbox5-1_all.deb'.
[INFO] 
[INFO] --- debian-maven-plugin:1.0.6:deploy (default-cli) @ downloadform ---
[INFO] Skipping deployment of non-existent package: /home/jenkins/jobs/georchestra-pipelines/jobs/georchestra-sandbox-pipeline/workspace/downloadform/target/georchestra-downloadform_16.12+201702031515+sandbox5-1_all.deb
```

pmauduit:
Seems fixed: It was missing a clean target (to remove & recreate target/), so the same date (modified date for target/ subdirectory) was used again and again.

PR #1266, opened 2016-02-18
debian packaging - debian:deploy can fail

pmauduit:
If the `debian:build` target generates a debian package at `minute n` and the `deb:deploy` target tries to publish the package on the repository at `minute n+1`, the deployment does not succeed silently because the debian filename is calculated (in case of snapshot builds) given the current time (with a precision of a minute).

Using a generated modification time of a file (as `target/`) shall solve the issue.

This should address issue #1265

Tests: manual, but given the triviality of the modification, it should not harm the CI.


fvanderbiest:
LGTM, please merge

pmauduit:
Successful on the CI :+1: :fireworks:

Vampouille:
:-)

PR #1278, opened 2016-02-23
mapfishapp - deactivate tests if travis is detected

pmauduit:
Title says it all.


fvanderbiest:
LGTM, but how will we know that travis has restored a working JVM ?

pmauduit:
We won't :)

fvanderbiest:
Hmm, might be good to keep it up in a fork, just in case ?

pmauduit:
> Hmm, might be good to keep it up in a fork, just in case ?

:+1:

Issue #1279, opened 2016-02-23
analytics / ogc-server-stats - no logs are shown


fvanderbiest:
Analytics is empty, so either it is broken or the logger (ogc-server-statistics) is...

pmauduit:
There are several problems:

- Logging on sdi.g.o seems broken since 2016-02-09 ; this is due to mismatch between slf4j and log4j libraries mixed together in the webapp.
- there actually is an issue with the db migration: if you look into tomcat-02's catalina.out:

```
[http-127.0.0.1-8280-5] ERROR org.georchestra.analytics - Caught exception while executing service: 
org.postgresql.util.PSQLException: ERREUR: droit refusé pour la relation ogc_services_log
```

fvanderbiest:
Indeed, missing 
```
grant all on ogcstatistics.ogc_services_log to "www-data";
```

fvanderbiest:
I'm handling the db migration, can you investigate the mismatch between slf4j and log4j libraries ?
Thanks @pmauduit

Vampouille:
Maybe, we can change owner of georchestra database to `www-data`. This could simplify installation and allow webapp to update db schema.

fvanderbiest:
Still an issue with sdi.geor.org: figures do not evolve when consuming local OGC services.

Vampouille:
```
georchestra=> INSERT INTO ogcstatistics.ogc_services_log (user_name , date , service , layer , request , org , roles) VALUES ('anomymous', NOW(), 'WMS', 'test', 'getmap', '', NULL);
ERREUR:  droit refusé pour la séquence ogc_services_log_id_seq1
```
Il manque les droits de lecture sur la séquence

Il y a peut être aussi un soucis sur la séquence car il y a 2 séquences : 

* ogcstatistics.ogc_services_log_id_seq1 (last_value=1)
* ogcstatistics.ogc_services_log_id_seq (last_value=2890211)

La définition actuelle de la table utilise ogcstatistics.ogc_services_log_id_seq1 ce qui devrait générer des `id` en double

Vampouille:
This second sequence is generated by upgrade procedure, so we need to modify new `ogc_services_log` to use old sequence and not create a new one.

Vampouille:
```
ALTER TABLE ogcstatistics.ogc_services_log
   ALTER COLUMN id SET DEFAULT nextval('ogcstatistics.ogc_services_log_id_seq'::regclass);
```
[...]

PR #1280, opened 2016-02-24
analytics - fixing logging

pmauduit:
Adding slf4j-log4j12 as a direct dependency. slf4j is introduced via transitive dependency with spring-data-commons.

Should address #1279 (logging part).

### DIFF
--- a/analytics/pom.xml
+++ b/analytics/pom.xml
@@ -166,6 +166,11 @@
       <version>1.4</version>
     </dependency>
     <dependency>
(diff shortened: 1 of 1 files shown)


pmauduit:
Tests: runtime in a local tomcat. /tmp/analytics.log is created, and a hit on the webapp produces logs.

fvanderbiest:
Nice, thanks Pierre. Please merge.

Issue #1277, opened 2016-02-23
analytics - ClassNotFoundException: org/apache/xalan/processor/TransformerFactoryImpl


fvanderbiest:
```
java.lang.ClassNotFoundException: org/apache/xalan/processor/TransformerFactoryImpl
    java.lang.Class.forName0(Native Method)
    java.lang.Class.forName(Class.java:278)
    javax.xml.transform.FactoryFinder.getProviderClass(FactoryFinder.java:125)
    javax.xml.transform.FactoryFinder.newInstance(FactoryFinder.java:182)
    javax.xml.transform.FactoryFinder.find(FactoryFinder.java:250)
    javax.xml.transform.TransformerFactory.newInstance(TransformerFactory.java:101)
    org.springframework.http.converter.xml.SourceHttpMessageConverter.<init>(SourceHttpMessageConverter.java:80)
```

I guess related to eg: https://github.com/georchestra/georchestra/blob/master/mapfishapp/src/main/java/org/georchestra/mapfishapp/ws/classif/SLDClassifier.java#L129-L141

pmauduit:
Yup a quick win might to add xalan as a dependency in analytics ...

pmauduit:
I think a better way is to remove the code you pointed and use a more convenient method to set the parser for the analytics scope (not polluting the other webapps), following the documentation:

https://docs.oracle.com/javase/7/docs/api/javax/xml/transform/TransformerFactory.html

> Use the javax.xml.transform.TransformerFactory system property.
> Use the properties file "lib/jaxp.properties" in the JRE directory. This configuration file is in standard java.util.Properties format and contains the fully qualified name of the implementation class with the key being the system property defined above. The jaxp.properties file is read only once by the JAXP implementation and it's values are then cached for future use. If the file does not exist when the first attempt is made to read from it, no further attempts are made to check for its existence. It is not possible to change the value of any property in jaxp.properties after it has been read for the first time.
> Use the Services API (as detailed in the JAR specification), if available, to determine the classname. The Services API will look for a classname in the file META-INF/services/javax.xml.transform.TransformerFactory in jars available to the runtime.
> Platform default TransformerFactory instance.

PR #1283, opened 2016-02-24
mapfishapp - using a specific transformer

pmauduit:
Related to issue #1277

Tests:
- Testsuite OK
- Runtime OK (seems to fail after the getSLD() call because of misconfigured database)

TODO
Would deserve a more thourough check in a servlet container with multiple webapps


fvanderbiest:
LGTM, thanks Pierre.
Merging and we'll backport to 15.12 later if relevant.

PR #1287, opened 2016-02-29
specify old sequence as default value for primary key

Vampouille:
This avoid creation of a new sequence when migrating to 15.12


fvanderbiest:
Indeed, thanks Julien !

PR #1298, opened 2016-03-04
[Mapfishapp] Add WMC's roles to datadir

jusabatier:
As seen in this PR : https://github.com/georchestra/georchestra/pull/1022

The datadir config load CONTEXTS directly from wmc files, but those files haven't any field for ACL.

This patch load allowed roles from an XML file with the same name as the WMC one.

For example, for example.wmc the loaded XML will be example.xml and have to contain : 
```
<?xml version="1.0" encoding="UTF-8"?>
<AllowedRoles>
    <Role>ROLE_ADMINISTRATOR</Role>
    <Role>ROLE_SER_URBANISME</Role>
    <Role>ROLE_SER_INGENIERIE</Role>
    <Role>ROLE_PERM_COMMUNES</Role>
</AllowedRoles>
```

If the file doesn't exist, all roles will be enabled.

Maybe add this example XML file in the datadir README.md ?

### DIFF
--- a/mapfishapp/src/main/java/org/georchestra/mapfishapp/ws/ContextController.java
+++ b/mapfishapp/src/main/java/org/georchestra/mapfishapp/ws/ContextController.java
@@ -79,6 +79,8 @@ public class ContextController implements ServletContextAware {
         // image
         // defaulting to default.png
         String image = "context/image/default.png";
+        // roles
+        JSONArray roles = new JSONArray();
 
         File imagePath = new File(pathCtx, "images");
         if (! imagePath.isDirectory()) {
@@ -92,6 +94,30 @@ public class ContextController implements ServletContextAware {
                 }
             }
         }
+        
+        File rolePath = new File(pathCtx);
+        Collection<File> files = FileUtils.listFiles(rolePath, new String[] { "XML", "xml" }, false);
+        for (File curRoleFile : files) {
+            if (FilenameUtils.getBaseName(curRoleFile.toString()).equalsIgnoreCase(title)) {
+                DocumentBuilderFactory domFactory = DocumentBuilderFactory.newInstance();
+                DocumentBuilder builder = domFactory.newDocumentBuilder();
+                Document doc = builder.parse(curRoleFile);
+                XPath xpath = XPathFactory.newInstance().newXPath();
+                
+                // Parsing roles
+                XPathExpression xpRole = xpath.compile("//AllowedRoles/Role");
+                Object oRole = xpRole.evaluate(doc, XPathConstants.NODESET);
+                
+                if (oRole instanceof NodeList) {
+                    NodeList nl = (NodeList) oRole;
+                    for (int i = 0; i < nl.getLength(); ++i) {
+                        roles.put(nl.item(i).getTextContent());
+                    }
+                }
+                break;
+            }
+        }
+        
         // filename
         String wmcUrl = "context/" + f.getName();
 
@@ -103,6 +129,7 @@ public class ContextController implements ServletContextAware {
         info.put("wmc", wmcUrl);
         info.put("tip", xmlInfos.get("tip").equals("unset") ? title : xmlInfos.get("tip"));
         info.put("keywords", xmlInfos.getJSONArray("keywords").put(title));
+        info.put("roles", roles);
 
         return info;
     }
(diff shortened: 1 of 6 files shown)


mapfishapp/src/main/java/org/georchestra/mapfishapp/ws/ContextController.java
```diff
@@ -92,6 +94,30 @@ private JSONObject getContextInfo(File f) throws Exception {
...
+        Collection<File> files = FileUtils.listFiles(rolePath, new String[] { "XML", "xml" }, false);
+        for (File curRoleFile : files) {
+            if (FilenameUtils.getBaseName(curRoleFile.toString()).equalsIgnoreCase(title)) {
+                DocumentBuilderFactory domFactory = DocumentBuilderFactory.newInstance();
+                DocumentBuilder builder = domFactory.newDocumentBuilder();
+                Document doc = builder.parse(curRoleFile);
+                XPath xpath = XPathFactory.newInstance().newXPath();
+                
+                // Parsing roles
+                XPathExpression xpRole = xpath.compile("//AllowedRoles/Role");
```
pmauduit: you could use `//AlowedRoles/Role/text()` as xpath so that no need to check if it is a nodeset, it would allow you to get the actual value for the role more rapidly.

jusabatier:
On the examples iI found on the web, they do : 

```
XPathExpression expr = xpath.compile("//AllowedRoles/Role/text()");
NodeList nodes = (NodeList) expr.evaluate(doc, XPathConstants.NODESET);
for (int i = 0; i < nodes.getLength(); i++)
    roles.put(nodes.item(i).getNodeValue());
```

So as there is a cast to NodeList, isn't it better to check if it's a NodeList before ?

pmauduit:
> So as there is a cast to NodeList, isn't it better to check if it's a NodeList before ?

To be checked, I am not sure, it might fail for example if you only have one role in your XML doc (as it could return something else than a NodeList or something).

jusabatier:
I tested it with only one role in the xml : 
```
<?xml version="1.0" encoding="UTF-8"?>
<AllowedRoles>
    <Role>ROLE_ADMINISTRATOR</Role>
</AllowedRoles>
```
It works

And also tested with no roles : 
```
<?xml version="1.0" encoding="UTF-8"?>
<AllowedRoles></AllowedRoles>
```
And it works too

pmauduit:
Ok, since it will end up in the mapfishapp codebase, I'd like to make these manual tests as automatic unit-tests for maintenability purposes (or if you care to do it, it's welcome !)

jusabatier:
I have absolutely no ideas how tests works in maven x)

pmauduit:
> I have absolutely no ideas how tests works in maven x)

Just review my previous commit, then ;-)

I left some comments in my code anyway, worth to be discussed:

- https://github.com/jusabatier/georchestra/commit/3fe67a782b531a34a3ea71fb88ead91fee3cdc82#diff-5054db8805c7fa74b0b77678a0ba9c94R131: If the user actually knows the context name, he could have access to it easily, this PR does not add extra checks on the user's role before serving the WMC, is that acceptable ?

- https://github.com/jusabatier/georchestra/commit/3fe67a782b531a34a3ea71fb88ead91fee3cdc82#diff-5054db8805c7fa74b0b77678a0ba9c94R148: we might avoid to loop on each xml files from the directory, considering default.xml should get along with the context default.wmc.

mapfishapp/src/main/java/org/georchestra/mapfishapp/ws/ContextController.java
```diff
@@ -29,6 +29,7 @@
 import javax.servlet.http.HttpServletResponse;
 import javax.xml.parsers.DocumentBuilder;
 import javax.xml.parsers.DocumentBuilderFactory;
+import javax.xml.parsers.ParserConfigurationException;
```
fvanderbiest: Codacy Issue found: [Avoid unused imports such as 'javax.xml.parsers.ParserConfigurationException'](https://www.codacy.com/app/francois-vanderbiest/georchestra/file/2313893969/issues/source?bid=3200181&fileBranchId=3200203#l32)

fvanderbiest: Codacy Issue found: [Avoid unused imports such as 'javax.xml.parsers.ParserConfigurationException'](https://www.codacy.com/app/francois-vanderbiest/georchestra/file/2313893969/issues/source?bid=3200181&fileBranchId=3200203#l32)

pmauduit:
Last comment addressed with https://github.com/georchestra/georchestra/pull/1298/commits/8c5701ace38aa864454680cbfb31586c06320695

pmauduit:
Looks good :)

fvanderbiest:
LGTM, merging.
Thanks a lot for your contribution Julien !

Issue #1313, opened 2016-03-16
[sec-proxy] fix SNI support / upgrade to httpclient 4.3.2 ?


landryb:
Similar to geonetwork/core-geonetwork#1368, our sec-proxy trips on an https server with SNI:
```
security [ERROR] Exception occured when trying to connect to the remote host:
javax.net.ssl.SSLException: hostname in certificate didn't match: <wms.craig.fr> != <ids.craig.fr> OR <ids.craig.fr> OR <craig.fr>
```

We use httpclient 4.2.1 in security-proxy, and if i replace the jar by httpclient 4.3.2 (which fixes/supports SNI) and httpcore-4.3.1 (the versions i have in geonetwork) it's still broken/fails to fetch the getcapabilities.

STR: on sdi.georchestra.org/mapfishapp, try to load https://wms.craig.fr/ortho

landryb:
Same issue with httpclient-4.3.3 & httpcore-4.3.2 copied from extractorapp webapp

landryb:
More contect for the traceback:
```
javax.net.ssl.SSLException: hostname in certificate didn't match: <wxs.craig.fr> != <ids.craig.fr> OR <ids.craig.fr> OR <craig.fr>
        at org.apache.http.conn.ssl.AbstractVerifier.verify(AbstractVerifier.java:231)
        at org.apache.http.conn.ssl.BrowserCompatHostnameVerifier.verify(BrowserCompatHostnameVerifier.java:54)
        at org.apache.http.conn.ssl.AbstractVerifier.verify(AbstractVerifier.java:152)
        at org.apache.http.conn.ssl.AbstractVerifier.verify(AbstractVerifier.java:133)
        at org.apache.http.conn.ssl.SSLSocketFactory.verifyHostname(SSLSocketFactory.java:559)
        at org.apache.http.conn.ssl.SSLSocketFactory.connectSocket(SSLSocketFactory.java:534)
        at org.apache.http.conn.ssl.SSLSocketFactory.connectSocket(SSLSocketFactory.java:401)
        at org.apache.http.impl.conn.DefaultClientConnectionOperator.openConnection(DefaultClientConnectionOperator.java:177)
        at org.apache.http.impl.conn.ManagedClientConnectionImpl.open(ManagedClientConnectionImpl.java:304)
        at org.apache.http.impl.client.DefaultRequestDirector.tryConnect(DefaultRequestDirector.java:611)
[...]

PR #1314, opened 2016-03-16
SP - rewrite HttpClient object instanciations

pmauduit:
The use of deprecated classes was causing SNI for HTTPS to fail. Also includes a version bump of apache-HttpClient (which supports SNI).

Should address #1313 ;

@landryb : care to test it at runtime from your side ? since the code should be called very often when using the whole SDI, it would deserve more tests.

Tests:
- runtime OK
- testsuite OK


fvanderbiest:
LGTM, please merge :-)

landryb:
Works fine to fix the SNI issue here.

PR #1319, opened 2016-03-21
Listen on unix socket

mcanevet:
Required for backup via slapcat

### DIFF
--- a/ldap/docker/start.sh
+++ b/ldap/docker/start.sh
@@ -3,7 +3,7 @@ set -e
 
 /populate_ldap.sh &
 sleep 1
-exec /entrypoint.sh slapd -d 32768 -u openldap -g openldap
+exec /entrypoint.sh slapd -d 32768 -u openldap -g openldap -h ldap:/// ldapi:///
 
 
 


pmauduit:
Looks good, but from what I remember, slapcat will also dump the extra ldap attributes (uuid, ...). What is your way of restoring a ldif backup ? (if ldapadd, it won't work because of some read-only attributes), poke @Vampouille (since we discussed this issue some minutes ago).

Vampouille:
You can filter read-only attributes and restore only writable one with ldapadd. This way we can restore ldap with `slapadd` or `ldapadd`.
:+1:

pmauduit:
ok

landryb:
> You can filter read-only attributes and restore only writable one with ldapadd. This way we can restore ldap with slapadd or ldapadd.

Which option of ldapadd or slapcat is this ? Is it a complex slapcat -a filter, manually excluding all the operational attributes ? So far i'm using
```
slapcat | egrep -v '^(contextCSN|entryCSN|modifiersName|modifyTimestampentryUUID)'
```
but if there's a better way i'm all ears :)

Vampouille:
Unix socket is not needed for backup and following commit improve ldap image :
0cbeb505e3f65e3c1734243bb85bcb04790d6bd8

PR #1320, opened 2016-03-21
remove UUID from code and some wildcard imports of package

Vampouille:


fvanderbiest:
LGTM, but breaking change, will require to inform users already having the 15.12 db in production.

pmauduit:
looks good

Vampouille:
This modification introduce database changes. So if you have already installed 15.12, you will need to change your database schema.

Affected tables : 
* `admin_log` fields : 
  * `admin`
  * `target`
* `admin_emails` fields :
  * `recipient`
  * `sender`

Those columns change from `UUID` type to `character varying(255)`

If you don't care about table contents, you can just drop this 2 tables and restart ldapadmin webapp. Tables will be recreated with correct data type.
If you do care about data, you will need to :
* rename previous fields with `_old` suffix
* create previous fields with type `character varying(255)`
* populate previous fields based on information in `*_old` fields
* drop `*_old` fields

PR #1326, opened 2016-03-23
Fix composition

Vampouille:
This PR fixes several bugs:
* Add waitForDb feature on analytics webapp
* Add missing geoserver volumes in composition
* Add some missing webapp in composition : 
  * analytics
  * catalogapp
  * downloadform
* Fix argument type on SQL function (ogc-stats)


fvanderbiest:
LGTM but would have been nicer if separated PRs.
(looks like a "pot pourri", sorry ;-)

PR #1328, opened 2016-03-23
Persist mapfishapp and extractor volumes in composition

Vampouille:


pmauduit:
Looks good but also missing a volume for geonetwork

pmauduit:
And I don't know if it is relevant to also include in this composition a standalone geowebcache, either case we would need to add it to the compo + define its own volume

PR #1336, opened 2016-03-29
Update geofence database schema

Vampouille:


Vampouille:
WIP, this PR include geofence DB schema, may be not wanted by some setup without geofence.

postgresql/06-geofence.sql
```diff
@@ -82,9 +83,14 @@ create table gf_rule (
   request varchar(255),
   service varchar(255),
   workspace varchar(255),
-  gsuser_id int8,
-  instance_id int8,
-  userGroup_id int8,
+  gsuser_id bigint,
+  instance_id bigint,
+  userGroup_id bigint,
+	ip_high bigint,
```
fvanderbiest: tab ?

postgresql/06-geofence.sql
```diff
@@ -93,6 +99,7 @@ create table gf_rule_limits (
   id int8 not null,
   area public.geometry,
   rule_id int8 not null,
+	catalog_mode character varying(255),
```
fvanderbiest: tab ?

postgresql/06-geofence.sql
```diff
@@ -109,7 +116,9 @@ create table gf_usergroup (
   enabled bool not null,
   extId varchar(255) unique,
   name varchar(255) not null unique,
-  primary key (id)
+  primary key (id),
+	CONSTRAINT gf_usergroup_extid_key UNIQUE (extid),
```
fvanderbiest: tab ?

fvanderbiest:
> WIP, this PR include geofence DB schema, may be not wanted by some setup without geofence.

Right, this has to be discussed first.

pmauduit:
> Right, this has to be discussed first.

My point: it does not cost more money to have it, even if it is unused (we already do the same in the rancher/docker composition and downloadform / analytics, which is not used everywhere)

geoserver/pom.xml
```diff
@@ -11,8 +11,8 @@
   <name>GeoServer 2.x root module</name>
   <properties>
     <maven.test.skip>true</maven.test.skip>
-    <gs.version>2.8.2</gs.version>
-    <gt.version>14.2</gt.version>
+    <gs.version>2.8.3</gs.version>
```
pmauduit: This should come along with an update of the `geoserver/geoserver-submodule/`, pointing onto the `2.8.3-georchestra` branche.

fvanderbiest:
It's not only the db, but also the geoserver_minimal_datadir which is impacted by geofence here ...

pmauduit:
> It's not only the db, but also the geoserver_minimal_datadir which is impacted by geofence here ...

Then it should not be done at the `Dockerfile` level, but into the `pom.xml`, maybe a profile GeoFence which would fetch the expected datadir (ie. using the relevant branche) ? Like in https://github.com/georchestra/georchestra/blob/15.12/mapfishapp/pom.xml#L506-L524

fvanderbiest:
> Then it should not be done at the Dockerfile level, but into the pom.xml, maybe a profile GeoFence which would fetch the expected datadir (ie. using the relevant branche)

:+1:

geoserver/pom.xml
Vampouille: Done with dd1a8767a7531b21beb5955d517797944224a297

Vampouille:
+1 pour le profile maven

pmauduit:
> +1 pour le profile maven

Tu veux que je le fasse ?

Vampouille:
This commit 4a76a58 add some properties for geoserver minimal datadir (branch to checkout from). Build steps for geoserver without geofence : 
```bash
 cd geoserver/geoserver-submodule/src
mvn clean install -DskipTests
cd ../../webapp
mvn clean install docker:build -DskipTests -Pdocker
```
Build steps for geoserver with geofence : 
```bash
cd geoserver/geoserver-submodule/src
mvn clean install -DskipTests -Pgeofence-server
cd ../../webapp
mvn clean install docker:build -DskipTests -Pdocker -DdockerGeoserverDatadirScmVersion=15.12-docker-geofence
```
So this is not a maven profile but a properties with default value. @pmauduit what do you think about this way to build docker images for geofence ?

pmauduit:
> dockerGeoserverDatadirScmVersion=15.12-docker-geofence

You can move the property into the geofence profile at the end of the pom.xml file. Then you can remove the -DdockerGeoserverDatadir... and use:

```
mvn clean install docker:build -Pdocker,geofence
```

geoserver/webapp/pom.xml
```diff
@@ -1050,12 +1074,5 @@
         </plugins>
       </build>
     </profile>
-    <profile>
-      <id>geofence</id>
```
pmauduit: why did you remove the `geofence` profile ? Even if we actually don't need some of the variables here, I think we can override the `dockerGeoserverDatadirScmVersion` here.

Also as we have discussed, why not renaming the default dockerImageName in the geofence profile by

In 15.12:
```
<dockerImageName>georchestra/geoserver:${geofence}-${project.parent.version}
```

In master:

```
<dockerImageName>georchestra/geoserver:${geofence}-latest
```

Vampouille:
Geofence is now working with this PR

pmauduit:
Looks good !

pmauduit:
Just a note though: we have to be aware that when merging into master, we shall update the versions for the docker image tags

Vampouille:
I just pushed a new version of docker image : `georchestra/database` because this PR change SQL file : `06-geofence.sql`

fvanderbiest:
We should add the above [build instructions ](https://github.com/georchestra/georchestra/pull/1336#issuecomment-203350889) in the documentation

PR #1339, opened 2016-03-30
Only listen on loopback interface when populating ldap

Vampouille:


Vampouille:
We need to push this on docker hub

pmauduit:
Looks good (but what is the purpose of limiting the interfaces to the loopback ?)

Vampouille:
If ldap server listen on all interface during "populate" script, WaitForLdap class will consider ldap as installed and populated. With this modification, WaitForLdap will wait (as expected) for end of population script.

fvanderbiest:
> If ldap server listens on all interfaces during "populate" script, WaitForLdap class will consider ldap as installed and populated. With this modification, WaitForLdap will wait (as expected) for end of population script.

Thanks Julien.
It's always a good idea to give some context with a PR.

PR #1340, opened 2016-03-30
Fix permissions on /tmp dir of geoserver docker image

Vampouille:
This PR moves geoserver minimal datadir from /tmp to /var/local/geoserver.orig. This avoids permission modification on /tmp directory by maven when copying ressources.

### DIFF
--- a/geoserver/webapp/src/docker/docker-entrypoint.d/00-geoserver-bootstrap
+++ b/geoserver/webapp/src/docker/docker-entrypoint.d/00-geoserver-bootstrap
@@ -5,7 +5,7 @@ then
     echo 'Datadir already initialized'
 else
     echo 'Initializing datadir from georchestra/geoserver_minimal_datadir'
-    cp -Rax /tmp/geoserver_minimal_datadir/* /var/local/geoserver
-    cp -Rax /tmp/geoserver_minimal_datadir/.git /var/local/geoserver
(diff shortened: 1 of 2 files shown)


geoserver/webapp/src/docker/docker-entrypoint.d/00-geoserver-bootstrap
```diff
@@ -5,7 +5,7 @@ then
     echo 'Datadir already initialized'
 else
     echo 'Initializing datadir from georchestra/geoserver_minimal_datadir'
-    cp -Rax /tmp/geoserver_minimal_datadir/* /var/local/geoserver
-    cp -Rax /tmp/geoserver_minimal_datadir/.git /var/local/geoserver
-    cp -Rax /tmp/geoserver_minimal_datadir/.gitignore /var/local/geoserver
+    cp -Rax /var/local/geoserver.orig/* /var/local/geoserver
+    cp -Rax /var/local/geoserver.orig/.git /var/local/geoserver
+    cp -Rax /var/local/geoserver.orig/.gitignore /var/local/geoserver
```
fvanderbiest: what about `mv /var/local/geoserver.orig /var/local/geoserver` ?

Vampouille: We can move `/var/local/geoserver.orig` but  `/var/local/geoserver.orig` will be recovered if we recreate container.

PR #1355, opened 2016-04-06
SP - re-introducing default target

pmauduit:
Once upon a time (relying on the source code), there was a parameter
"default target" in the SP which allowed to redirect onto a default
servlet container.

Since the defaultTarget in the properties file as well as in the
proxy-servlet.xml somehow disappeared, I think that this parameter is
not used anymore.

This commit proposes to reintroduce it, but modifying slightly its
original purpose: By defining a defaultTarget in your
security-proxy.properties, you can select a default redirection path. By
default, when hitting the SP at its root, eg:

GET http://georchestra.mydomain.org/

will not return a 404 by default anymore, but send a 302 found
(temporary redirect) onto /header/

but defining defaultTarget=/portail/ in your sp.properties file would
redirect by default to http://georchestra.mydomain.org/portail/.

Tests:
- Added a utest but would deserve some integration tests.
- Testsuite OK
- runtime tests OK

### DIFF
--- a/security-proxy/src/main/filtered-resources/WEB-INF/proxy-servlet.xml
+++ b/security-proxy/src/main/filtered-resources/WEB-INF/proxy-servlet.xml
@@ -17,8 +17,8 @@
 
           <property name="checkHealth" value="${checkHealth}"/>
           <property name="maxDatabaseConnections" value="${max.database.connections}"/>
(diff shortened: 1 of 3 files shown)


Vampouille:
Tested and working ! you can merge !

PR #1358, opened 2016-04-12
Add save and load features to MF annotation

jusabatier:
This PR enable to save annotations and reload them after.

It also set the pointRadius property of labels to 0 in order to display only the text label.

Last upgrade is that annotation layer is added to map on addon's show event if necessary, because when load a context, annotation's layer is deleted, and so we can reload it by close and re-open the addon.


fvanderbiest:
Replaced by #1547

PR #1367, opened 2016-05-02
header - force default encoding of jsp to utf-8

pmauduit:
if not specified, under certain circumstances, jsp are sent as encoded in iso-8859-1, but everything in geOrchestra should be considered as utf-8.

Tests: maven jetty plugin: OK
Testsuite: OK (but unrelated to the current modification)


fvanderbiest:
LGTM

PR #1368, opened 2016-05-03
[Mapfishapp] Contexts CSS update

jusabatier:
This rule allows to align thumbs correctly when title is on multiple lines in contexts selection.


fvanderbiest:
Looks good, but is the window size fixed, so that there are always 4 thumbs in a line ?

jusabatier:
No, you're right, the window size is not fixed.

What about the comportement on : https://opendata.agglo-lepuyenvelay.fr/mapfishapp/ ?

The thumbs resize with the window width, so there are always 4 per row.

If it's OK for you I can add the CSS in this PR

jusabatier:
Finally, the use of the new CSS3 flex display property is a better solution I think.

fvanderbiest:
Thanks a lot Julien !

PR #1371, opened 2016-05-10
Add services monitoring

Vampouille:
This new page checks :
 * geonetwork "GetRecord" query from CSW
 * geoserver "GetCapabilities" query from WMS
 * header with "html" keyword

### DIFF
--- a/security-proxy/src/main/java/org/georchestra/security/Proxy.java
+++ b/security-proxy/src/main/java/org/georchestra/security/Proxy.java
@@ -251,6 +251,11 @@ public class Proxy {
         redirectStrategy.sendRedirect(request, response, uriBuilder.build().toString());
     }
 
+    @RequestMapping("/services_monitoring")
+    public void servicesMonitoring(HttpServletRequest request, HttpServletResponse response) throws IOException {
+        (new ServicesMonitoring(this.georchestraConfiguration.loadCustomPropertiesFile("targets-mapping"))).checkServices(request, response);
+    }
+
     @RequestMapping(params = { "login", "url" }, method = { GET, POST })
     public void login(HttpServletRequest request, HttpServletResponse response, @RequestParam("url") String sURL) throws ServletException, IOException {
         redirectStrategy.sendRedirect(request, response, sURL);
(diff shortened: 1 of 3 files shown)


Vampouille:
Work in progress, please do not merge

Vampouille:
This new page checks : 
* geonetwork "GetRecord" query from CSW
* geoserver "GetCapabilities" query from WMS
* header with "html" keyword

fvanderbiest:
Nice, LGTM :-)

fvanderbiest:
More checks could be implemented in here later on ... (mapfish print, extractorapp services, etc)

security-proxy/src/main/java/org/georchestra/security/Proxy.java
```diff
@@ -251,6 +251,11 @@ public void login(HttpServletRequest request, HttpServletResponse response) thro
         redirectStrategy.sendRedirect(request, response, uriBuilder.build().toString());
     }
 
+    @RequestMapping("/services_monitoring")
+    public void servicesMonitoring(HttpServletRequest request, HttpServletResponse response) throws IOException {
+        (new ServicesMonitoring(this.georchestraConfiguration.loadCustomPropertiesFile("targets-mapping"))).checkServices(request, response);
```
pmauduit: It will probably only work with georchestra.datadir mode

security-proxy/src/main/java/org/georchestra/security/ServicesMonitoring.java
```diff
@@ -0,0 +1,114 @@
...
+        if(success)
+            response.setStatus(200);
+        else
+            response.setStatus(500);
+
+        writer.write("</body></html>");
+    }
+
+    private String checkGeonetwork() throws Exception {
+        String geonetworkTestURL = this.mappings.getProperty("geonetwork");
```
pmauduit: might be null (in case of ! datadir mode)

pmauduit:
In case of `! datadir` mode, you should be able to do the same by getting a hand on the proxy bean, something like

```
Bean proxy = ApplicationContext.getBean("proxy"); // see proxy-servlet.xml
ArrayList proxy.getTargets();
[...]
```

But I think this is non critical for now.

PR #1372, opened 2016-05-10
Removes dev.mapfish.org maven repository

pmauduit:
the dev.m.o server has been shut down recently, no need to keep the reference in the pom.xml files. In addition, we should rely on build.g.o for the dependencies.

Tests: Only tested that the pom.xml are still valid and launching a mvn clean install does launch the maven process. It will probably be too hard to track every missing dependencies.

Note: need to be cascade-merged until reaching master.

### DIFF
--- a/epsg-extension/pom.xml
+++ b/epsg-extension/pom.xml
@@ -31,15 +31,6 @@
 			<scope>test</scope>
 		</dependency>
 	</dependencies>
-	<repositories>
-		<repository>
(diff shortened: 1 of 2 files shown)


fvanderbiest:
LGTM

Vampouille:
LGTM

pmauduit:
cascade-merged (15.06 -> 15.12 -> master)

PR #1374, opened 2016-05-10
server-deploy - Fixing dependency resolutions in the .m2 repo

pmauduit:
Removing classifiers in the dependencies: since the 15.12 version, the naming convention of the generated wars changed a bit. The classifier is not used anymore, making server-deploy unable to find the artifacts back into the local .m2 repository.

Caveats: This is no longer possible to have multiple wars for different configurations in the same m2 repository, since the last compiled webapp will be used.


pmauduit:
Note:

Before 15.12, the maven-war-plugin was configured in the root pom, I am not sure it was a good idea anyway, and as of the 15.12 version, the configuration has been removed. It might be a proper way to do it anyway ... But I don't want to break things related to docker / deb|rpm packages, so the modifications in the pom should be done with caution.

fvanderbiest:
I'm a bit reluctant to merge this, because of the caveat you mentioned.
If there is really no other way, I'm willing to merge.

pmauduit:
There are probably other ways, which would require to fiddle with all the pom files / maven once again, and I'm more reluctant to this than what is done in this PR.

PR #1376, opened 2016-05-12
Add test data and make them optional with IGNORE_DATA=yes

Vampouille:


Vampouille:
Test data should not be loaded in prod env, so with this commit each files ending with -data.sql will be ignored if IGNORE_DATA=yes is set

Vampouille:
Same apply for ldap service

pmauduit:
Looks good

fvanderbiest:
LGTM, please merge.

PR #1378, opened 2016-05-18
Admin console for master branch

Vampouille:


Vampouille:
We need to protect new console url on security proxy. Maybe here : https://github.com/georchestra/datadir/blob/master/security-proxy/security-mappings.xml

fvanderbiest:
What is the URL for the console ?

Maybe we should integrate it into the header too...

pmauduit:
> We need to protect new console url on security proxy. Maybe here : https://github.com/georchestra/datadir/blob/master/security-proxy/security-mappings.xml

Right (for the datadir mode), also here for the non-datadir mode:
https://github.com/georchestra/georchestra/blob/15.12/security-proxy/src/main/webapp/WEB-INF/applicationContext-security.xml#L44-L71

pmauduit:
some platforms (rennes-metropole) are overriding this file in their configuration profile, this should be addressed there as well

tonio:
URL for the header : `/ldapadmin/console/public/#/home`

fvanderbiest:
> URL for the header : /ldapadmin/console/public/#/home

Can we make it easier to remember, eg `/ldapadmin/console/` ? (`#/home` appended on load, this is ok)
Maybe @Vampouille ?

Vampouille:
I added security mapping for console :
* https://github.com/georchestra/datadir/commit/f9b50aa546b7c54d0619a77c7150cdc6605614ac
* https://github.com/georchestra/datadir/commit/c7b57ddfc5861ae7be9fcec8439897e344e36f22
* https://github.com/georchestra/datadir/commit/4958e8d827f559b1c7d184f46ecb07275c2d100f

So modifications in following branches :
* docker
* master
* docker-master

@fvanderbiest , @pmauduit can you post-review those commits ?

pmauduit:
> @fvanderbiest , @pmauduit can you post-review those commits ?

It does not harm to add some rules in these places, so that looks ok to me.

fvanderbiest:
LGTM ;-)

PR #1379, opened 2016-05-19
Save load filters - update to styler and use OpenLayers.Console

jdenisgiguere:


mapfishapp/src/main/webapp/WEB-INF/jsp/debug-includes.jsp
```diff
@@ -2,6 +2,7 @@
     <script type="text/javascript" src="<%= context %>/lib/externals/ext/ext-all-debug.js"></script>
     <script type="text/javascript" src="<%= context %>/lib/Ext.ux/lib/Ext.ux.js"></script>
     <script type="text/javascript" src="<%= context %>/lib/proj4js/lib/proj4js-combined.js"></script>
+    <script type="text/javascript" src="<%= context %>/lib/externals/openlayers/lib/Firebug/firebug.js"></script>
```
fvanderbiest: There's already one on line 86 which is IMO better to use, we just have to uncomment it.

jdenisgiguere: I agree. I removed it.

PR #1393, opened 2016-06-03
add ID to south panel in order to allow access from addons

hsquividant:
it's the only way I find for using selected features from addons, especially in WPS execution

### DIFF
--- a/mapfishapp/src/main/webapp/app/js/GEOR.js
+++ b/mapfishapp/src/main/webapp/app/js/GEOR.js
@@ -253,6 +253,7 @@ Ext.namespace("GEOR");
         
         var southPanel = new Ext.TabPanel({
             region: "south",
+            id: "southpanel",
             split: true,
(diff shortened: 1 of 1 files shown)


fvanderbiest:
I think @jdenisgiguere found a nice way while developing the atlas addon, see for instance 
https://github.com/georchestra/georchestra/compare/master...jdenisgiguere:atlas_ui

hsquividant:
By this way, my addon inputs would be split in 3 different places (action menu in result panel, action menu in tree layers and in toolbar), not sure this would be clear for user but I might be wrong.
Ext.getCmp is certainly ugly but an easy way to center all functions in addon ... as cadastrapp and even in mapfishapp master ;-)

fvanderbiest:
Since there's no harm in merging this, I'm OK with it.
Folks can do whatever they want with their addons ;-)

PR #1398, opened 2016-06-08
Limit memory usage of webapps with XMX environement variable

Vampouille:
* `sh -c` is used to replace environment variable
* `exec` is used to keep PID 1 on java process

### DIFF
--- a/analytics/src/docker/Dockerfile
+++ b/analytics/src/docker/Dockerfile
@@ -4,5 +4,5 @@ ADD . /
 
 RUN java -jar "$JETTY_HOME/start.jar" --add-to-startd=jmx,jmx-remote,stats
 
-CMD ["java","-Djava.io.tmpdir=/tmp/jetty", "-Dgeorchestra.datadir=/etc/georchestra","-jar","/usr/local/jetty/start.jar"]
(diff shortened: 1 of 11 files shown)


Vampouille:
This PR is linked with https://github.com/georchestra/geonetwork/pull/39

fvanderbiest:
Same idea as https://github.com/georchestra/ansible/pull/41#issuecomment-224810965 : I'm wondering if we could loop on an array of options.

fvanderbiest:
In addition, I think it would be safer to have Xms independant from Xmx

pmauduit:
After reading official Oracle doc: https://docs.oracle.com/javase/8/docs/technotes/tools/windows/java.html

```
For server deployments, -Xms and -Xmx are often set to the same value. See the section "Ergonomics" in Java SE HotSpot Virtual Machine Garbage Collection Tuning Guide at http://docs.oracle.com/javase/8/docs/technotes/guides/vm/gctuning/index.html.
```

fvanderbiest:
OK, this means that, for production deployments, it's better to have XMS = XMX.
But for dev or demo purposes, It might be a pain to consume all available memory at startup.

pmauduit:
Tested locally, seems working fine

PR #1403, opened 2016-06-15
SP - redirect every word to word/

pmauduit:
This was previously done by the frontend webserver configuration. The idea behind this commit is to allow the redirection of "words" to "words/", eg

- http://security-proxy/geonetwork -> /geonetwork/
- ...

Using the "^/[A-Za-z0-9]$" regex pattern should avoid side-effects / regressions.

Tests: runtime tested.

### DIFF
--- a/security-proxy/src/main/webapp/WEB-INF/urlrewrite.xml
+++ b/security-proxy/src/main/webapp/WEB-INF/urlrewrite.xml
@@ -44,8 +44,11 @@
         <to type="forward" last="true">/$1</to>
     </rule>
     <rule>
+        <from>^/([a-zA-Z0-9]+)$</from>
+        <to type="redirect" last="true">/$1/</to>
+    </rule>
+    <rule>
         <from>^/(?!sec/|favicon.ico|receptor|j_spring_security_logout|j_spring_cas_security_check|header.jsp|_static/|cas-logout.jsp|403.jsp|404.jsp|casfailed.jsp)(.*)$</from>
         <to type="forward" last="true">/sec/$1</to>
     </rule>
-
 </urlrewrite>


fvanderbiest:
If I understand correctly, this means that the documentation should be updated with this PR to take into account the fact that we do not need several rewrite rules

https://github.com/georchestra/georchestra/blob/15.12/doc/setup/apache.md

I'm also concerned by the fact that you're attempting to update a stable release here.

pmauduit:
Actually, I'd prefer to migrate the apache configuration pointed by the documentation into the SP's urlrewrite, it will be less error/side-effect-prone (and adapt the documentation consequently).

landryb:
I'm all for this, will ease the transition to nginx config by default :)
Why not doing it only in master for 16.06 ?

fvanderbiest:
> Why not doing it only in master for 16.06 ?

Seems we need it in production ...

pmauduit:
I dislike the "catch-all words" approach in this PR, i'd prefer to match the same patterns as the apache2 default configuration does.

About the master vs 15.12 concern: Actually I can rebase on master, and use the configuration profile to override the urlrewrite file.

About updating the doc: I think we should keep in both sides, to address the 2 following cases:
- If no frontend webserver, the SP will a the trailing /
- If there is a frontend webserver (apache2 or nginx or whatever), it will also need to manage this trailing / issue, before it its the SP.

fvanderbiest:
> I dislike the "catch-all words" approach in this PR, i'd prefer to match the same patterns as the apache2 default configuration does.

I would feel more comfortable indeed.

pmauduit:
rebasing it onto master.

PR #1410, opened 2016-06-23
GS - force TLSv1.1 on Java7

pmauduit:
This won't resolve every SSL-related issues, but at least should force tlsv1.1 on HTTPS connections.

Tested onto a custom Java class which opens an Https connection onto dev.cigalsace.org:

```
$ java -Dhttps.protocols=TLSv1.1,TLSv1.2  HttpsClient https://dev.cigalsace.org/
Response Code : 302

$ java  HttpsClient https://dev.cigalsace.org/
javax.net.ssl.SSLHandshakeException: Remote host closed connection during handshake
```


fvanderbiest:
Hopefully this won't break the service when remote servers feature other SSL protocols than TLSv1.1,TLSv1.2

pmauduit:
We can actually add tlsv1.0 in the list. The main point is to have 1.1 (or 1.2) in the list, because it is supported by java7 but not activated by default. Having them in the list activates them at the JVM level.

Issue #1420, opened 2016-07-04
[mapfishapp] allow to load addons with minified/debug versions


landryb:
As mfapp allows for debug=1 in the URL to load separate js files, would be nice to be able to do the same for addons.

Implem ideas:
- in manifest.json, have different values for 'js' (for production) and 'debugjs' (for debug), listing different js files (minified version for js, full list for debugjs)
- in GEOR_tools.js, fetchAndLoadTools (https://github.com/georchestra/georchestra/blob/master/mapfishapp/src/main/webapp/app/js/GEOR_tools.js#L197) have a different behaviour to populate the 'js' array, depending on the presence of debug in the viewer url ?

fvanderbiest:
Your proposal sounds like a neat idea.

landryb:
Tested working fine at runtime with cadastrapp.js built with georchestra/cadastrapp#235

fvanderbiest:
OK, implemented with https://github.com/georchestra/georchestra/pull/1430

Now we might want to minify the addons we provide by default, when they load more than 1 file...

jusabatier:
Is there a way to load the debug version for addons ?

Maybe with /mapfishapp?debug, but it was not working last time I tried on cadastrapp.js

fvanderbiest:
> /mapfishapp?debug=true

It should work, yes.

jusabatier:
Nope, with /mapfishapp?debug=true I still get the minified version of cadastrapp.js

Only the mapfishapp javascript files are the debug ones.

landryb:
You have to list both versions (minified in the js array, and unminified in the debugjs array) in the manifest file:

```
        "js": [ "js/build/cadastrapp.js" ],
        "debugjs": [
        "js/printBordereauParcellaire.js",
        "js/printRelevePropriete.js",
                "js/selectFeature.js",
                "js/resultParcelle.js",
                "js/resultProprietaire.js",
                "js/searchUtils.js",
                "js/componentUtils.js",
                "js/searchParcelleByOwner.js",
                "js/searchParcelleByRef.js",
                "js/searchCoPropriete.js",
                "js/informationsRequest.js",
                "js/detailUniteFonciere.js",
                "js/habitationDetails.js",
                "js/detailUniteCadastrale.js",
                "js/infoBulle.js",
        "js/menu.js",
        "js/main.js",
        "js/cadastrappUtil.js",
                "js/external/fileUploadField.js"
    ],   
```

[...]

PR #1430, opened 2016-07-12
Allow to provide a debug version of an addon alongside a minified js …

landryb:
…version

If debug is set in the url, the debug version will be loaded.

Implements #1420


fvanderbiest:
This looks good !

PR #1432, opened 2016-07-15
Fix bugs organizations implementation

Vampouille:
This PR fix following bugs : 
* org with no description field generate NPE
* Unable to add a user to an org with '_' in dn
* Unable to create a user with no org
* js lib in privateui folder are not available in account creation form

### DIFF
--- a/ldapadmin/src/main/webapp/account/css/select2.css
+++ b/ldapadmin/src/main/webapp/account/css/select2.css
@@ -1,0 +1,484 @@
+.select2-container {
+  box-sizing: border-box;
+  display: inline-block;
+  margin: 0;
+  position: relative;
+  vertical-align: middle; }
+  .select2-container .select2-selection--single {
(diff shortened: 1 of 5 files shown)


ldapadmin/src/main/webapp/account/css/select2.css
```diff
@@ -0,0 +1,484 @@
...
+  .select2-container .select2-search--inline {
+    float: left; }
+    .select2-container .select2-search--inline .select2-search__field {
+      box-sizing: border-box;
+      border: none;
+      font-size: 100%;
+      margin-top: 5px;
+      padding: 0; }
+      .select2-container .select2-search--inline .select2-search__field::-webkit-search-cancel-button {
+        -webkit-appearance: none; }
```
pmauduit: weird indent (but who does read CSS files anyway :))

pmauduit:
looks good

PR #1437, opened 2016-07-19
Atlas improvements

pmauduit:
- fixing error appearing at runtime on rennes-metropole with velocity (mail template) configuration:

```
Caused by: org.apache.velocity.exception.VelocityException: Error initializing log: Failed to initialize an instance of org.apache.velocity.runtime.log.Log4JLogChute with the current runtime configuration.
	at org.apache.velocity.runtime.RuntimeInstance.initializeLog(RuntimeInstance.java:875)
	at org.apache.velocity.runtime.RuntimeInstance.init(RuntimeInstance.java:262)

```

- Better MFPrintv3 error handling


Vampouille:
It's a good idea to always send emails even if atlas generation failed, so LGTM !

PR #1449, opened 2016-08-10
Create 'admin_emails' table before before loading test data

Vampouille:
`admin_emails` table is created by webapp at runtime but if we want test data in this table. We need to create it on an empty database.


fvanderbiest:
okay

PR #1453, opened 2016-08-23
Rename sequence name

Vampouille:
in order to match sequence name generated by postgres on (big)serial
data type. This will avoid duplicates sequences when we load test datas.


fvanderbiest:
LGTM

PR #1454, opened 2016-08-23
Change DB column type for email body and subject

Vampouille:
Use text column instead of character varying(255)


pmauduit:
Looks good

PR #1456, opened 2016-08-24
Extractor - Allow dash character in WCS store name

Vampouille:
Fix regexp used to find layer in GetCapabilities document. Allow number, dash and underscore characters in store name


pmauduit:
Looks good, but I think we should parse the getcapabilities using a XML parser instead of regexping (we can keep the regexp anyway as a fallback, but a first attempt should be done the xml way). This is a general remark, not targetting this PR though.

fvanderbiest:
Since this is considered as a bugfix, I think it should have been merged into our oldest supported release, or at least 15.12

Let's backport it to 15.12 ... thanks.

Vampouille:
Done with : https://github.com/georchestra/georchestra/pull/1457

PR #1462, opened 2016-08-30
ldap: Change REFINT to true

Vampouille:
Enable integrity check between member and memberOf values. Without this
parameter member and memberOf values are out of sync after some
modifications.


fvanderbiest:
Excellent, please merge !

PR #1468, opened 2016-09-01
Check bytes written by java Channel and continue relaunch transferTo

Vampouille:
transferTo() (and transferFrom()) may not transfer all bytes, so we need
to check written bytes count, compare with source file size and recall
transfer method if necessary.


fvanderbiest:
Nice :-)
LGTM

PR #1474, opened 2016-09-11
Externalized geOrchestra datadir for docker images

fvanderbiest:
This PR adds the ability to run geOrchestra on Docker without rebuilding images when the configuration changes. 

In this setup, a directory on the host machine (`/etc/georchestra`) overrides the default configuration directory, which is already shipped inside the docker image. Since it is a folder on the host, it has to be bootstrapped with an existing configuration directory (If it were a named volume, it would not have to, but I'm not sure it would be easy for newcomers to edit those files)


docs/docker.md
```diff
@@ -0,0 +1,58 @@
...
+
+Run geOrchestra with
+```
+docker-compose up
+```
+
+Open [http://localhost:8080/header/](http://localhost:8080/header/) in your browser
+
+
+## Building your own images
```
Vampouille: I think this part should be moved to `docker/README` page. Because it targets skilled users ?

Vampouille:
Good idea ! This PR allow to customize installation/demo without rebuilding docker images. So docker composition will fit many use cases not only demo ! So LGTM

fvanderbiest:
Thanks for the review. I think we'll also need a reverse proxy sitting in front of the SP, because the docker-master branch in the datadir repository assumes `georchestra.mydomain.org` is the FQDN, not localhost:8080

fvanderbiest:
> I think we'll also need a reverse proxy sitting in front of the SP, because the docker-master branch in the datadir repository assumes georchestra.mydomain.org is the FQDN, not localhost:8080

https://github.com/jwilder/nginx-proxy ?

fvanderbiest:
> because the docker-master branch in the datadir repository assumes georchestra.mydomain.org is the FQDN, not localhost:8080

Nope, I was wrong, eg https://github.com/georchestra/datadir/blob/docker-master/security-proxy/security-proxy.properties#L1-L16 clearly expects localhost:8080

I think (but this will be discussed in another thread) that:
 * the docker-master branch should be updated in order to have `georchestra.mydomain.org` as the default frontend FQDN
 * the composition should include some kind of reverse proxy (eg https://hub.docker.com/r/jwilder/nginx-proxy/) sitting in front of our security-proxy, something like eg:
```yml
rp:
  image: jwilder/nginx-proxy
  ports:
    - "80:80"
    - "443:443"
  volumes:
    - /var/run/docker.sock:/tmp/docker.sock:ro
    - /path/to/certs:/etc/nginx/certs

proxy:
  image: georchestra/security-proxy:latest
  ports:
    - "8080:8080"
  volumes:
    - /etc/georchestra:/etc/georchestra
  links:
    - database
    - ldap
    - cas
    - mapfishapp
    - ...
  environment:
    - VIRTUAL_HOST=georchestra.mydomain.org
    - VIRTUAL_PROTO=https
    - XMX=1g
```

PR #1478, opened 2016-09-13
Enable jetty gzip module to compress static files (js, css, ...)

Vampouille:


fvanderbiest:
Nice, but it should be activated by default for all webapps IMO.

Vampouille:
317a016 add gzip module on all webapps

fvanderbiest:
LGTM.
@pmauduit any objection ?

pmauduit:
> @pmauduit any objection ?

No, it looks good, I think we can still manage to get files without compression by tweaking the request headers if we need anyway (for debugging purposes)

PR #1481, opened 2016-09-19
Introducing GEOR.config.METADATA_VIEW_BASE_URL

fvanderbiest:
This new config option allows SDI administrators to declare a custom metadata view service.

By default, the metadata view service is `/geonetwork/?uuid=` (which is fine for a standard geOrchestra).

In complex setups, for instance when the local catalog harvests a remote node, the SDI admin might want to choose a different metadata view service (eg the original one `/portail/geocatalogue?uuid=`).
The current PR addresses this issue.


fvanderbiest:
The datadir has been updated too (with https://github.com/georchestra/datadir/commit/610a032e06bd282f4677d261e228bec6e71ad088)

PR #1486, opened 2016-09-23
Proxy - Use URLDecoder to handle spaces in url

Vampouille:
URLDecoder converts encoded entities then URI constructor converts space
to %20

### DIFF
--- a/security-proxy/src/main/java/org/georchestra/security/Proxy.java
+++ b/security-proxy/src/main/java/org/georchestra/security/Proxy.java
@@ -33,6 +33,7 @@ import java.net.URI;
 import java.net.URISyntaxException;
 import java.net.URL;
 import java.net.URLEncoder;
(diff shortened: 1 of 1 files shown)


fvanderbiest:
LGTM, thanks !

We also have other proxies in webapps, which are only used for dev purposes.
eg:
 * https://github.com/georchestra/georchestra/blob/15.12/mapfishapp/src/main/java/org/georchestra/mapfishapp/ws/OGCProxy.java
 * https://github.com/georchestra/georchestra/blob/15.12/catalogapp/src/main/java/org/georchestra/catalogapp/Proxy.java
 * https://github.com/georchestra/georchestra/blob/15.12/extractorapp/src/main/java/org/georchestra/extractorapp/ws/doc/OGCProxy.java

Maybe they could be factorized in the `commons` module ?

PR #1487, opened 2016-09-26
Using SLAPD_DOMAIN for setting the LDAP domain

MaxiReglisse:
the LDAP domain should be in accordance with the variable SLAPD_DOMAIN.
we need to transform georchestra.org in dc=georchestra,dc=org.
i used for that code present in https://github.com/dinkel/docker-openldap/blob/master/entrypoint.sh


fvanderbiest:
Looks good, thanks Ernest !

PR #1488, opened 2016-09-27
config is not in /etc/georchestra/datadir

MaxiReglisse:
config files are not in /etc/georchestra/datadir but in  /etc/georchestra

### DIFF
--- a/docs/docker.md
+++ b/docs/docker.md
@@ -21,7 +21,7 @@ Feel free to comment out the apps you do not need.
 Create a configuration directory on your host machine:
 ```
 git clone -b docker-master https://github.com/georchestra/datadir.git
-sudo mv datadir /etc/georchestra
+sudo mv datadir/* /etc/georchestra/.
 ```
 
 Optional: adjust the configuration in `/etc/georchestra` according to your needs.


fvanderbiest:
If  `/etc/georchestra` does not exist yet, `sudo mv datadir /etc/georchestra` creates the directory as expected

MaxiReglisse:
you are right...
but so, the same command could give different results, according that the directory exists or not. my opinion is that it is not good thing. i would prefer : 
mkdir /etc/georchestra
cp datadir/* /etc/georchestra/.

docs/docker.md
```diff
@@ -21,7 +21,7 @@ Feel free to comment out the apps you do not need.
 Create a configuration directory on your host machine:
 ```
 git clone -b docker-master https://github.com/georchestra/datadir.git
-sudo mv datadir /etc/georchestra
+sudo mv datadir/* /etc/georchestra/.
```
Vampouille: Can you add mkdir command ?

MaxiReglisse: it seems i cannot modify the page... perhaps because the issue is closed ?

PR #1489, opened 2016-09-27
config is not in /etc/georchestra/datadir (2)

MaxiReglisse:
sorry, new PR because i could not edit the file any more...

### DIFF
--- a/docs/docker.md
+++ b/docs/docker.md
@@ -21,7 +21,8 @@ Feel free to comment out the apps you do not need.
 Create a configuration directory on your host machine:
 ```
 git clone -b docker-master https://github.com/georchestra/datadir.git
-sudo mv datadir /etc/georchestra
+mkdir /etc/georchestra
+mv datadir/* /etc/georchestra/.
 ```
 
 Optional: adjust the configuration in `/etc/georchestra` according to your needs.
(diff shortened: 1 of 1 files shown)


docs/docker.md
```diff
@@ -21,7 +21,8 @@ Feel free to comment out the apps you do not need.
 Create a configuration directory on your host machine:
 ```
 git clone -b docker-master https://github.com/georchestra/datadir.git
-sudo mv datadir /etc/georchestra
+sudo mkdir /etc/georchestra
+sudo mv datadir/* /etc/georchestra/.
```
fvanderbiest: I dislike this command since it fails to copy the `.git` folder, which is useful for those who will want to version their datadir.

MaxiReglisse: i understand... a solution could consist in downloading files in the right directory, like that:
sudo cd /etc
sudo git clone -b docker-master https://github.com/georchestra/datadir.git georchestra
if the georchestra directory already exists, the download is stopped, with this warning:
fatal: le chemin de destination 'georchestra' existe déjà et n'est pas un répertoire vide.
hint:
you should have a git prompt if you want to know the branch on which you are.
cat /root/.bashrc
# http://stackoverflow.com/questions/15883416/adding-git-branch-on-the-bash-command-prompt
source /usr/local/bin/git-prompt.sh
PS1='[\u@\h \W$(__git_ps1 " (%s)")]\$ '

MaxiReglisse: sorry for the size of the link... there is a # just before http.

fvanderbiest:
Fixed with https://github.com/georchestra/georchestra/commit/4b663b118753482af1e01f9dbd5deaa038ea0fd1

PR #1490, opened 2016-09-28
atlas - fixing bbox vs scale / center parameters

pmauduit:
A commit was kindof forgotten in the development of this webapp. This restores the possibility to provide a spec with a bbox, instead of scale / center to determine how the mfprintv3 should center the map.


Vampouille:
LGTM !

PR #1491, opened 2016-09-28
Atlas mfpv3 datadir

pmauduit:
This allows the bean responsible of instrumenting the datadir to configure itself onto the geOrchestra datadir.


Vampouille:
LGTM !

PR #1493, opened 2016-09-28
Atlas - Set default version for WMS service

Vampouille:
When version is not present in json spec, atlas generation failed because there is no default version for WMS.


fvanderbiest:
LGTM

PR #1502, opened 2016-10-04
Missing redirects

pmauduit:
This adds some redirections in known georchestra endpoints, that was previously needed in the frontend webserver configuration.

This also address a problem with the non-datadir mode on the SP.

Tests: runtime. utests ok (but the testsuite does not cover urlrewrite configuration anyway)

### DIFF
--- a/config/defaults/security-proxy/maven.filter
+++ b/config/defaults/security-proxy/maven.filter
@@ -14,6 +14,7 @@ ldap.admin.password=@shared.ldap.admin.password@
 authoritiesBaseDN=@shared.ldap.groupSearchBaseDN@
 groupRoleAttribute=@shared.ldap.groupRoleAttribute@
 groupSearchFilter=@shared.ldap.groupSearchFilter@
+orgsSearchBaseDN=@shared.ldap.orgsSearchBaseDN@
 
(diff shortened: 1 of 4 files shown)


fvanderbiest:
LGTM, but this PR needs to update the doc + release notes too (we do not need apache rewrite rules anymore):
 * https://github.com/georchestra/georchestra/blob/master/docs/setup/apache.md#configuration
 * https://github.com/georchestra/georchestra/blob/master/RELEASE_NOTES.md#version-1612

config/defaults/security-proxy/maven.filter
```diff
@@ -14,6 +14,7 @@ ldap.admin.password=@shared.ldap.admin.password@
 authoritiesBaseDN=@shared.ldap.groupSearchBaseDN@
 groupRoleAttribute=@shared.ldap.groupRoleAttribute@
 groupSearchFilter=@shared.ldap.groupSearchFilter@
+orgsSearchBaseDN=@shared.ldap.orgsSearchBaseDN@
```
fvanderbiest: is not related to current PR...

pmauduit: I did need it to be able to runtime test the sp actually

PR #1514, opened 2016-10-12
mfapp - sorting contexts alphabetically

pmauduit:
This feature was asked by several customers, in datadir mode, since the contexts are not set in the GEOR_custom.js file anymore, we need to set a convenient way to sort them.

an alphabetical sort (case-insensitive) seems to be a good choice.

Tests: u-test added, not tested at runtime yet (planned to do a live-testing on pigma-dev)

### DIFF
--- a/mapfishapp/src/main/java/org/georchestra/mapfishapp/ws/ContextController.java
+++ b/mapfishapp/src/main/java/org/georchestra/mapfishapp/ws/ContextController.java
@@ -22,7 +22,10 @@ package org.georchestra.mapfishapp.ws;
 import java.io.File;
 import java.io.IOException;
 import java.util.Collection;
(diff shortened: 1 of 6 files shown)


fvanderbiest:
LGTM, please merge

pmauduit:
I'd like to wait for feedback from PIGMA before, we might need to sort by title alphabetically instead of filenames. Not sure of the expected behaviour.

fvanderbiest:
... then ping @pigma ;-)

fvanderbiest:
> we might need to sort by title alphabetically instead of filenames

Since the context titles appears in the viewer window, this seems logic, indeed.

fvanderbiest:
Maybe this should rather be merged into master.

Issue #1515, opened 2016-10-14
ROLE_ANONYMOUS does not exist anymore


fvanderbiest:
It was removed with https://github.com/georchestra/georchestra/pull/223

Yet there are many remnants in the code, eg:
```
fvanderbiest@vaio:~/workspace/georchestra/georchestra (15.12 *$%)$ g grep ROLE_ANONYMOUS
catalogapp/src/main/filtered-resources/WEB-INF/jsp/index.jsp:        // ROLE_ANONYMOUS is added by the security proxy:
catalogapp/src/main/filtered-resources/WEB-INF/jsp/index.jsp:        if (roles[i].equals("ROLE_ANONYMOUS")) {
extractorapp/src/main/filtered-resources/WEB-INF/jsp/index.jsp:        // ROLE_ANONYMOUS is added by the security proxy:
extractorapp/src/main/filtered-resources/WEB-INF/jsp/index.jsp:        if (roles[i].equals("ROLE_ANONYMOUS")) {
geowebcache-webapp/src/main/webapp/WEB-INF/acegi-config.xml:        <property name="userAttribute" value="anonymousUser,ROLE_ANONYMOUS" />
header/src/main/filtered-resources/WEB-INF/jsp/index.jsp:        // ROLE_ANONYMOUS is added by the security proxy:
header/src/main/filtered-resources/WEB-INF/jsp/index.jsp:        if (roles[i].equals("ROLE_ANONYMOUS")) {
ldapadmin/src/main/java/org/georchestra/ldapadmin/ws/HomeController.java:        if (roles != null && !roles.equals("ROLE_ANONYMOUS")) {
ldapadmin/src/main/java/org/georchestra/ldapadmin/ws/HomeController.java:        if(roles != null && !roles.equals("ROLE_ANONYMOUS")) {
ldapadmin/src/test/java/org/georchestra/ldapadmin/ws/HomeControllerTest.java:        request.addHeader("sec-roles", "ROLE_ANONYMOUS");
mapfishapp/src/main/filtered-resources/WEB-INF/jsp/index.jsp:        // ROLE_ANONYMOUS is added by the security proxy:
mapfishapp/src/main/filtered-resources/WEB-INF/jsp/index.jsp:        if (roles[i].equals("ROLE_ANONYMOUS")) {
mapfishapp/src/main/filtered-resources/WEB-INF/jsp/index.jsp:            js_roles_array[0] = "'ROLE_ANONYMOUS'";
[...]

PR #1517, opened 2016-10-14
removing ROLE_ANONYMOUS from 15.12 on

fvanderbiest:
PR for #1515 

There are other traces of `ROLE_ANONYMOUS` but I'm unsure how to handle them:
```
geowebcache-webapp/src/main/webapp/WEB-INF/acegi-config.xml:        <property name="userAttribute" value="anonymousUser,ROLE_ANONYMOUS" />
security-proxy/src/main/webapp/WEB-INF/applicationContext-security.xml:     <s:anonymous granted-authority="ROLE_ANONYMOUS" />
```

### DIFF
--- a/security-proxy/src/main/filtered-resources/WEB-INF/security-proxy.properties
+++ b/security-proxy/src/main/filtered-resources/WEB-INF/security-proxy.properties
@@ -4,7 +4,6 @@
 public.host=${public_host}
 
 # -------  applicationContext-security.xml   -------
-anonymousRole=ROLE_ANONYMOUS
 proxy.contextPath=/sec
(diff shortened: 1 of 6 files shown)


security-proxy/src/main/filtered-resources/WEB-INF/security-proxy.properties
```diff
@@ -4,7 +4,6 @@
 public.host=${public_host}
 
 # -------  applicationContext-security.xml   -------
-anonymousRole=ROLE_ANONYMOUS
```
fvanderbiest: unused in the code !

PR #1523, opened 2016-10-18
add script to migrate organization from 'o' attribut to new ou=orgs

Vampouille:
This script will create organization under ou=orgs from 'o' attribut of
users. Then, it will add user to new organization and delete 'o'
attribut on user entry.


fvanderbiest:
Looks good, thanks !

PR #1538, opened 2016-10-26
Fix URL encoding of query part by removing decoding on this part

Vampouille:
sURl comes from spring with path part not encoded but query part encoded
so we need to encode path part with URI constructor but keep query part
as is.
Tested with space in path and encoded entities in query part :
email=psc%2Btestuser%40georchestra.org


fvanderbiest:
Thanks Julien, this looks good.
This would seriously deserve unit tests in a second time (ping @pmauduit ).

PR #1543, opened 2016-11-07
SP - refactor / adding utests on PR1538

pmauduit:
Following @fvanderbiest's feedback on previous PR, refactor the URI calculation in its own method to make it testable. Adding utests.

Tests: utests added, since there is no functional changes, it should not harm to be merged as is, but would deserve a runtime test on a dev environment.


pmauduit:
Tested on dev.pigma.org btw

fvanderbiest:
LGTM, please merge.

PR #1544, opened 2016-11-08
analytics - cast db count(*) results from bigint to integer to avoid …

Vampouille:
…jpa bug

Is seems that there is a bug in jpa implementation, results of count(*)
queries are multiply by ten after five requests one way to fix this
issue is to explicitly cast count column to 'integer' type. This way, we
always get good results even after five requests.

In postgres, count(*) return a bigint type

test:
@NamedNativeQuery(name="Stats.test",
query = "SELECT CAST(1 AS bigint) AS count"),
this will return 10 after 5 requests

### DIFF
--- a/analytics/src/main/java/org/georchestra/analytics/model/Stats.java
+++ b/analytics/src/main/java/org/georchestra/analytics/model/Stats.java
@@ -35,100 +35,114 @@ import org.hibernate.annotations.Type;
 
 @Entity
 @NamedNativeQueries({
+
+/*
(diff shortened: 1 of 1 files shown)


fvanderbiest:
LGTM, but needs a rebase.
Thanks :-)

You can merge after rebase.

Vampouille:
This PR needs this one : https://github.com/georchestra/georchestra/pull/1545 to fix SQL error :

```
ERROR:  column "ogc_services_log.org" must appear in the GROUP BY clause or be used in an aggregate function
```

PR #1549, opened 2016-11-15
Coordinates addon

fvanderbiest:
New addon, usable from 15.12 on, to allow users to copy/paste a point location using different SRSes.
Greatly inspired from https://github.com/geobretagne/coordinates (but this version still lacks the display of the Z coordinate).


landryb:
Z relies on what ? a WPS/WFS GetFeatureInfo on a given DEM webservice to give you the current altitude ?

fvanderbiest:
See https://github.com/geobretagne/coordinates/blob/master/js/coordinatesquery.js#L58-L82

PR #1552, opened 2016-11-18
ldapadmin - fix empty member list bug on org creation

Vampouille:
and avoid creation of org when form contains error


fvanderbiest:
LGTM, thanks

PR #1568, opened 2016-11-30
console - restore pending orgs in org list

Vampouille:
Webservice return all orgs and then orgs are filtered on user page to
remove pending orgs for already validated users.


Vampouille:
@tonio can take a look at this PR ?

PR #1569, opened 2016-11-30
ldapadmin - only check 'name' field for org edition by admin

Vampouille:
Configured required fields are only used for public user creation form


fvanderbiest:
LGTM !

PR #1570, opened 2016-12-02
console - Add extractions stats on user page

Vampouille:
Extractions stats was available for all users or groups in analytics
page, this PR also adds this statistics for users.


ldapadmin/src/main/webapp/console/app/components/user/user.es6
```diff
@@ -114,7 +114,8 @@ class UserController {
 
     this.config = {
       layers: [ 'layer', 'count' ],
-      requests: [ 'date', 'count' ]
+      requests: [ 'date', 'count' ],
+      extraction: [ 'layer', 'count' ]
```
fvanderbiest: extractions rather than extraction ?

PR #1584, opened 2016-12-13
Fix roles association

tonio:
@Vampouille please review (-:

### DIFF
--- a/ldapadmin/src/main/webapp/console/app/components/user/user.es6
+++ b/ldapadmin/src/main/webapp/console/app/components/user/user.es6
@@ -246,9 +246,6 @@ class UserController {
       'USER': true
     })
     this.$injector.get('$cacheFactory').get('$http').removeAll()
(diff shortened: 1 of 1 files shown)


ldapadmin/src/main/webapp/console/app/components/user/user.es6
```diff
@@ -245,10 +245,8 @@ class UserController {
       'PENDING': false,
       'USER': true
     })
-    this.$injector.get('$cacheFactory').get('$http').removeAll()
-    this.$injector.get('$timeout')(() =>
-      this.groupPromise.$promise.then(this.bindGroups.bind(this))
-    )
+    let $httpDefaultCache = this.$injector.get('$cacheFactory').get('$http')
```
Vampouille: We can keep orignal line : `this.$injector.get('$cacheFactory').get('$http').removeAll()`

PR #1600, opened 2016-12-31
atlas - normalize dockerfile

fvanderbiest:


fvanderbiest:
This will require that we also create & push the georchestra/atlas image to docker hub.

Issue #1604, opened 2017-01-05
mapfishapp - styling vector features which result from a query


fvanderbiest:
Rennes Metropole would like to add the ability to style vector features which result from a query in mapfishapp.

Our proposal is to add this feature as a new button in the resultspanel. When triggered, a window pops open, showing a `gx_(point|line|polygon)symbolizer`  instance + a couple of cancel/apply buttons.

The symbology should be applied to all features from the current resultspanel.

fvanderbiest:
Done for 16.12 with #1605

PR #1605, opened 2017-01-05
mapfishapp - styler for selection

fvanderbiest:
This is a WIP proposal for #1604

Current state:
capture du 2017-01-05 15 19 10


fvanderbiest:
With this proposal, the symbolizer is persisted in browser localstorage, indexed with the feature's featureType (got from feature `fid`).

This means that the symbolizer will always be the same for the layer, whether the query is a WMS getFeatureInfo or a WFS getFeature.

The reset button discards the localStorage record, and restores the previous behavior.

PR #1606, opened 2017-01-05
postgresql docker - fixing patch for the entrypoint

pmauduit:
The build on jenkins is failing since a probable change upstream on the postgresql entrypoint. The patch needs to be adjusted.

Anyway, even if I am pretty confident in this patch, I'd prefer to go through a review since if something changed upstream it might be a reason.

Tests: docker build OK, but might deserve a runtime test.

Note: I'm not very fond of a patch to be applied in the image, it makes it hard to review/understand and reminds me of GeoNetwork first integration in geOr :(

### DIFF
--- a/postgresql/docker-entrypoint.sh.patch
+++ b/postgresql/docker-entrypoint.sh.patch
@@ -1,10 +1,10 @@
---- docker-entrypoint.sh	2016-05-12 18:41:12.719077954 +0200
-+++ docker-entrypoint_modified.sh	2016-05-12 18:33:38.349967649 +0200
-@@ -79,6 +79,7 @@
+--- docker-entrypoint.sh	2017-01-05 19:05:53.734607718 +0100
++++ docker-entrypoint_modified.sh	2017-01-05 19:07:20.810103301 +0100
+@@ -103,6 +103,7 @@
  		for f in /docker-entrypoint-initdb.d/*; do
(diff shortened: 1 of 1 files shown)


pmauduit:
Weird, on my new CI, this patch is not necessary to have the georchestra postgres docker image built

fvanderbiest:
Before building an image depending on image X the CI should docker pull image X
here: `docker pull postgres:9.4`

fvanderbiest:
When I locally pull image `postgres:9.4`, your patch is required
```
Step 11 : RUN cd / && patch -p 0 < docker-entrypoint.sh.patch
 ---> Running in cc9841c19940
patching file docker-entrypoint.sh
Hunk #1 FAILED at 79.
1 out of 1 hunk FAILED -- saving rejects to file docker-entrypoint.sh.rej
The command '/bin/sh -c cd / && patch -p 0 < docker-entrypoint.sh.patch' returned a non-zero code: 1
```

Vampouille:
One way to remove this patch is to replace this patch by a shell `00-ignore-data.sh` script that will remove `*-data.sql` files based on `IGNORE_DATA` value.
This need to be tested because this line : 
https://github.com/docker-library/postgres/blob/master/9.3/docker-entrypoint.sh#L106
will be executed on a deleted file and will failed...

Issue #1475, opened 2016-09-12
docker composition - needs a reverse proxy


fvanderbiest:
As discussed earlier, (in https://github.com/georchestra/georchestra/pull/1474#issuecomment-246480763) I think that:
- the docker-master branch should be updated in order to have `georchestra.mydomain.org` as the default frontend FQDN
- the composition should include some kind of reverse proxy (eg https://hub.docker.com/r/jwilder/nginx-proxy/) sitting in front of our security-proxy, something like eg:

``` yml
rp:
  image: jwilder/nginx-proxy
  ports:
    - "80:80"
    - "443:443"
  volumes:
    - /var/run/docker.sock:/tmp/docker.sock:ro
    - /path/to/certs:/etc/nginx/certs

proxy:
  image: georchestra/security-proxy:latest
  ports:
    - "8080:8080"
  volumes:
    - /etc/georchestra:/etc/georchestra
  links:
    - database
    - ldap
    - cas
    - mapfishapp
    - ...
  environment:
    - VIRTUAL_HOST=georchestra.mydomain.org
    - VIRTUAL_PROTO=https
    - XMX=1g
```

MaxiReglisse:
i agree to say that a reverse proxy in a container is useful when have a single georchestra instance. 
but what happens if you have several georchestra instances on the same docker host, georchestra-prod, georchestra-preprod and georchestra-dev ? i think a single reverse proxy is sufficient, which can be installed on the host.
you can also have other applications in other docker containers on the same host, and then a single reverse proxy is also sufficient. 
consequently, the reverse proxy should be optional.

fvanderbiest:
True but you're an advanced user.  
Advanced users know how to configure their reverse proxies.
Newbies don't ...

MaxiReglisse:
for the moment, i know how to configure an nginx reverse proxy in htttp, but not in https... ERR_SSL_PROTOCOL_ERROR. i will propose you something later.

fvanderbiest:
Well, it should work out of the box with `jwilder/nginx-proxy` (provided it finds the certs he expects ...)

fvanderbiest:
There's also https://github.com/containous/traefik to consider.

MaxiReglisse:
yes yes yes ! :+1:

fvanderbiest:
[...]

PR #1630, opened 2017-01-22
traefik as reverse proxy

fvanderbiest:
PR for https://github.com/georchestra/georchestra/issues/1475

It comes along with https://github.com/georchestra/datadir/pull/50


fvanderbiest:
Note to self: this PR should also update https://github.com/georchestra/georchestra/blob/master/README.md#install

fvanderbiest:
Should also update https://github.com/georchestra/georchestra/blob/master/docs/docker.md

Issue #1631, opened 2017-01-24
docker-compose.override.yml


fvanderbiest:
We should have a docker-compose.override.yml file to specify the path to the dockerfiles for non-georchestra services.
See https://docs.docker.com/compose/extends/

This would allow us to do `docker-compose build`.

fvanderbiest:
> This would allow us to do docker-compose build.

better: `make docker-build` !

PR #1634, opened 2017-01-25
docker compose override and makefile for build

fvanderbiest:
A proposal for #1631

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -1,6 +1,33 @@
-docker-build:
+docker-build: pull-docker-deps dev-docker-build gn3-docker-build geoserver-docker-build georchestra-docker-build
+
+pull-docker-deps:
 	docker pull debian:jessie
 	docker pull dinkel/openldap 
 	docker pull postgres:9.4
 	docker pull tianon/apache2
+
+dev-docker-build:
 	docker-compose build
+
+gn3-docker-build:
+	cd geonetwork; \
+	../mvn -DskipTests clean install; \
+	cd web; \
+	../../mvn -P docker -DskipTests package docker:build
+
+geoserver-docker-build:
+	cd geoserver/geoserver-submodule/src; \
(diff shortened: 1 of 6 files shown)


Vampouille:
Why do you use a separate file ? What is the difference with `build` instruction in main `docker-compose.yml` file ?
@pmauduit do you think we can add webapp build with maven in `Makefile` ?

fvanderbiest:
> Why do you use a separate file ? What is the difference with build instruction in main docker-compose.yml file ?

There's no difference, indeed. This is just to not add clutter into the main file.
Also we could take advantage of this file to override env vars for dev purposes... (later)

> do you think we can add webapp build with maven in Makefile ?

+1 same idea here after I submitted the PR.

fvanderbiest:
Another idea would be to take advantage of this file to remove more clutter from the main compose file.
Namely these services could go into the override compose file:
```
  smtp:
    image: camptocamp/smtp-sink

  courier-imap:
    image: camptocamp/courier-imap
    volumes_from:
      - smtp:rw

  webmail:
    image: camptocamp/sqwebmail
    ports:
      - "8081:80"
    volumes_from:
      - smtp:rw

  geodata:
    image: georchestra/ssh_data
    ports:
      - "2222:22"
    volumes:
      - geoserver_geodata:/mnt/geoserver_geodata
```

Opinions ?

fvanderbiest:
With https://github.com/georchestra/georchestra/pull/1634/commits/0a5d0db3edf0f0ff433a03448850eba951684f12, for those who would like to run all modules:
```
docker-compose -f docker-compose.yml -f docker-compose.other-apps.yml
```

fvanderbiest:
7fbcfe2 adds more targets to Makefile.
Please review again.

pmauduit:
> @pmauduit do you think we can add webapp build with maven in Makefile ?

That is fine for me

PR #1635, opened 2017-01-25
geonetwork - update submodule

Vampouille:
wait for db and ldap service before starting webapp


fvanderbiest:
Oh, you could have done it without a PR ;-)

PR #1639, opened 2017-01-27
[Mapfishapp] Add events when layers are cleared from map

jusabatier:
This PR aim to add 2 events in MFapp : 

- aftercontextrestore in GEOR_wmc.js

- aftercontextcleared in GEOR_managelayers.js

In think thos two events can be very usefull for addons, and tools with their own layers loaded (annotation, print, cadastrapp, search, ...).
With them each layer could catch the event and define a callback for reload his layer(s) when a context is loaded of map is cleared.


fvanderbiest:
This loooks very good, thanks !

PR #1643, opened 2017-01-30
security-proxy - Don't forward headers that start with 'sec-'

Vampouille:


fvanderbiest:
LGTM

security-proxy/src/main/java/org/georchestra/security/HeadersManagementStrategy.java
```diff
@@ -113,7 +114,8 @@ public synchronized void configureRequestHeaders(HttpServletRequest originalRequ
             if (headerName.equalsIgnoreCase(HOST)) {
                 continue;
             }
-            if (headerName.equalsIgnoreCase(SEC_USERNAME) || headerName.equalsIgnoreCase(SEC_ROLES) ) {
+            // Don't forward 'sec-*' headers, those headers must be managed by security-proxy
+            if(headerName.substring(0,4).equalsIgnoreCase(PROTECTED_HEADER_PREFIX)){
```
pmauduit: instead of substring() you can also use startsWith(): https://docs.oracle.com/javase/7/docs/api/java/lang/String.html#startsWith(java.lang.String)

Vampouille: Fix with bb156c8

PR #1654, opened 2017-02-01
mapfishapp - Remove pMode from thesaurus query

Vampouille:
When this parameter is set, geonetwork failed with a 500 error.
Geonetwork try to add an element on XML response with thesaurus
identifier and mode.

Problem is in GetKeywords class :
org.fao.geonet.services.thesaurus.GetKeywords#exec
new Element("thesaurus") throw an exception


Vampouille:
Should fix https://github.com/georchestra/georchestra/issues/1333

Issue #1657, opened 2017-02-07
make war-build-geoserver-geofence does not take into account GeoServerBasePage.html


fvanderbiest:
On master, `make war-build-geoserver-geofence` creates `~/.m2/repository/org/geoserver/web/gs-web-app/2.8.3/gs-web-app-2.8.3.war` which does not have the customized html that we provide in https://github.com/georchestra/georchestra/blob/master/config/defaults/geoserver-webapp/WEB-INF/classes/org/geoserver/web/GeoServerBasePage.html

pmauduit:
I will fix the Makefile, there are some inconsistencies in it:

* You are looking at the gs-web-app-2.8.3.war in your .m2 repository. this webapp is generated by the geoserver submodule and should be considered as the "almost vanilla geoserver version". More concisely, this is the webapp issued from georchestra's geoserver fork, not the actual georchestra geoserver webapp.

* The actual georchestra-geoserver webapp should be built into the `geoserver/webapp` subdirectory. The mvn commands in the Makefile is running `-pl geoserver`, yet it should build transitively the webapp module, it does not for a reason I cannot figure out. So at the end of the build, you are currently only installing the pom for georchestra-geoserver in your .m2.

* If you wish to build the actual georchestra-geoserver webapp, you should select the geoserver/webapp maven module (`-pl geoserver/webapp`).

The last point gives me the following after having launched the Makefile target:

```
[INFO] --- maven-install-plugin:2.4:install (default-install) @ geoserver-webapp ---
[INFO] Installing /home/pmauduit/projects/georchestra/georchestra/geoserver/webapp/target/geoserver-template.war to /home/pmauduit/.m2/repository/org/georchestra/geoserver-webapp/16.12-SNAPSHOT/geoserver-webapp-16.12-SNAPSHOT.war
[INFO] Installing /home/pmauduit/projects/georchestra/georchestra/geoserver/webapp/pom.xml to /home/pmauduit/.m2/repository/org/georchestra/geoserver-webapp/16.12-SNAPSHOT/geoserver-webapp-16.12-SNAPSHOT.pom
[INFO] ------------------------------------------------------------------------
[INFO] BUILD SUCCESS
[INFO] ------------------------------------------------------------------------
[INFO] Total time: 6.394 s
[INFO] Finished at: 2017-02-08T14:57:18+01:00
[INFO] Final Memory: 69M/711M
[INFO] ------------------------------------------------------------------------
```

[...]

PR #1659, opened 2017-02-08
Makefile - fixing geoserver build

pmauduit:
See #1657


pmauduit:
Tests:

- make geoserver without geofence
- make geoserver with geofence

the generated webapp ends here: `${homedir}/.m2/repository/org/georchestra/geoserver-webapp/16.12-SNAPSHOT/geoserver-webapp-16.12-SNAPSHOT.war`

PR #1661, opened 2017-02-08
Make account creation works in IE11

### LINKED ISSUE #1596: ldapadmin - error 400 on create new account with IE 11 or out of date FireFox
We have a serious problem : many of users that are customed to use IE 11.
After hav filled up the form on /ldapadmin/account/new and validate they obtain an error 400 page

```
HTTP Status 400 - Required String parameter 'orgCities' is not present
type Status report
message Required String parameter 'orgCities' is not present
description The request sent by the client was syntactically incorrect.
```

Users with an 'old' FireFox met this bug too. Many users have a FireFox 36 installed by default. The problem is that FireFox < to v 38 or 39 don't update automatically.

We should detect them and show a specific message.

tonio:
So:

 - `Array.from` [polyfill](https://developer.mozilla.org/en/docs/Web/JavaScript/Reference/Global_Objects/Array/from#Polyfill) doesn’t work for *spread syntax*,
 - IE11 can’t load [georchestra WFS](https://sdi.georchestra.org/geoserver/ows?SERVICE=WFS&REQUEST=GetFeature&typeName=gadm:gadm_for_countries&outputFormat=json&BBOX=-9.6200,41.1800,10.3000,51.5400) due to dark certificates issues, even just to display it in a new tab, neither in the app.
 - IE11 CORS policy [differs](https://developer.mozilla.org/en-US/docs/Web/Security/Same-origin_policy#IE_Exceptions) from other browsers (port are not taken into account)…

I finally successully test with the following changes & [geobretagne WFS](https://geobretagne.fr/geoserver/mig/wfs?service=wfs&version=2.0.0&request=getfeature&typename=regions_nouvelles&outputformat=application/json&srsName=EPSG:4326). 

Please note that CORS headers are needed in the response.

Hopefully fix #1596

### DIFF
--- a/ldapadmin/src/main/webapp/console/app/components/area/area.es6
+++ b/ldapadmin/src/main/webapp/console/app/components/area/area.es6
@@ -2,9 +2,9 @@ require('components/area/area.tpl')
 
 class AreaController {
 
-  static $inject = [ '$injector', '$q' ]
+  static $inject = [ '$injector', '$q', '$http' ]
 
-  constructor ($injector, $q) {
+  constructor ($injector, $q, $http) {
     this.$injector = $injector
     const CONFIG_URI = this.$injector.get('LDAP_PUBLIC_URI') + 'orgs/areaConfig.json'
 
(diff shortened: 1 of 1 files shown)


fvanderbiest:
> Please note that CORS headers are needed in the response.

You mean something like this ?
```
<Location /geoserver>
  Header set Access-Control-Allow-Origin "*"
  Header set Access-Control-Allow-Headers "Origin, X-Requested-With, Content-Type, Accept"
</Location>
```

PR #1673, opened 2017-02-27
security-proxy - Use human readable org name for ogc stats

Vampouille:
instead of internal identifier


fvanderbiest:
Does this mean that past records from the ogc logs database need to be updated to be kept in sync with the new convention ?

PR #1675, opened 2017-03-01
analytics - use response writer to fix encoding

Vampouille:
Previous commit : https://github.com/georchestra/georchestra/pull/1625/commits/9f0fa5cff719e29091178f420a3625b195475cbb does not fix issue on tomcat. This PR works with tomcat and jetty.


pmauduit:
Looks good, but I cannot figure out why the behaviour is different between tomcat & jetty, nor why the annotation for the returned content has no effect.

PR #1678, opened 2017-03-15
header - Add empty img/ and fix missing logo.png

Vampouille:


fvanderbiest:
"ceinture et bretelle" !
LGTM

header/src/main/webapp/.gitignore
```diff
@@ -1,3 +1,4 @@
 /css
-/img
+/img/*
+!/img/README.md
```
pmauduit: What does the "!" mean ?

Vampouille: except

> What does the "!" mean ?

except

PR #1679, opened 2017-03-16
mapfishapp - ignore invalid WMC in db

pmauduit:
Reported by Rennes-metropole, the webservice which lists the WMCs is broken if one of the document is malformed in database.

See (restricted access):
https://github.com/camptocamp/georchestra-rennes-configuration/issues/540

Tests: runtime ok
testuite ok, but this code is untested by the testsuite (this should deserve a one).


pmauduit:
I added a unit test for the method (see 2nd commit), so modified code is unit-tested now, code coverage for A_DocService: 62%, WMCDocService: 98%

Issue #378, opened 2013-10-10
google closure compiler rather than jsbuild ?


fvanderbiest:
... or just as a second pass optimizer after jsbuild...

pmauduit:
:+1: with this move, so that we can have a CI with no extra needed packages. I'd like to have a look on this.

pmauduit:
Impacted webapps:

- [] catalogapp
- [] mapfishapp
- [] extractorapp

fvanderbiest:
I'm not sure we could completely remove the need for jsbuild, since all source files (including openlayers & geoext sources) have the required headers for jsbuild.

fvanderbiest:
We'd better change our viewer to use more modern libraries.

pmauduit:
Could be of interest:

https://github.com/geonetwork/core-geonetwork/commit/da250f9b5ea7e77d6afd939db5f50e4311410d30

PR #1700, opened 2017-03-25
extractorapp,mfapp,catalogapp - make build idempotent

pmauduit:
When the repository is moved to another directory, jsbuild is unable to
build the JS anymore, making maven to fail. This is due to python
virtualenv which has to know the actual absolute directory to store its
dependencies.

This case is not supposed to happen, but it might in some specific
environments (Jenkins CI). Removing the generated virtualenv seems
acceptable, since it does not take ages to rebake a one from scratch.

Tests: compilation level on each webapp:
* mvn package after a move fails (logically, it requires a clean)
* mvn clean package OK
* mvn clean: removes target/ and also jsbuild/env

Note: a similar fix could also have been applied onto jsbuild.sh, but I
prefer to have the build logic maven-side, and taking some time to work
on the build minification would be better spent trying some more
integrated tools than a shell script plus a virtualenv plus quite old
dependencies, see https://github.com/georchestra/georchestra/issues/378

### DIFF
--- a/catalogapp/pom.xml
+++ b/catalogapp/pom.xml
@@ -185,6 +185,20 @@
           <target>${java-version}</target>
         </configuration>
       </plugin>
+      <plugin>
+        <artifactId>maven-clean-plugin</artifactId>
+        <version>3.0.0</version>
+        <configuration>
+          <filesets>
+            <fileset>
+              <directory>jsbuild</directory>
+              <includes>
+                <include>env/**</include>
+              </includes>
+            </fileset>
+          </filesets>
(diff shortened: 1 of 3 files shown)


fvanderbiest:
Why does Jenkins CI move the directories ?

pmauduit:
No idea but this is the behaviour on the new official c2c CI. I guess on some circumstances this can happen, so having the build being able to manage this situation sounds like a good move

fvanderbiest:
OK, LGTM

PR #1704, opened 2017-04-03
security-proxy - Disallow proxy request with Class C hostname

Vampouille:
Internal proxy (/proxy?url=...) request with be filtered based on
address class of hostname.


Vampouille:
See also : https://github.com/georchestra/geonetwork/pull/46 and https://github.com/georchestra/geonetwork/pull/45

PR #1706, opened 2017-04-12
[Mapfishapp] Correct annotation bug

jusabatier:
Actually, when I load an exported JSON from Annotation addon, I have a message : it is not valid.

This little correction solve the problem.


fvanderbiest:
Thanks, but to be fixed on 15.12 instead ...

fvanderbiest:
OK, code does not exist before 16.12...

PR #1731, opened 2017-05-29
mapfishapp - fix getfeatureinfo on layers from /geoserver/ns/ows

fvanderbiest:


fvanderbiest:
"upported" to 16.12 and master

PR #1733, opened 2017-05-29
Forgot password, try to avoid stack traces.

cmangeat:
Stack traces are erratic.

Don't know how to test it.

Suppose it has to do with attempt to reuse yesterday connections (it happens in the morning ?), which are not discarded from the connection pool.

Apache docs states that when "testOnBorrow" true, validation query is "The SQL query that will be used to validate connections from this pool before returning them to the caller. If specified, this query MUST be an SQL SELECT statement that returns at least one row. If not specified, connections will be validation by calling the isValid() method".

Theory is that isValid method is not sufficient to leverage the connection up.

### DIFF
--- a/analytics/src/main/webapp/WEB-INF/ws-servlet.xml
+++ b/analytics/src/main/webapp/WEB-INF/ws-servlet.xml
@@ -24,6 +24,7 @@
         <property name="url" value="${dlJdbcUrlOGC}"/>
         <property name="driverClassName" value="org.postgresql.Driver"/>
         <property name="testOnBorrow" value="true"/>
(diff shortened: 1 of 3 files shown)


fvanderbiest:
Looks good, but analytics also uses waitForDB IIRC

cmangeat:
Can be reproduced by 
"SELECT pg_terminate_backend(pid) FROM pg_stat_activity WHERE datname='georchestra';"
On the db before launching the procedure.

fvanderbiest:
> UserTokenDao exists methods implementation may beg for improvement, as with try/catch/finally mechanism and exception thrown on c.close() in finally, exception thrown on cmd.execute() in try never goes up.

Go ahead, please :-)

PR #1737, opened 2017-06-01
Do not throw exception in finally 2

cmangeat:
cf. #1733


ldapadmin/src/test/java/org/georchestra/ldapadmin/ds/UserTokenDaoTest.java
```diff
@@ -17,6 +17,7 @@
 import org.junit.Before;
 import org.junit.Test;
 import org.mockito.Mockito;
```
fvanderbiest: `import org.mockito.Mockito;` ?

fvanderbiest:
Leaving the full review to @pmauduit 
Thanks.

PR #1738, opened 2017-06-06
ldapadmin - Update email validator to allow new domains like .alsace

Vampouille:


Vampouille:
I just restore import of LdapTemplate class because Travis failed to build ldapadmin webapp.

```java
import org.springframework.ldap.core.LdapTemplate;
```

PR #1742, opened 2017-06-12
add ability to export features as WKT

fvanderbiest:
The SRS of the exported geometry is managed by the mouse position control.


fvanderbiest:
Implements #1742

PR #1747, opened 2017-06-20
create dirs for volumes and grant ownership

fvanderbiest:
Having `jetty` owning the `/var/local/tiles` folder is required to let gwc create new folders.

Same is true for the georchestra datadir folder, which is thus writeable by user whose uid is 999

### DIFF
--- a/analytics/src/docker/Dockerfile
+++ b/analytics/src/docker/Dockerfile
@@ -6,7 +6,7 @@ RUN java -jar "$JETTY_HOME/start.jar" --add-to-startd=jmx,jmx-remote,stats
 
 ADD . /
 
-RUN mkdir -p /etc/georchestra && \
+RUN mkdir /etc/georchestra && \
     chown jetty:jetty /etc/georchestra
 
 VOLUME [ "/tmp", "/run/jetty" ]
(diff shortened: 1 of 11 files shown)


analytics/src/docker/Dockerfile
```diff
@@ -6,6 +6,9 @@ RUN java -jar "$JETTY_HOME/start.jar" --add-to-startd=jmx,jmx-remote,stats
 
 ADD . /
 
+RUN mkdir -p /etc/georchestra && \
+    chown jetty:jetty /etc/georchestra
+
 VOLUME [ "/tmp", "/run/jetty" ]
```
Vampouille: Why not adding `/etc/georchestra` as volume ?

fvanderbiest: Because I want to leave the choice to people running the image to have the config in `/etc/georchestra` as an external volume or to create self-contained images (app+config)

fvanderbiest:
```
RUN mkdir /etc/georchestra && \
￼         chown jetty:jetty /etc/georchestra
```
is required because docker overwrites the volume's root directory ownership with the one coming from the image.

fvanderbiest:
Surprisingly ...
```
Step 7 : RUN mkdir /etc/georchestra /var/local/extracts &&     chown jetty:jetty /etc/georchestra /var/local/extracts
 ---> Running in a84f0132f3c8
mkdir: cannot create directory ‘/etc/georchestra’: File exists
```

fvanderbiest:
a priori fixed with eaf0899

Issue #1750, opened 2017-06-27
SP should not convert downstream HTTP errors into 200 OK


fvanderbiest:
As stated in #1406 this is not the role of the SP, rather a front LB's job.

fvanderbiest:
See also https://github.com/georchestra/georchestra/pull/963/files which introduced this "feature".

fvanderbiest:
Also fixes #701 at the same time

fvanderbiest:
This [comment](https://github.com/georchestra/georchestra/issues/506#issuecomment-30402497) makes me think there's work to be done in ldapadmin too...

PR #1752, opened 2017-06-27
proxy - do not catch downstream HTTP error codes

fvanderbiest:
... reverts most of #963 since this has to be done by the front LB rather than the SP.

Implements most of #1750


fvanderbiest:
Right, let's wait... 
Indeed this PR will not help fix our issue with haproxy.

Issue #1753, opened 2017-07-04
Mapfishapp - Enable text selection in grids


jusabatier:
I think it can be interesting to enable text selection in grids, so users can copy fields returned by getfeatureinfo in the clipboard.

Example available here : https://opendata.agglo-lepuyenvelay.fr/mapfishapp/

It can be done by redefining the grids default template like explained in this topic : 
https://ahlearns.wordpress.com/2011/10/27/ext-js-make-grid-text-selectable/

In the main.css adding : 
```
.x-selectable, .x-selectable * {
        -moz-user-select: text!important;
        -khtml-user-select: text!important;
}
```

And in a js file : 
```
if (!Ext.grid.GridView.prototype.templates) {
   Ext.grid.GridView.prototype.templates = {};
}
Ext.grid.GridView.prototype.templates.cell = new Ext.Template(
   '<td class="x-grid3-col x-grid3-cell x-grid3-td-{id} x-selectable {css}" style="{style}" tabIndex="0" {cellAttr}>',
   '<div class="x-grid3-cell-inner x-grid3-col-{id}" {attr}>{value}</div>',
   '</td>'
);
```
I don't know in which js file it's more apropriate to put this code.

We also start an issue about this function in cadastrapp issue https://github.com/georchestra/cadastrapp/issues/341, but it can be a better option to put this function directly in mapfishapp.

We have to test if it not break anything in addons, functionalities, ...

When this checks are done and if everythings are ok, do you think we can backport this function directly in 16.12 or it will be in master/17.xx ?

fvanderbiest:
> I don't know in which js file it's more apropriate to put this code.

Probably in GEOR.js as for the other monkey patches...

fvanderbiest:
> When this checks are done and if everythings are ok, do you think we can backport this function directly in 16.12 or it will be in master/17.xx ?

Let's merge this in master, and after some tests, backport it to 16.12 + 15.12

fvanderbiest:
backported to 16.12

landryb:
Any chance to backport it to 15.12, or it's too risky ?

fvanderbiest:
No, it should not be. You can submit a PR targeting 15.12.
Thanks.

PR #1754, opened 2017-07-06
[Mapfishapp] Enable text selection in gridviews

jusabatier:
As discussed in those threads : 
https://github.com/georchestra/georchestra/issues/1753
https://github.com/georchestra/cadastrapp/issues/341

This PR enable to select text in gridviews (GFI results, addons results tabs, ...) and copy/paste it.

Seems like it doesn't break anything, but just tested since 3~4 days, so need to confirm with more time.

After that it could be a good thing to backport it on 16.12 & 15.12

### DIFF
--- a/mapfishapp/src/main/webapp/app/css/main.css
+++ b/mapfishapp/src/main/webapp/app/css/main.css
@@ -519,3 +519,8 @@ div.olMapViewport {
 	background-color: #fff;
 }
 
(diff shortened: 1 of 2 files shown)


fvanderbiest:
Thanks Julien, merging into master for tests.

fvanderbiest:
backported to 16.12 with 81f6581

PR #1763, opened 2017-08-01
OGC statistics - Reopens the JDBC connection if closed

pmauduit:
It seems that the network connection to the database used to persist OGC statistics entry can be closed, and nothing in the code was done to detect it. This commit proposes to reopen the connection in case it is detected as closed.

Note: a clearly proper way to address this issue would be to introduce c3p0 to this module.

Tests: clean package ok. Would deserve runtime tests.

Motivations: see
https://github.com/camptocamp/georchestra-rennes-configuration/issues/611


fvanderbiest:
LGTM !

PR #1769, opened 2017-08-16
Improve ogcstats performance

Vampouille:
On large stats database, analytics queries can take a long time before before returning results. This is caused by a combination of partitioned tables and sql parameterized queries. In this case, SQL execution plan is computed without parameters values. This result in a complete scan of all tables.
In order to fix this issue, this PR send non-parameterized SQL query without losing benefit of parameterized SQL query: 
* add name on parameter
* try to reconnect one time if tests query failed

Example:
```java 
String sql = "SELECT * FROM users WHERE group_name = {group} LIMIT {count}";
Map<String, Object> sqlValues = new HashMap<String, Object>();
sqlValues.put("group", "admin");
sqlValues.put("count", 100);
 
DBConnection db = new DBConnection(jpaDataSource);
String rawSql = db.generateQuery(sql, sqlValues);
ResultSet res = db.execute(rawSql);
 
```


fvanderbiest:
Wow, this is huge !
Impressive works, thanks @Vampouille :-)

MaelREBOUX:
Did you had indexes on the tables ? It misses currently.

Vampouille:
> Did you had indexes on the tables ? It misses currently.

I don't think we need indexes because most operations on stats tables are updates. Those tables are only read by admin. So adding indexes will not improve performances.

Issue #1773, opened 2017-09-01
extractorapp,mfapp,catalogapp - minification is broken when using python2 venv


pmauduit:
See comments here for the details and a patch proposal:
https://github.com/georchestra/docker/issues/5

fvanderbiest:
Should be fixed now.

PR #1775, opened 2017-09-05
jstools switched to python3

fvanderbiest:
Patch proposal for https://github.com/georchestra/docker/issues/5#issuecomment-326613974 / https://github.com/georchestra/georchestra/issues/1773


fvanderbiest:
This should fix minification issues from 15.12 on.

PR #1777, opened 2017-09-07
Remove spring mvc dependency from common

Vampouille:
And move route definition in mapfishapp and extractorapp
Including spring mvc in common module conflict with other version of
spring mvc included in webapp.

### DIFF
--- a/atlas/pom.xml
+++ b/atlas/pom.xml
@@ -208,6 +208,11 @@
         </dependency>
         <dependency>
           <groupId>org.springframework</groupId>
+          <artifactId>spring-context-support</artifactId>
+          <version>${spring.version}</version>
+        </dependency>
+        <dependency>
+          <groupId>org.springframework</groupId>
           <artifactId>spring-expression</artifactId>
           <version>${spring.version}</version>
         </dependency>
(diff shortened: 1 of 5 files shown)


pmauduit:
This seems to break the atlas webapp

pmauduit:
Something is broken at runtime on PPIGE-dev after redeploying:

https://dev.ppige-npdc.fr/mapfishapp/ws/app/js/GEOR_custom.js -> 404

I wonder if urlrewrite is misconfigured afterwards

pmauduit:
Ok, got it, since the current annotation:
https://github.com/georchestra/georchestra/blob/16.12/mapfishapp/src/main/java/org/georchestra/mapfishapp/ws/HomeController.java#L42

The actual path where you can get the JS file is the following:
`http://georchestra/mapfishapp/ws/home/app/js/GEOR_custom.js`

the mapping is still working on extractorapp because the global RequestMapping on the class is "/" instead of "/home":
https://github.com/georchestra/georchestra/blob/16.12/extractorapp/src/main/java/org/georchestra/extractorapp/ws/doc/HomeController.java#L50

PR #1793, opened 2017-10-06
[WIP] Fix failed docker build because upstream image change

Vampouille:
New jetty docker image now set user as jetty also for build step, so we
cannot add package. This commit temporary switch to root user for
installaing packages.
See :
https://github.com/appropriate/docker-jetty/commit/19b433117c3c22550ce87be81cbb6f6952890928#diff-dad6c4a2a53860754faaf15e201d8b0d


pmauduit:
Already applied

PR #1798, opened 2017-10-17
extractorapp - Fix SQL query to check acceptance

Vampouille:
Change operator used to compare with null. Use `IS NULL` instead of `=
NULL`. Also reformat SQL query.


fvanderbiest:
LGTM

PR #1802, opened 2017-10-25
extractor - fixing anonymous extracts

pmauduit:
Bare-metal sql queries won't pass if the spec contains empty (null) fields, especially the roles.split(",") call, leaving the extraction in an unknown (but running) state which never end up.

Tests:
* Integration test added, but requires a PostGIS db with the extractor schema (since one extractor db log requires a geometry column).
* Already tested on georhena-prod by redeploying modified class files

### DIFF
--- a/extractorapp/src/main/java/org/georchestra/extractorapp/ws/extractor/task/ExtractionTask.java
+++ b/extractorapp/src/main/java/org/georchestra/extractorapp/ws/extractor/task/ExtractionTask.java
@@ -428,9 +428,9 @@ public class ExtractionTask implements Runnable, Comparable<ExtractionTask> {
 					"VALUES (?, ?, ?, ?)",
 					Statement.RETURN_GENERATED_KEYS);
 
-			pst.setString(1, this.requestConfig.username);
-			pst.setArray(2, c.createArrayOf("varchar", this.requestConfig.roles.split("\\s*;\\s*")));
-			pst.setString(3, this.requestConfig.org);
+			pst.setString(1, this.requestConfig.username == null ? "" : this.requestConfig.username);
+			pst.setArray(2, c.createArrayOf("varchar",this.requestConfig.roles == null ? new String[0] : this.requestConfig.roles.split("\\s*;\\s*")));
+			pst.setString(3, this.requestConfig.org == null ? "" : this.requestConfig.org);
 			pst.setString(4, this.requestConfig.requestUuid.toString());
 
 			this.logId = this.executeAndGetGeneratedKey(pst);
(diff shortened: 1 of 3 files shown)


extractorapp/src/test/java/org/georchestra/extractorapp/ws/extractor/task/ExtractionTaskTest.java
```diff
@@ -0,0 +1,76 @@
...
+
+    private boolean pgAvailable;
+    private int beforeCount;
+    @Before
+    public void setUp() throws Exception {
+        try {
+            Connection c = dataSource.getConnection();
+            this.beforeCount = getLayerLogCount(c);
+        } catch(Exception e) {
+            pgAvailable = false;
```
pmauduit: ... forgot a return here. fix on the go.

extractorapp/src/main/java/org/georchestra/extractorapp/ws/extractor/task/ExtractionTask.java
```diff
@@ -468,14 +468,8 @@ private void statSetRunning() {
...
-		} catch (SQLException e) {
-			LOG.error(e.getMessage());
-		} catch (TransformException e) {
-			LOG.error(e.getMessage());
-		} catch (NoSuchAuthorityCodeException e) {
-			LOG.error(e.getMessage());
-		} catch (FactoryException e) {
-			LOG.error(e.getMessage());
+		} catch (Exception e) {
+			LOG.error("Unable to log the extraction parameters in database", e);
```
Vampouille: +1

extractorapp/src/main/java/org/georchestra/extractorapp/ws/extractor/task/ExtractionTask.java
```diff
@@ -428,9 +428,9 @@ private void statSetRunning() {
 					"VALUES (?, ?, ?, ?)",
 					Statement.RETURN_GENERATED_KEYS);
 
-			pst.setString(1, this.requestConfig.username);
-			pst.setArray(2, c.createArrayOf("varchar", this.requestConfig.roles.split("\\s*;\\s*")));
-			pst.setString(3, this.requestConfig.org);
+			pst.setString(1, this.requestConfig.username == null ? "" : this.requestConfig.username);
```
fvanderbiest: I would have prefered to set these fields to NULL, eventually making an evolution to the schema.
https://github.com/georchestra/georchestra/blob/16.12/postgresql/09-extractor-app.sql

PR #1805, opened 2017-10-27
security-proxy - Remove custom HTTP Host header

Vampouille:
In order to fix an outdated geoserver bug, Host header is modified to
match original request. So Host header does not match host used in URL.
This can lead to error when using virtual host based on host with
apache.


Vampouille:
Merged in master with ad91299bea38bc85c10a5c74f68111baed2322ae

fvanderbiest:
There was no related test ?

Vampouille:
> There was no related test ?

not yet, this implies geoserver and security-proxy. I think we need integration tests for this.

Vampouille:
> In order to fix an outdated geoserver bug

not so outdated, geoserver still use `Host` header to generate redirection. So when `Host` does not match public domain, redirection failed. We need to revert this PR.

PR #1818, opened 2017-11-14
Revert "security-proxy - Remove custom HTTP Host header"

Vampouille:
This reverts commit 993d501883943311491f7ae524b21619bb0922ce.

Because transparent proxy should copy `Host` header as is. RFC : https://www.w3.org/Protocols/rfc2616/rfc2616-sec14.html#sec14.23 say that `Host` header should match hostname of original request.


Vampouille:
This revert PR #1805

fvanderbiest:
okay

Vampouille:
merge into master with b4900f7

Issue #1745, opened 2017-06-20
Upgrade jetty images to 9.4 ?


fvanderbiest:
We're currently using `jetty:9.3-jre8` but `jetty:9.4-jre8` is available
https://hub.docker.com/_/jetty/

fvanderbiest:
also in the poms ... https://github.com/georchestra/georchestra/blob/c9762a25aa948bb3e3258655493c9adb561a667a/mapfishapp/pom.xml#L277

fvanderbiest:
Proposal : use tag `9-jre8` instead ?

PR #1840, opened 2017-12-06
Jetty9

fvanderbiest:
PR for #1745


pmauduit:
No objection so far, jetty 9 should come with a recent javax.servlet implementation, but this should not harm the webapps, would deserve a runtime test though.

PR #1864, opened 2018-01-02
Remove useless 'LdapContextSource' bean

Vampouille:


Vampouille:
It's used to populate : 
https://github.com/georchestra/georchestra/blob/16.12/security-proxy/src/main/java/org/georchestra/security/LdapUserDetailsRequestHeaderProvider.java#L66-L67
So we need to keep `LdapContextSource` and `ldapTemplate` beans.

PR #1875, opened 2018-01-17
GN 3.4.1 / GS 2.12 integration

pmauduit:
# GS integration

This commit partially reverts 4400757aa87ed3baf9423c84af977e806b2bf4c4 which introduced a regression. There are 2 types of POST request:

* the request body is in a URLencoded form,
* the request body is a payload

The second type is mainly used in some applications (LDAPAdmin), where the body is a JSON payload. We need to send it, the previous commit was stripping the body, hence the POST data was null.

# GN integration

In order to protect the webapp for Cross-site-request-forgery (CSRF) attacks, a token is computed which can be fetch from a POST call onto /geonetwork/srv/{lang}/xml.info. The ajax call leads to a 403 because of the missing CSRF token, but the header Set-Cookie provides the token, but the SP was not transmitting this header to the browser. This commit ensures that a webapp providing such a header is correctly provided to the actual client.

Tests: Tested on the GGE instance.

### DIFF
--- a/security-proxy/src/main/java/org/georchestra/security/Proxy.java
+++ b/security-proxy/src/main/java/org/georchestra/security/Proxy.java
@@ -708,8 +708,18 @@ public class Proxy {
                 // 403 and 404 are handled by specific JSP files provided by the
                 // security-proxy webapp
                 if ((statusCode == 404) || (statusCode == 403)) {
-                    finalResponse.sendError(statusCode);
-                    return;
+                    // Hack for GN3.4: to protect against CSRF attacks, a token
+                    // is provided by the xml.info service. Even if the return
+                    // code is a 403, we are interested in getting the
+                    // Set-Cookie value.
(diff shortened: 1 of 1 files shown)


pmauduit:
We might want to wait before merging, because the code is currently missing a URL normalization:
https://github.com/georchestra/georchestra/blob/16.12/security-proxy/src/main/java/org/georchestra/security/Proxy.java?utf8=%E2%9C%93#L829

if `targets.mapping.properties` references a `geoserver=http://geoserver-lb/geoserver/`, the adjust location will fail to actually adjust the location returned by GS, because the expected pseudo-code pointed above will do the following:

```
  if resolvedURI "http://geoserver-lb:80/geoserver/..." startsWith "http://geoserver-lb/geoserver/"
```
Which will fail because of the port specification in one of the string.

pmauduit:
> We might want to wait before merging

Actually, there is a workaround by specifying the port in the targets-mapping.properties

PR #1877, opened 2018-01-18
CI - debian packages - fix datadir checkout

pmauduit:
Warning: on the following cascade-merge in master, one will have to merge replacing `${project.parent.version}` by `master`.

Tests: tested on catalogapp (which disappeared from master, making the build fail).

### DIFF
--- a/analytics/pom.xml
+++ b/analytics/pom.xml
@@ -362,6 +362,8 @@
               <checkoutDirectory>${project.build.directory}/deb/etc/georchestra</checkoutDirectory>
               <connectionUrl>scm:git:https://github.com/georchestra/datadir.git</connectionUrl>
               <pushChanges>false</pushChanges>
+              <scmVersion>${project.parent.version}</scmVersion>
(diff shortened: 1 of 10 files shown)


fvanderbiest:
LGTM !

fvanderbiest:
I merged it, but thinking about it twice, I'm wondering if we should not introduce a property such as
```
<debianDatadirScmVersion>master</debianDatadirScmVersion>
```
as we did for the docker build

This would maybe also make it clearer that, when we release, this should be updated

fvanderbiest:
After discussion with @pmauduit we agreed to do so.

landryb:
The buildbot liked this merge: https://packages.georchestra.org/bot/builders/buildjob/builds/180 - and new wars are up on https://packages.georchestra.org/bot/wars/16.12/

pmauduit:
I propose `packageDatadirScmVersion` instead, because I decided to also keep the rpmPackage profiles (even if the CI does not produce these artifacts anymore).

Issue #1878, opened 2018-01-18
Use a dedicated packageDatadirScmVersion var to decide which datadir branch to ship in artifacts


fvanderbiest:
As discussed in https://github.com/georchestra/georchestra/pull/1877#issuecomment-358665646

Target : 16.12 branch, cascade merged into master

PR #1879, opened 2018-01-18
CI - Packaging - use a dedicated variable (#1878)

pmauduit:
See #1878 and #1877

tests: build-time on catalogapp:

```
mvn clean package deb:package -PdebianPackage
```

OK

### DIFF
--- a/analytics/pom.xml
+++ b/analytics/pom.xml
@@ -532,7 +532,7 @@
       <properties>
         <dockerImageName>georchestra/${project.artifactId}</dockerImageName>
         <dockerDatadirScmUrl>scm:git:https://github.com/georchestra/datadir.git</dockerDatadirScmUrl>
-        <dockerDatadirScmVersion>docker-16.12</dockerDatadirScmVersion>
+        <dockerDatadirScmVersion>docker-${packageDatadirScmVersion}</dockerDatadirScmVersion>
       </properties>
       <build>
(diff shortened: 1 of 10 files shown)


fvanderbiest:
Can we use the branch name instead, taking into account the fact that the datadir branch name to use is the same as the geOrchestra branch name ?

pmauduit:
> Can we get the current branch name instead, taking into account the fact that the datadir branch name to use is the same as the geOrchestra branch name ?

It might mess up if a new branch is created (for a PR for example). So we should leave it for the packaging purposes (and not apply the same variable for docker containers, for instance).

fvanderbiest:
What about `<dockerDatadirScmVersion>docker-${packageDatadirScmVersion}</dockerDatadirScmVersion>` ?

pmauduit:
> What about <dockerDatadirScmVersion>docker-${packageDatadirScmVersion}</dockerDatadirScmVersion> ?

ok

fvanderbiest:
Also https://github.com/georchestra/geonetwork/blob/georchestra-gn3-16.12/web/pom.xml#L1027 ?

pmauduit:
See added commit

fvanderbiest:
LGTM

fvanderbiest:
Merged into master with ea72ee774009b2cba113ea219b6ac6415bd8c01c

PR #1880, opened 2018-01-19
SP - fixing runtime error w/ log4j / spring-security version mismatch

pmauduit:
tests: runtime (just jetty:run), integration test added.

PR #1843, opened 2017-12-13
Modify strategy to protect proxy

Vampouille:
This PR try to update strategy used to deny or allow proxy request. 
* Create a new configuration files in datadir : `proxy-permissions.xml` that define permissions of proxy request
* Add domain parameter to Permissions class
* Remove use of urlIsProtected() method

This Pr try to address issue when trying to use another website with same IP but another domain.

PR #1883, opened 2018-01-24
[WIP] Fixing Security-Proxy Spring-Security version bump

pmauduit:
Following previous PR (https://github.com/georchestra/georchestra/pull/1843 and https://github.com/georchestra/georchestra/pull/1880):

* At runtime in georchestra-datadir mode, the webapp instanciation is broken because of a version mismatch between `log4j-over-slf4j` transitively needed by the new version of the cas-client. Fixed by bumping to a newer version of `log4j-over-slf4j`.

* The new version of spring introduced by the previous changes requires now to specify the value parameter of the `RequestMapping` annotation.

Tests: Tested at runtime onto sandbox.georchestra.org. 
* [x] analytics
* [x] extractorapp
* [x] header
* [x] ldapadmin
* [x] mapfishapp
* [x] security-proxy

No deploy issues encountered so far.

Note: The logging management in geOrchestra is currently a bit messy, we might consider migrate to logback, but we don't really have a hand on the transitive dependencies ...

Note 2: The version bump has been done in the 16.12 version, so it might require a backport to this branch.

### DIFF
--- a/analytics/pom.xml
+++ b/analytics/pom.xml
@@ -120,7 +120,7 @@
     <dependency>
       <groupId>log4j</groupId>
       <artifactId>log4j</artifactId>
-      <version>1.2.16</version>
+      <version>${log4j1.version}</version>
       <type>jar</type>
     </dependency>
     <dependency>
--- a/commons/pom.xml
+++ b/commons/pom.xml
@@ -21,7 +21,7 @@
     <dependency>
       <groupId>log4j</groupId>
       <artifactId>log4j</artifactId>
-      <version>1.2.17</version>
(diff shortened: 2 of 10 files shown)


Vampouille:
> Note 2: The version bump has been done in the 16.12 version, so it might require a backport to this branch.

I think this PR should target 16.12 branch instead of master

pmauduit:
> I think this PR should target 16.12 branch instead of master

rebased

pmauduit:
Still in a WIP state, by the way: We have a "405 - method not allowed" when connected onto CAS but a `?login` parameter is passed.

PR #1886, opened 2018-01-30
fix merge of org with name that differ on numeric characters

Vampouille:
and try to display similar orgs


Vampouille:
merged in master with 56fbbf2

Issue #1890, opened 2018-02-02
proxy on master cannot find org.apache.log4j.rolling.RollingFileAppender


fvanderbiest:
proxy_1                     | log4j:ERROR Could not instantiate class [org.apache.log4j.rolling.RollingFileAppender].
proxy_1                     | java.lang.ClassNotFoundException: org.apache.log4j.rolling.RollingFileAppender
proxy_1                     | 	at java.net.URLClassLoader.findClass(URLClassLoader.java:381)
proxy_1                     | 	at java.lang.ClassLoader.loadClass(ClassLoader.java:424)
proxy_1                     | 	at java.lang.ClassLoader.loadClass(ClassLoader.java:357)
proxy_1                     | 	at org.eclipse.jetty.webapp.WebAppClassLoader.loadClass(WebAppClassLoader.java:560)

fvanderbiest:
Also:
geoserver_1                 | ERROR StatusLogger Log4j2 could not find a logging implementation. Please add log4j-core to the classpath. Using SimpleLogger to log to the console...

pmauduit:
Are you using the georchestra/docker composition ? I've got a fix for the first message, but I cannot get how you encountered the second one in GS. There is no log4j2 in the gs classpath, but it seems normal here are the logging-related dependencies present in the GS classpath:

* slf4j-api-1.7.21.jar
* slf4j-log4j12-1.6.4.jar
* log4j-1.2.17.jar
* apache-log4j-extras-1.2.17.jar

So unless a specific log4j configuration file, I cannot see where this reference to log4j2 comes from

fvanderbiest:
OK, let's drop this strange Log4j2 related message.
What's your proposal regarding the RollingFileAppender ?

pmauduit:
> What's your proposal regarding the RollingFileAppender ?

Adding log4j-extras in the same versions as everywhere in georchestra modules (I fixed the version in a previous PR some days ago).

PR #1896, opened 2018-02-07
SP - fix proposal for #1890 (partial)

pmauduit:
Adding the missing log4j dependency so that the class can be resolved.

Tests: build time /runtime only


fvanderbiest:
LGTM

PR #1911, opened 2018-02-26
Remove useless warning about 'anonymous' user not found

Vampouille:


Vampouille:
see also : https://github.com/georchestra/geoserver/pull/11

PR #1925, opened 2018-03-07
Capitalize role name

Vampouille:
This force capitalized name for roles and check role name against following regexp : [A-Z0-9_]+

### DIFF
--- a/console/src/main/java/org/georchestra/console/ws/backoffice/roles/RolesController.java
+++ b/console/src/main/java/org/georchestra/console/ws/backoffice/roles/RolesController.java
@@ -23,6 +23,7 @@ import java.io.IOException;
 import java.util.HashMap;
 import java.util.LinkedList;
 import java.util.List;
+import java.util.regex.Pattern;
 
 import javax.servlet.ServletInputStream;
 import javax.servlet.http.HttpServletRequest;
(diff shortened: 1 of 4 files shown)


console/src/main/java/org/georchestra/console/ws/backoffice/roles/RolesController.java
```diff
@@ -519,12 +523,21 @@ private Role createRoleFromRequestBody(ServletInputStream is) throws IOException
 				throw new IllegalArgumentException(RoleSchema.COMMON_NAME_KEY + " is required" );
 			}
 
+			// Capitalize role name and check format
+			commonName = commonName.toUpperCase();
+			Pattern p = Pattern.compile("[A-Z0-9_]+");
```
tonio: Looks like:
 - `_TRUITE`
 - `TRUITE__VELUE`
 - `TRUITE_`

will be validated but shouldn't. — or maybe it's out of this PR scope? —

fvanderbiest: You're right @tonio, I forgot these ones...

console/src/main/java/org/georchestra/console/ws/backoffice/roles/RolesController.java
```diff
@@ -519,12 +523,21 @@ private Role createRoleFromRequestBody(ServletInputStream is) throws IOException
...
+			if(!p.matcher(commonName).matches())
+				throw new IllegalArgumentException(RoleSchema.COMMON_NAME_KEY +
+						" must only contains upper case letter, digits and underscore" );
```
fvanderbiest: "should only contain uppercased letters, digits and underscores (not leading nor ending with)"

Issue #1932, opened 2018-03-14
[ci/packaging] Differentiate geoserver with and without geofence


landryb:
Right now in https://github.com/georchestra/georchestra/blob/17.12/Makefile#L71 there's an alternative to build geoserver with or without geofence, but that difference is explicitely shown nowhere in the resulting wars/packages.

Would it be possible to suffix this in the war/package name so that both can be provided by a CI without having to make a choice between one or the other ? ie maybe georchestra-geoserver & georchestra-geoserver-geofence (& geoserver/geoserver-geofence for the war)

Up to the ppl deploying to choose the one he/she wants and name it accordingly when dropping the war in tomcat, but at least the choice is provided to the user.

pmauduit:
> but that difference is explicitely shown nowhere in the resulting wars/packages.

First, your war package should contain some `*geofence*` jars in `WEB-INF/libs`. You can test it with `jar tf mywebapp.war` or simplier `unzip -t`.

For the version suffix, you can try passing `-Ddeb.package.version=17.12~geofence` to your maven command (see http://debian-maven.sourceforge.net/usage.html)

landryb:
> For the version suffix, you can try passing -Ddeb.package.version=17.12~geofence to your maven command (see http://debian-maven.sourceforge.net/usage.html)

I'm not sure that would solve the problem as i'm not sure a repository can provide different versions of the same package name..

pmauduit:
`-Ddeb.package.name=geoserver-geofence` then ?

landryb:
Ah, so there's an option in the maven debian plugin.. will try it, but then i think the two packages will need to have a conflict marked as the war names will stay the same, unless there's another maven option to change the war name.

Ultimately this should get commited to the Makefile so that the CI picks it up :)

landryb:
Ah - think i found one way. In https://github.com/georchestra/georchestra/blob/17.12/Makefile#L90 - only the `debianPackage` profile is passed to maven, while if i use `-PdebianPackage,geofence` then the resulting debian package is correctly set to `georchestra-geoserver-geofence`.

Now to figure out how to differentiate the war name...

landryb:
Okay, i have something that works but is a bit of a hack imo:
```
diff --git a/Makefile b/Makefile
index e6ad2651b..72906ae4d 100644
--- a/Makefile
+++ b/Makefile
@@ -89,7 +89,7 @@ deb-build-geoserver: war-build-geoserver

[...]

PR #1937, opened 2018-03-16
Ensure the war and package name are different when using deb-build-ge…

landryb:
…oserver-geofence and war-build-geoserver-geofence Makefile targets (#1932)

Same thing, targeted at 17.12 and 16.12. Welcoming improvements from maven wizards.


landryb:
needs merge to 16.12 and 17.12 :)

Issue #1900, opened 2018-02-08
backgrounds addon - Cannot read property 'id' of null


fvanderbiest:
When the layer is removed, actionning the control again leads to
```
TileManager.js:462 Uncaught TypeError: Cannot read property 'id' of null
    at initialize.clearTileQueue (TileManager.js:462)
    at initialize.removeLayer (TileManager.js:245)
    at initialize.triggerEvent (Events.js:875)
    at initialize.removeLayer (Map.js:1156)
    at S._onSelect (main.js:112)
```
ping @jusabatier

PR #1946, opened 2018-03-21
[Background Addon] Correct removed layer bug

jusabatier:
This correct the bug when layer is manually removed from layerlist.

Cf. https://github.com/georchestra/georchestra/issues/1900


fvanderbiest:
Thanks a lot !

PR #1947, opened 2018-03-22
Add common.properties file at root level of datadir

Vampouille:
This file will provide "default" values for all webapps.
Values can be overrided by webapps.


commons/src/main/java/org/georchestra/commons/configuration/GeorchestraConfiguration.java
```diff
@@ -66,22 +67,15 @@ public GeorchestraConfiguration(String context) {
...
-                activated = false;
-                return;
-            }
+                this.applicationSpecificProperties = this.loadCustomPropertiesFile(context);
+            } catch (Exception e) {}
+
+            // loads common context property file
+            try {
+                this.commonProperties = this.loadPropertiesFile(new File(String.format("%s%s%s",
+                        globalDatadir, File.separator, "common.properties")));
```
fvanderbiest: shared.properties ?

landryb: wont this confuse/bring bad memories about shared.properties in the template build ? :)

fvanderbiest: You're right, but we want to get rid of the config jar ...

fvanderbiest: In addition, you're free not to use this shared.properties file

Vampouille: What about `default.properties` ? This name explicitly explain priority between webapp properties and common/shared/default one

Vampouille: > What about default.properties ?

Done with 0ebecfd

PR #1966, opened 2018-03-27
console - removing dependency on config.jar

pmauduit:
Users willing to customize the webapp now have 2 options:

* Using a datadir (`-Dgeorchestra.datadir`)
* Forking, then rebuilding the webapp

as the georchestra/config has been removed from the build process.

Tests: tested locally + `mvn jetty:run`, no error encountered at runtime so far.


fvanderbiest:
Thanks !

Issue #1924, opened 2018-03-07
console - move console app at webapp root


fvanderbiest:
... so that we don't have `/console/console/` paths but `/console/`

BTW we should pay attention to https://github.com/georchestra/georchestra/blob/8021b1569bdce88ebe84b4e30ebac89ccdbd4093/console/src/main/java/org/georchestra/console/ws/HomeController.java#L84
and https://github.com/georchestra/datadir/blob/5799c0839fb6e1a2d703cda30abacc20fc885102/security-proxy/security-mappings.xml#L12-L13

fvanderbiest:
Work in progress in https://github.com/georchestra/georchestra/tree/console_root

fvanderbiest:
Abandoned in favor of moving to `/console/manager/` with #1976

PR #1976, opened 2018-03-29
Rename console folder

fvanderbiest:
Proposal for #1924 : moving the UI to /console being not a good idea, at least the full path has to change to convey meaning.


fvanderbiest:
goes along with https://github.com/georchestra/datadir/pull/69

PR #1979, opened 2018-04-04
console - Change return value of delegation ws

Vampouille:
Update webservice now return JSON containing delegation description.


tonio:
#worksforme

PR #1988, opened 2018-04-06
console - international log date

fvanderbiest:
... rather than english formatted dates

### DIFF
--- a/console/src/main/webapp/manager/app/components/logs/logs.es6
+++ b/console/src/main/webapp/manager/app/components/logs/logs.es6
@@ -74,5 +74,15 @@ let filterLogs = () => {
   }
 }
 
-angular.module('manager').controller('LogsController', LogsController)
+let logDateFilter = () => {
+  return (date) => {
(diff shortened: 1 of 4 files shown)


console/src/main/webapp/manager/app/components/logs/logs.es6
```diff
@@ -74,5 +74,15 @@ let filterLogs = () => {
...
 }
 
-angular.module('manager').controller('LogsController', LogsController)
+let logDateFilter = () => {
+  return (date) => {
+    let format = 'YYYY-MM-DD HH:mm'
+    return moment(date).format(format)
+  }
+}
+
```
tonio: one-liner:

```javascript
const logDateFilter = () => date => moment(date).format('YYYY-MM-DD HH:mm')
```

fvanderbiest: Impressive !

fvanderbiest:
> Just not sure the filter is declared in the best place — because used on home page, too —.

What would you advise ?

PR #1992, opened 2018-04-10
SP - fixing parameters passed in GET

pmauduit:
Following the SP revamp, every parameters passed as in a form-encoded POST were also passed as GET parameters to the security-proxified webapps, causing problems with every url-formencoded based pages (CAS server login if behind the SP and geonetwork) .

This PR fixes the logic, so that only parameters present in the URI are copied.

Also there was a potential issue with UTF-8 accented characters, that this commit shall address (only the value of the parameter was encoded).

Tests:
* Unit tests added
* Runtime tested on Jetty using a simple controller
* Runtime tested on Tomcat (sandbox.georchestra.org)

### DIFF
--- a/security-proxy/src/main/java/org/georchestra/security/Proxy.java
+++ b/security-proxy/src/main/java/org/georchestra/security/Proxy.java
@@ -42,6 +42,8 @@ import java.nio.charset.Charset;
 import java.util.ArrayList;
 import java.util.Collections;
 import java.util.Enumeration;
+import java.util.HashMap;
+import java.util.Hashtable;
 import java.util.List;
 import java.util.Map;
 import java.util.Properties;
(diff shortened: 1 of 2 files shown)


security-proxy/src/main/java/org/georchestra/security/Proxy.java
```diff
@@ -464,6 +474,45 @@ private String buildForwardRequestURL(HttpServletRequest request) {
...
+                String value = p.getValue();
+                // Skip login & ticket parameter
+                if (name.equals(ServiceProperties.DEFAULT_CAS_ARTIFACT_PARAMETER) || name.equals("login")) {
+                    continue;
+                }
+                if (query.length() > 1) {
+                    query.append('&');
+                }
+                if (value == null) {
+                    query.append(name);
```
Vampouille: Here you can factorize `query.append(name)`

Vampouille:
@pmauduit Can you also fix unused import as mentioned by codacy ?

fvanderbiest:
Does it require a backport to 17.12 ? 
I do not remember when this SP revamp happened !

pmauduit:
> I do not remember when this SP revamp happened !

master after the 17.12 release

PR #1994, opened 2018-04-11
Console - Filter webservices

Vampouille:
This PR adds filtering on all webservices and also check authorizations for webservices that modify the model.


console/src/main/java/org/georchestra/console/dao/AdminLogDao.java
```diff
@@ -38,4 +40,6 @@
 
     List<AdminLogEntry> findByTarget(String target, Pageable range);
 
+    List<AdminLogEntry> myFilter(@Param("targets") Collection<String> targets, Pageable range);
```
fvanderbiest: `myFilter` is not a very explicit method naming

console/src/main/webapp/manager/app/styles/app.less
```diff
@@ -5,12 +5,6 @@
 @tablet: 768px;
 @large: 992px;
 
-
-@media (min-width: @tablet) {
```
Vampouille: @tonio Are you agree with this deletion ?

console/src/main/java/org/georchestra/console/dao/AdminLogDao.java
Vampouille: Done with 8ae2b5d

console/src/main/webapp/manager/app/components/area/area.es6
```diff
@@ -89,6 +85,17 @@ class AreaController {
...
       })
+
+      if (!config.map) {
+        this.map.getView().fit(vector.getSource().getExtent(), map.getSize())
+      } else {
+        let view = new ol.View({
+          center: ol.proj.fromLonLat(config.map.center),
+          zoom: config.map.zoom
+        })
+        this.map.setView(view)
```
tonio: You can just update view center & zoom rather than creating a new one.

console/src/main/webapp/manager/app/styles/app.less
Vampouille: I tested pages with few content, it looks  good.

console/src/main/webapp/manager/app/components/area/area.es6
Vampouille: Done with : ec79291

PR #1995, opened 2018-04-12
GS - bumping to 2.13.0

pmauduit:
There are mainly one issue detected at runtime in case of geofence usage: cache invalidation is not working.

Also while runtime testing on GGE: there seemed to be an issue with their datadir, because the REST interface on `layer.xml` was broken. No idea which layer was causing this.

Note: GS2.13.0 is needed for a certain GS extension only available in GS >= 2.12.3.

### DIFF
--- a/geoserver/pom.xml
+++ b/geoserver/pom.xml
@@ -11,8 +11,8 @@
   <name>GeoServer 2.x root module</name>
   <properties>
     <maven.test.skip>true</maven.test.skip>
-    <gs.version>2.12.1</gs.version>
-    <gt.version>18.1</gt.version>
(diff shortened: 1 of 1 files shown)


fvanderbiest:
> cache invalidation is not working.

This sounds rather critical.

> GS2.13.0 is needed for a certain GS extension only available in GS >= 2.12.3

Which ones ?

pmauduit:
> Which ones ?

the importer plugin and/or the WPS one, I am not aware of every details or why this version bump is needed (requested by a C2C partner)

pmauduit:
PR updated, should be good to go now, for runtime testing.

PR #1996, opened 2018-04-12
SP - encode parameter names and values in query string (rework on partial fix in #1992)

pmauduit:
This addresses the partial fix from the previous PR (#1992), which transmits parameters without URL-encoding them, which causes issues with GeoNetwork as it sometimes sends parameter values with spaces.

Note: Every servlet containers and/or webservers should understand url-encoded characters, even if the HTTP spec allows the use of UTF-8 chars, I could not manage to encode problematic characters like space and in the same time leaving accented characters.

Tests: 
* utests added / adapted
* runtime tests directly onto a deployed env.


pmauduit:
This breaks already-encoded queries in GeoNetwork, see https://github.com/georchestra/georchestra/issues/2018

PR #1999, opened 2018-04-16
mfapp - Docker - remove the war after having unzipped it

pmauduit:
It seems that there are some ambiguities if the war is left into the webapps/ subdir, being unzipped by Jetty in its own tempdir and used in place of the actual already unzipped webapp.

This shall address issues with mfprint PNG output format (GSEST-191).


mapfishapp/src/docker/Dockerfile
```diff
@@ -14,6 +14,7 @@ RUN apt-get update && \
    rm -rf /var/lib/apt/lists/*
 
 RUN unzip -d /var/lib/jetty/webapps/mapfishapp /var/lib/jetty/webapps/mapfishapp.war
+RUN rm -f /var/lib/jetty/webapps/mapfishapp.war
```
fvanderbiest: Better in a one-liner ?

PR #2024, opened 2018-04-25
Console - Use ngPluralize to adapt delegations list title

### LINKED ISSUE #1989: console - use ngPluralize
https://docs.angularjs.org/api/ng/directive/ngPluralize would be useful to prevent
capture du 2018-04-06 14-30-28

Vampouille:
Fix #1989 
Please do not merge because this PR need translations.


console/src/main/webapp/manager/app/components/delegations/delegations.tpl.html
```diff
@@ -4,7 +4,13 @@
 
     <input type="text" class="filter-table form-control pull-right" ng-model="delegations.q" placeholder="{{'delegations.filter' | translate}}"></input>
     <i class="glyphicon glyphicon-remove-sign filter-table-reset" ng-show="delegations.q != ''" ng-click="delegations.q = ''"></i>
-    <h3 class="delegations-title">{{delegations.delegations.length}} <span translate>delegations.title</span></h3>
+    <h3 class="delegations-title">
+      <ng-pluralize count="delegations.delegations.length"
+                    when="{'0': '{{ &quot;delegations.title_no_delegations&quot; | translate }}',
+                           '1': '{} {{ &quot;delegations.title&quot; | translate }}',
+                           'other': '{} {{ &quot;delegations.title&quot; | translate }}s'}">
```
tonio: Not sure suffixing with `s` will be correct for languages other than `fr` or `en`

Vampouille: Another option is to add a translation for plural, but this will increase workload of translators. @fvanderbiest your opinion ?

PR #2025, opened 2018-04-25
Console - Allow roles to be renamed

### LINKED ISSUE #1822: ldapadmin console - roles cannot be edited
They can be deleted, but their name cannot be changed currently.
Provide a way to edit them.

Vampouille:
Fix #1822
* [x] TODO: Change ldap command to avoid deleting and recreating role
instead of using ldap rename command

### DIFF
--- a/console/src/main/webapp/manager/app/components/role/role.es6
+++ b/console/src/main/webapp/manager/app/components/role/role.es6
@@ -18,15 +18,17 @@ class RoleController {
     translate('role.deleted', this.i18n)
     translate('role.deleteError', this.i18n)
 
-    this.role = $injector.get('Role').get({id: $routeParams.role})
-    this.originalID = $routeParams.role
+    // Save original cn in case we change cn, because we need to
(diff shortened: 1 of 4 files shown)


Vampouille:
The memberOf attribute is not correctly updating when deleting role

console/src/main/webapp/manager/app/components/role/role.es6
```diff
@@ -19,14 +19,21 @@ class RoleController {
     translate('role.deleteError', this.i18n)
 
     this.role = $injector.get('Role').get({id: $routeParams.role})
```
tonio: you can use a function as a second argument to get to avoid storing `orginalID` in the controller

Vampouille: Done with https://github.com/georchestra/georchestra/pull/2025/commits/0b241d87c7e16e53f2f68ee91cdb8d66ced72f61

PR #2026, opened 2018-04-26
Console - Fix loading of email configuration

### LINKED ISSUE #2011: console - user holding a delegation on Org should be notified by email when another user creates account in his Org
I tested on master and it seems this is not working as expected.
Only SUPERUSERs are notified.

Vampouille:
It seems that we cannot populate a Properties and then construct Session
object with the Properties instance. So we need to first create Session
instance and then use setProperty() method on it.

It also seems that we need to use a property called `mail.smtp.port`
instead of `mail.protocol.port`.

Fix #2011


fvanderbiest:
3 times same code ?

Issue #1904, opened 2018-02-13
SP - URIs with a "url" parameter are not caught correctly by the security-proxy


pmauduit:
When a URL contains a parameter named "url", the SP is not able to wire to the correct method in proxy.java, this has been reproduced on GGE environment and also mentionned via IRC by Le-Puy-En-Velay.

reproduced on sdi-stable.georchestra.org:

https://sdi-stable.georchestra.org/mapfishapp/pdf/info.json?url=%2Fmapfishapp%2Fpdf
Status Code:405 Method Not Allowed

As a result, the mapfish-print from mapfishapp is broken.

fvanderbiest:
A priori fixed with #1902 IIRC.
@pmauduit do you confirm ?

fvanderbiest:
yes

pmauduit:
The bug is still present in 17.12, 16.12.

PR #2045, opened 2018-05-21
SP - fixing 405 on GET '?url=...'

pmauduit:
Waiting for the SP rewrite in master to override these legacy versions.

Shall address #1904 

Tests: runtime on geop-dev, still possible to login, use the proxy for loading external layers and obviously the '/mapfishapp/pdf/info.json?url=...' reported to fail.

Note: not to be cascade-merged into master, since there is still a WIP with a spring / spring-security update, and the Proxy.java code slightly evolved.


fvanderbiest:
Thanks Pierre ! 
Please port to later branches as well...

PR #1944, opened 2018-03-20
console - fixing JS build via brunch

pmauduit:
Encountered several issues when building, Like:
https://github.com/brunch/sass-brunch/issues/168

```
[ERROR]
[ERROR] npm ERR! code ELIFECYCLE
[ERROR] npm ERR! errno 1
[ERROR] npm ERR! node-sass@3.8.0 postinstall: `node scripts/build.js`
[ERROR] npm ERR! Exit status 1
[ERROR] npm ERR!
[ERROR] npm ERR! Failed at the node-sass@3.8.0 postinstall script.
[ERROR] npm ERR! This is probably not a problem with npm. There is
likely additional logging output above.
[ERROR]
```

Tests: in a s2i image, as well as locally.

PR #2047, opened 2018-05-22
LDAPAdmin - fixing npm / brunch build

pmauduit:
Almost the same fix as https://github.com/georchestra/georchestra/pull/1944

* Fixing version for save-svg-as-png
* Upgrade to newer version of npm/node so that we can add the package-lock.json file

Tests: compilation OK, runtime tested directly on PPIGE-dev

Note: also to be cascade-merged on 17.12, but not on master

### DIFF
--- a/ldapadmin/pom.xml
+++ b/ldapadmin/pom.xml
@@ -611,8 +611,8 @@
           </execution>
         </executions>
         <configuration>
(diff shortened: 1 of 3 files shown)


fvanderbiest:
Thanks Pierre, please do the merges.

PR #2056, opened 2018-05-25
remove docker-compose files to eliminate duplication with georchestra/docker

fvanderbiest:
These files offer nothing more than https://github.com/georchestra/docker
We should remove them because it's error-prone.


fvanderbiest:
In addition, the fact that this compose file targets images tagged latest (built from master branch) is highly questionable

Issue #1417, opened 2016-07-01
Remove the need for a config jar


fvanderbiest:
... except for client-side overrides.

Eg: currently, having a config jar is to my knowledge the only way to customize the header image...

pmauduit:
The main issue with config.jar is that it can erase versioned files, since it basically follows the current workflow:

1. the config.jar is generated
2. copied into ~/.m2 repository
3. on each webapps which requires it, it is unpacked into <module>/target/
4. maven then copies <module>/target/generated-config/... onto <module>src/ 

Erasing files into src is not a good practice, and is even dangerous (if someone git add / commit / push sensitive informations).

If we remove the config.jar, we would need to find another way to override / add java sources though.

fvanderbiest:
> If we remove the config.jar, we would need to find another way to override / add java sources though.

I'm not sure we really need to override java sources.

pmauduit:
> I'm not sure we really need to override java sources.

If we do, the config.jar is not a way (in term of maven / java sources management) to do it correctly.

fvanderbiest:
First step is https://github.com/georchestra/template/pull/30 (not to be merged for now)

fvanderbiest:
Second step happens gradually in branch https://github.com/georchestra/georchestra/tree/datadir-only

fvanderbiest:
This won't make it for 16.12 unfortunately.

fvanderbiest:
Apps:
- [ ] console
- [ ] mapfishapp
- [ ] extractorapp
- [ ] analytics
- [ ] atlas
- [ ] cas
- [ ] geonetwork
- [ ] geoserver
- [ ] geowebcache
- [ ] header
- [ ] proxy

severo:
@fvanderbiest shouldn't https://github.com/georchestra/georchestra/tree/datadir-only branch be deleted?

severo:
I understand the idea of this issue, and I approve it, mainly because it's a lot simpler to understand for the users that will install and configure an instance. But it's important to note that we lose one of the properties: avoiding the redundancy. See https://github.com/georchestra/georchestra/tree/master/config:

> (...) The second problem is that many projects have shared configuration parameters like database urls.
> 
> There are two parts that are designed to overcome these problems:
> 
[...]

PR #2082, opened 2018-06-14
remove config.jar dependency for mapfishapp

fvanderbiest:
Proposal for #1417


fvanderbiest:
missing bits fixed with 7f294367aafb860a2fb998eb6f6e5fcd844cfffd

PR #2033, opened 2018-04-30
Factorize public url in default.properties

Vampouille:
Done in : 
* extractorapp (basic startup test, need more tests)
* console (unit test fixed, run time tested)

PR #2090, opened 2018-06-26
SP - fix for pr 2033

pmauduit:
In #2033, the official docker-composition was not working, previous PR introduced a regression, making the Proxy bean unable to resolve the public URL variable, following error was arising:

```
org.springframework.beans.factory.BeanCreationException: Error creating
bean with name 'proxy' defined in ServletContext resource
[/WEB-INF/proxy-servlet.xml]: Invocation of init method failed; nested
exception is java.net.MalformedURLException: no protocol: ${publicUrl}
```

The main idea behind this PR is to let the logic of variable resolution to Spring, instead of trying to do it from our side into georchestra-commons.

Tests: runtime in a docker compo.

### DIFF
--- a/security-proxy/src/main/java/org/georchestra/security/Proxy.java
+++ b/security-proxy/src/main/java/org/georchestra/security/Proxy.java
@@ -142,6 +142,7 @@ public class Proxy {
      */
     private String defaultTarget;
     private String publicHostname = "https://georchestra.mydomain.org/";
+
     private Map<String, String> targets = Collections.emptyMap();
     private HeadersManagementStrategy headerManagement = new HeadersManagementStrategy();
     private FilterRequestsStrategy strategyForFilteringRequests = new AcceptAllRequests();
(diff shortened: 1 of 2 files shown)


security-proxy/src/main/java/org/georchestra/security/Proxy.java
```diff
@@ -142,6 +142,7 @@
      */
     private String defaultTarget;
     private String publicHostname = "https://georchestra.mydomain.org/";
```
fvanderbiest: IIRC `publicHostname` should not include a final slash, see eg https://github.com/georchestra/datadir/blob/509185f550cab4ebc098baf76fc5a5a1d8ae53d3/default.properties#L3-L5

fvanderbiest:
Thanks Pierre, this looks good to me, except maybe for the final slashes.

pmauduit:
Ok updated

PR #2119, opened 2018-07-04
mfapp - print-lib 2.1.2

pmauduit:
Moving to minor version 2.1.2 for mfprint to fix a race condition encountered by renne-metropole.

Tests: totally untested - compilation OK so far -, a similar fix has been already applied on rennes's portail-test.


fvanderbiest:
Awaiting runtime tests before merge.

Issue #2061, opened 2018-05-30
console - moderatorEmail should be removed on master


fvanderbiest:
Email recipients are now dynamically selected based on their roles: SUPERUSER and ORGADMIN
```
fvanderbiest@vaio:~/workspace/georchestra/georchestra (master *$)$ g grep moderatorEmail
console/src/main/java/org/georchestra/console/bs/Moderator.java:        private String moderatorEmail = "moderator@mail";
console/src/main/java/org/georchestra/console/bs/Moderator.java:                return moderatorEmail;
console/src/main/java/org/georchestra/console/bs/Moderator.java:        public void setModeratorEmail(String moderatorEmail) {
console/src/main/java/org/georchestra/console/bs/Moderator.java:                this.moderatorEmail = moderatorEmail;
console/src/main/java/org/georchestra/console/mailservice/MailService.java:     public void sendNewAccountRequiresModeration(ServletContext servletContext, final String uid, final String userName, final String userEmail, final String moderatorEmail) {
console/src/main/java/org/georchestra/console/mailservice/MailService.java:                     NewAccountRequiresModerationEmail email = this.emailFactory.createNewAccountRequiresModerationEmail(servletContext, userEmail, new String[]{moderatorEmail});
console/src/main/webapp/WEB-INF/spring/console.properties:#   * an email is sent to moderatorEmail for validation,
console/src/main/webapp/WEB-INF/spring/console.properties:moderatorEmail=georchestra+testadmin@georchestra.mydomain.org
console/src/main/webapp/WEB-INF/spring/webmvc-config.xml:    <property name="moderatorEmail" value="${moderatorEmail}"/>
```

fvanderbiest:
TODO: clean remaining references to `moderatorEmail` in the main repo and datadir

landryb:
Does that mean that all users in the 'SUPERUSER' group will receive a notification for a pending new account, and users in the 'ORGADMIN' group will receive a notification if a pending new account registered with their org ?

PR #2124, opened 2018-07-11
Console - send registration email to all admins

Vampouille:
and delegated admins if any. So moderatorEmail config param in datadir is no more used.

This commit also:
* Refactor email related classes to have a more readable code
* Use replyTo variable to fill Reply-To: mail header
* Create a findByRole() method in AccountDao to replace findUsers() in
  RoleDao that does not work as expected
* Remove notion of language in email (language parameter in datadir)
* Add a new parameter in datadir to declare encoding of mail template:
  `templateEncoding=UTF-8` and correctly handle file encoding
* Convert emailHtml field to boolean value
* Improve templating system to be able to replace variable multiple
times
* Add unit tests

Linked with this PR on datadir : https://github.com/georchestra/datadir/pull/83

Should fix https://github.com/georchestra/georchestra/issues/2061


Vampouille:
There is some NPE on some unit test, we need to fix this before merging

PR #2128, opened 2018-07-17
root pom - update to new version of the maven plugin

pmauduit:
Needed to avoid 400 bad requests from maven-central


fvanderbiest:
Should it be backported to earlier branches as well ?

PR #2130, opened 2018-07-18
Proxy - Override referer only on proxy request (/proxy/?url=...)

Vampouille:
Don't override referer for request to internal geOrchestra webapps.


fvanderbiest:
Nice bugfix !

Issue #2080, opened 2018-06-14
console - org and role pages should display user list & enable membership management


fvanderbiest:
On pages `/console/manager/#!/org/XXX/infos` and `/console/manager/#!/role/XXX/infos` (or maybe in another tab) an administrator should be able to see & manage member users.

fvanderbiest:
GeoBretagne and CRAIG have a strong interest for this feature.

landryb:
Now that i started playing with the console in master, this is definitely a feature we need. And since the members of an org/role are already stored in the ldap, i guess we just need the UI glue ?

fvanderbiest:
> i guess we just need the UI glue ?

That's it. Maybe custom services will be required to handle membership in batch, I don't know ATM.

fvanderbiest:
Could you specify here the complete requirements for this UI ? That would help a lot :-)

landryb:
Well i guess on an org/role page, being able to see the current member list, with a red tick box next to each of them to remove a member from the org/role, and an autocomplete field to add members not yet in the current org/role ? I'll discuss it tmrw with my colleagues to draft something.

Dunno if there are some integrity checks to do (ie not being able to remove a delegated admin from the group he admins ? the manager thing ?)

landryb:
from an org/role page, we need to:
* see the member list (enventually with pagination for roles with hundreds of users, ie USER ?)
* the member name is a link to the user page
* as it's done for org membership on the user page
  * a cross allows to remove a member
  * an autocomplete field/dropdown allows to add a member

while here, would there be a way for a user to *see* the details of his org (ie coverage, member list, etc...) ? right now he only sees the org name.

landryb:
Dupe of #1590 but with a broader scope, also applying to roles ?

fvanderbiest:
right

fvanderbiest:
> while here, would there be a way for a user to see the details of his org (ie coverage, member list, etc...) ? right now he only sees the org name.
> 
> Sure, but this would qualify for a separate issue (so we have room to discuss it in details). Thanks :-)

To be discussed on #2125

PR #2154, opened 2018-09-07
Manage users in role/org forms

tonio:
See https://github.com/georchestra/georchestra/issues/2080


fvanderbiest:
LGTM, thanks !
Can you also update the translation files in this PR ?

fvanderbiest:
May I ask @severo to double check the spanish translations ? Many thanks.

landryb:
Note that i'm doing my testing with a manually migrated user database that was still using the previous `ou=groups` hierarchy, on a dev environment, so some users might be invalid here and there... but i have groups where the member list in ldap matches what the `private/roles/XXXX` API returns, and the total member amount is sometimes off by 20 (ie 551->531). Some filtering being done (i see the `role.associate` call?) on the content of `private/users` reply ? And yes, i have `olcMemberOfRefInt:TRUE` in my slapd, so the database **should** have a consistent content.

landryb:
Okay, i found out where the problem is, and it's a tricky one. Comparing the 'full' user list and the list actually shown in the role list, we have duplicate **valid** users in the sense that they have **several** accounts (ie `prenom.nom` and `prenom.nom1`, and sometimes `prenom.nom2`) but the LDAP entries have more or less the same givenName/sn (with differences in capitalizations, but you get the idea)

In that case, it seems the console only shown **one** entry in the user tab/list or in the role member list, while the two distinct uids are in the role.. and that might explain the differences in member total.

For example, for the role where i had 81 entries in `private/roles/XXXX` reply, in those i have two uids `admin` and `breuil` but both show up as 'Breuil Landry' and are `merged` by the UI:

```
$ldapsearch -x -h localhost -b dc=craig,dc=fr sn=BREUIL uid givenName SN

dn: uid=breuil,ou=users,dc=craig,dc=fr
uid: breuil
givenName: Landry
sn: Breuil

dn: uid=admin,ou=users,dc=craig,dc=fr
uid: admin
sn: Breuil
givenName: Landry
```

So this is imo a distinct (real, and already existing ?) issue from the one at hand here.

console/src/main/webapp/manager/app/assets/lang/es.json
```diff
@@ -34,6 +34,7 @@
   "tab.logs"              : "Logs",
   "tab.manage"            : "Administrar",
   "tab.area"              : "Zona de competencia",
+  "tab.users"             : "Miembro",
```
severo: Miembros

console/src/main/webapp/manager/app/assets/lang/es.json
```diff
@@ -96,6 +97,9 @@
   "org.select"            : "Seleccionar una entidad",
   "org.membersCount"      : "Miembros",
   "org.ohasdeleg"         : "Los usuarios de esta organización pueden ser administrados por los siguientes usuarios:",
+  "org.userremoved"       : "El usuario ha sido eliminado de la organización.",
+  "org.useradded"         : "El usuario se añadió a la organización.",
```
severo: El usuario ha sido añadido a la organización.

console/src/main/webapp/manager/app/assets/lang/es.json
```diff
@@ -119,6 +123,9 @@
   "role.cn"               : "Nombre",
   "role.favorite"         : "Favorite",
   "role.favorites"        : "Favorites",
+  "role.useradded"        : "El usuario se añadió al rol.",
```
severo: El usuario ha sido añadido al rol.

fvanderbiest:
> So this is imo a distinct (real, and already existing ?) issue from the one at hand here.

Yep. 
How do these users appear in the /console/private/users service ?

console/src/main/webapp/manager/app/components/role/role.tpl.html
```diff
@@ -30,6 +30,53 @@ <h4>
...
+            </select>
+          </div>
+        </form>
+
+        <table class="table table-striped table-condensed">
+
+          <thead>
+            <tr>
+              <th translate>users.user</th>
+              <th translate>user.role</th>
```
fvanderbiest: I'm not sure this is the correct translation key for a "remove" action

fvanderbiest: ping @tonio

console/src/main/webapp/manager/app/components/role/role.tpl.html
```diff
@@ -64,7 +64,7 @@ <h3 class="roles-title">
                 <a ng-link="user({id: user.uid, tab: 'infos'})">{{::user.sn}} {{::user.givenName}}</a>
               </td>
               <td>
-                <a href="javascript:void(0)" ng-click="role.associate(user.uid, true)" class="text-danger">
+                <a href="javascript:void(0)" ng-click="role.associate(user.uid, true)" class="text-danger" title="{{role.i18n.remove}}">
```
fvanderbiest: missing role.i18n.remove and org.i18n.remove in translation files ?

landryb:
Retested this, 👍 for me feature-wise.

Issue #2100, opened 2018-06-29
console: import/export org area as csv


landryb:
would be nice to be able to export/import the code list as csv (or geojson since we have the geometry available ? but then the file is huge) so that one can *replace* the area by uploading a csv.

we have cases where every year, an admin has to compare a list of 300 insee codes (what's in the db/ldap, what's the new list) , with 90 to add and 20 to remove....

landryb:
> Indeed.
> Can you iterate on your requirements for uploading / downloading this CSV ? (mostly: who, where, what...)
On the coverage page, a link would export the current codelist as csv (eventually with a heading, but the codes are not always inseeCodes..) - cadastrapp has that kind of feature for parcel exports. This should be available to anyone (i'm not sure it makes sense to keep that information secret, but right now the org pages arent available to users anyway)

On the same page an upload form/button with 'replace codelist' would allow to send a similarly formed csv (potentially removing the heading, if it's not a number ?), available for an ORGADMIN or a SUPERUSER, and per #2101 eventually available for a member user, but with a validation step.

Uploading the codelist would *replace* the current codelist, which is simpler than trying to mix/match/add existing vs new.

fvanderbiest:
IIUC, you mean a dumb CSV (actually without comma) like this one:
```csv
code
75001
75002
...
```

landryb:
Yes, no point in having more than this, we just need a codelist. Eventually with area names but only for informational purposes, as uploading the codelist should only care about the codes/keys.

PR #2159, opened 2018-09-10
Add support for import/export org's area as CSV

tonio:
See #2100

### DIFF
--- a/console/src/main/webapp/manager/app/assets/lang/de.json
+++ b/console/src/main/webapp/manager/app/assets/lang/de.json
@@ -148,8 +148,8 @@
   "area.all"              : "Alle",
   "area.none"             : "Keine",
   "area.search"           : "Suchen",
-  "area.import"           : "**TODO**",
-  "area.export"           : "**TODO**",
+  "area.import"           : "CSV importieren",
+  "area.export"           : "CSV exportieren",
 
   "role.allusers"         : "Alle Benutzer",
   "role.system"           : "geOrchestra-Rollen",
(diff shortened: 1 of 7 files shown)


landryb:
Basic usage works fine in FF 63 (ie export current area, get the codelist, add an existing value, reimport it, the added value correctly shows up), but yes please refrain from using 'too recent' js features like this. It should at least work on ESR branches for firefox i'd say.. we have to take into account potential new users with old IEs around..

Some nits:
* technically, it doesnt really support CSVs, it only supports *one entry per line*-style files. Tried using a file with two field (i know, this is not in what was asked for, but you know, sometimes users will try this..), separated by `;` and all entries were ignored without user feedback/error message. We should at least have user feedback with a helpful message when no entries are imported/recognized (ie *one insee code per line* popup or something like that)
* lines with content not matching codes are ignored, so that properly handles the classic CSV-style with a heading
* i tried using *import CSV* on `/console/account/new` but that directly lead to a java NPE, unrelated to the feature at hand but awkward:
```
2018-09-10 16:23:37 GlobalExceptionHandler [ERROR] java.lang.NullPointerException
        at org.georchestra.console.ws.utils.PasswordUtils.validate(PasswordUtils.java:51)
        at org.georchestra.console.ws.newaccount.NewAccountFormController.create(NewAccountFormController.java:192)
```
* once the fields filled with garbage, i retried, and it seems clicking on *import CSV* triggers the form validation method on the account creation page in all cases, which folds the *new org* section, and in the end one cant use the import CSV feature. Does it also make sense to allow potential new users to *export CSV* here ? i'm not sure..

landryb:
> Please disable the feature on the `/console/account/new` page. Thanks !

Well being able to import CSV for new account/org requests *could* be a valid usecase ....

fvanderbiest:
Wow, you're expecting really power users ;-)

Issue #1689, opened 2017-03-23
console - does not allow login renaming


fvanderbiest:
https://github.com/georchestra/georchestra/commit/d8ce11ee101d08b87b11311f2992b50283af5c9c introduced the possibility to change a user login but this was not implemented in the console, only in privateui (which is now deprecated)

landryb:
linked with #2104, yes please. Right now we want to enforce a login pattern policy, and if we cant we want at least to 'manually' apply this policy (renaming the user) before validating it.

landryb:
Completes what was done in #1109 (backend part)

PR #2160, opened 2018-09-11
Add support to change user login

tonio:
See #1689 

Looks like there's no format check for uid on the server-side… I can add one client-side if needed but it won't be enough imho.

### DIFF
--- a/console/src/main/webapp/manager/app/components/user/user.es6
+++ b/console/src/main/webapp/manager/app/components/user/user.es6
@@ -20,6 +20,7 @@ class UserController {
     this.tab = $routeParams.tab
 
     this.user = User.get({id: $routeParams.id}, (user) => {
+      user.originalID = user.uid
       if (user.org && user.org !== '') {
         user.orgObj = Orgs.get({'id': user.org}, org => {
           user.validOrg = org.status === 'REGISTERED'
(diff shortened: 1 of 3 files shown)


landryb:
Basic testing, works as expected, and correctly renames the LDAP object via MODRDN. What do you mean by 'format check for uid' ? Ie only numbers/letters/point & no crazy special chars? `makeUid` in the account creation JSP page does some but its clientside and doesnt 'fix' what a user could put in the field... cf https://github.com/georchestra/georchestra/blob/master/console/src/main/webapp/WEB-INF/views/validation.jsp#L225 too

tonio:
For now, what you give to server is saved, regardless of symbols/accentuated chars in it. I think there should be restrictions, at least the same as the one in `makeUid`.

fvanderbiest:
> I can add one client-side if needed

Yes, I think we should have the same checks on uid everywhere. 

https://github.com/georchestra/georchestra/blob/b069c0499c269d86f17a28d987ca88061ed58a83/console/src/main/webapp/WEB-INF/views/validation.jsp#L225-L240 looks good to me.

landryb:
Add a comment in both places reminding the next developer who changes the rule to update the other spot :)

landryb:
Retested tip of this branch, renaming works fine, uid is changed, and group membership is updated, ie the roles & org objects have the 'new' dn as member; :+1: to merge.

A comment linking the two places would still be welcome :)

Issue #2104, opened 2018-07-02
console/registration: allow to configure & enforce a login pattern


landryb:
Right now the registration panel uses the makeUid() function from https://github.com/georchestra/georchestra/blob/master/console/src/main/webapp/WEB-INF/views/createAccountForm.jsp#L179 which hardcodes 'first letter of name + surname'.

This has some issues:
- an org might want to apply username policies (here, name.surname)
- the field allows the user to choose his username, overriding the return value of makeUid()
- if the login the user choosed doesnt match the username policies, the validation step in the admin part of the console doesnt allow to 'fix' the username as the field is readonly (ie no way to 'rename' an ldap entry)

It would be nice to have two extra config fields:
- a boolean 'enforce username policy' that makes the login field readonly
- a string 'username pattern' allowing the admin to configure how the username should be built from the various fields available in the context

some examples that come to my mind:
  * `${name}.${surname}`
  * `${email}` (do we support usernames with uid=xxx@bar.com or that breaks other assumptions elsewhere ?)
  * `${name:[0-1]}${surname:[0-5]}` (to take substrings, ie RAymond BIDOCHon -> rabidoch)

landryb:
If the username policy is enforced, the message shown when the user tries to register with an already existing login should not show 'Please select another' :) The form already gracefully handles this case as it appends a number to the pattern-generated value (ie testuser1)

deneux:
Note to self: https://github.com/georchestra/georchestra/blob/18.06/console/src/main/webapp/WEB-INF/views/createAccountForm.jsp#L182 is where the current pattern is hardcoded.

fvanderbiest:
Still relevant

PR #2162, opened 2018-09-11
Allow an administrator to set the login field readonly on the account…

landryb:
… creation page (#2104)

setting `userCantChooseUid=true` in `console.properties` makes the field readonly.

A bit rough around the edges since my JSP isnt that good, but does what's expected, tested both with/without setting the parameter in `console.properties`. This is only a part of #2104...

### DIFF
--- a/console/src/main/webapp/WEB-INF/tags/input.tag
+++ b/console/src/main/webapp/WEB-INF/tags/input.tag
@@ -7,6 +7,7 @@
 <%@attribute name="cssClass" required="false" type="java.lang.String"%>
 <%@attribute name="label" required="false" type="java.lang.String"%>
 <%@attribute name="required" required="false" type="java.lang.Boolean"%>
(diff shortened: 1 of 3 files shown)


landryb:
On top of this i need to amend the message shown when the uid is already taken (ie https://github.com/georchestra/georchestra/issues/2104#issuecomment-402191063 - returning a different message than `uid.error.exist` if `userCantChooseUid=true`), will pile on top of this PR if deemed interesting.

landryb:
Friendly poke about this PR. I can also do one for `console/console.properties` in the datadir, but would need feedback on this one first :)

fvanderbiest:
LGTM but I don't like the var name `userCantChooseUid` -> `readonlyUid` instead ?

I'm surprised that setting the field to readonly still allows javascript updating it.

landryb:
rebased on master, & renamed var

fvanderbiest:
LGTM, thanks !

Issue #2165, opened 2018-09-12
docker / Jetty - missing Jetty module leading to uneeded redirections on regular http


pmauduit:
# Bug Description

geOrchestra sometimes has to send HTTP redirects, e.g. when calling the default target on the SP:

```
pmauduit@wrk51:~$ curl -i https://www.geograndest.fr/
HTTP/1.1 302 Found
[...]
Location: http://www.geograndest.fr/portail/
[...]
```
the header location here involves an extra redirect on http which will be caught by the frontend webserver to actually redirect onto the https endpoint. The main reason behind this behaviour, is that the SP sends a relative URL in the Location header, see here for example:
https://github.com/georchestra/georchestra/blob/18.06/security-proxy/src/main/java/org/georchestra/security/Proxy.java#L346-L358

This Location header is sent by the webapp but potentially readapted by the servlet container. Tomcat allows the use of relative redirects sent to the browser, whereas jetty sends absolute urls by default (The http spec seemed to have evolved in regards to this point, first stating that absolute urls were mandatory, then allowing the browser to determine the real meaning behind the relative URL). This has already been described in another issue, but I cannot find it back.

The problem here is that an extra jetty module is required, so that the location header is correctly computed. To ensure a correct computing of the URL, we need to activate the `http-forwarded` module in the different dockerfiles of the repository:

https://github.com/georchestra/georchestra/blob/18.06/security-proxy/src/docker/Dockerfile#L5

No doing so can lead to the following issues:

* If no http endpoint manages an https redirect, then some links can be broken, ending up on a "connection refused" browser error page,
* even with a https redirect, some sensitive info can leak between the browser and the frontend, because of some unexpected redirections wired on http
* Above is an example of relative redirects on the SP, but some other parts of geOrchestra can be affected (geoserver UI links). As a result, I think the module should be activated on all jetty-based docker images.

# affected geOrchestra versions

Every versions since the docker introduction.

# steps to reproduce

See curl above

# Expected behaviour

No unexpected redirects.

# Additional context

Encountered on DTFTTH, reproduced on GGE which also is docker/jetty based.

fvanderbiest:
Backported to 18.06 with b2b0a9c9bd4148ef06bf736d8814b6994fe0bc89

PR #2169, opened 2018-09-19
http-forwarded jetty module added

fvanderbiest:
implements #2165

To be backported to all supported versions later


fvanderbiest:
Let's not forget geonetwork ...

Issue #2125, opened 2018-07-12
console - user sees details of his org


fvanderbiest:
@landryb said elsewhere: "Would there be a way for a user to see the details of his org (ie coverage, member list, etc...) ? right now he only sees the org name.

fvanderbiest:
Would that be OK, to show a "more information about this Org" link next to the Org name displayed in the `/console/account/userdetails` page ?
This new console page would be read-only, and access would be restricted to users of this Org.

Can you iterate on the above "(ie coverage, member list, etc...)". Thanks.

fvanderbiest:
Oops, I did not see there's already #2101

fvanderbiest:
> Would that be OK, to show a "more information about this Org" link next to the Org name displayed

Or a 2-column display: 
 * left column as currently with user-centric fields
 * right column with org-related informations + map

landryb:
Isn't it simpler to have org name a link that reuses the actual 'org details' page, but making the fields displays as readonly instead of entries/textfields ?

Otherwise yes it's less clicks for the user to see his personal info *and* his org details on the same page. It's just that if you add more fields to an org page, you need to remember to add them too on the user page...

fvanderbiest:
Okay, will discuss this with @tonio during implementation

bchartier:
> Org-related informations must include readonly list of org members

Not sure that all org members would appreciate to see that their names appear in the description of the organisation (see General Data Protection Regulation).

landryb:
afaik, console is not available to anonymous users...

fvanderbiest:
> Would it be only visible to authenticated users?

Yes

fvanderbiest:
Just to make it clear, we're speaking of the /console/account/userdetails page

bchartier:
> Just to make it clear, we're speaking of the /console/account/userdetails page

Thank you. I was affraid that this kind of information could be visible in the user account creation form.

fvanderbiest:
Implemented with #2172 

Smallish issue found & reported with #2229 
Possible improvement with #2230

PR #2172, opened 2018-09-20
show org info in `userdetails`

tonio:
ping @cmangeat


tonio:
see #2125

tonio:
@cmangeat on last thing, there a [3 strings](https://github.com/georchestra/georchestra/pull/2172/commits/5b9651734e2a424dd1e47079b9e006b6c8698707) that still needs to be translated. I put them into arbitrary tags, could you please plug with the translation system?

console/src/main/webapp/WEB-INF/i18n/application_fr.properties
```diff
@@ -24,6 +24,9 @@ editUserDetailsForm.fieldset.userDetails=Détails du compte
 editUserDetailsForm.fieldset.credentials=Identification
 editUserDetailsForm.changePassword.link=Changer le mot de passe
 editUserDetailsForm.success=Compte actualisé
+editUserDetailsForm.organisation=Organisme
+editUserDetailsForm.areaOfCompetence=Aire de competence
```
fvanderbiest: compétence

landryb: 'Emprise de compétence' ? 'Aire' feels weird..

console/src/main/webapp/WEB-INF/i18n/application.properties
```diff
@@ -24,6 +24,9 @@ editUserDetailsForm.fieldset.userDetails=User details
 editUserDetailsForm.fieldset.credentials=Credentials
 editUserDetailsForm.changePassword.link=Change Password
 editUserDetailsForm.success=User details updated successfully
+editUserDetailsForm.organisation=Organisation
```
fvanderbiest: Organization

tonio: Both are corrects. We use an «s» in the console.

landryb:
Tested the tip of this branch, the map looks good ! Although it's still clickable, so users might think they can change their global area (which is of course forbidden) by selecting/unselecting areas, and then hitting 'send'. I think the map should really be readonly :)

I'm wondering if we should put a plaintext (paginated?) readonly list of org area similar to the one where you edit the area, for users to check their org has the correct list, or have an 'export area code list as csv' button reusing what's done in #2159 ?

Yes, it's a usecase we have...and when there's 200+ entries its painful to check by looking at a map... 

And still +1 for 'Emprise de compétence' which is a term i think everyone understands. At least that's the one we've been using here for years...

Visually, i really like the two pane approach. I'm wondering if we should move the 'Organisme' field from the left pane to the right pane as a single value line above 'Emprise de compétence', titling the right pane 'Organisme de rattachement' (so that the pane title has a static size, like 'Détails du compte' for the left pane)

This way, less text duplication.. right now on the page i have 'organisme' label & value twice.

cmangeat:
A commercial could have said "territoire" for" zone de compétence", but maybe connotation are weird for a geographer...

cmangeat:
I couldn't manage to reproduce /  hit the js assertion. Tried with an user with no org, tried with an user with an org, modifying or not the firstname.

About the post, there are two phases, first processing the input, registering the update, then providing an output, which reflect the changes. 9a4a080ff2b40a7fa5923f3aadb22f379cba5b4d deals with the second phase, i.e. the "refreshed" form has also to present the org.

Issue #636, opened 2014-04-11
styler - missing SLD_VERSION


fphg:
According to the SLD 1.1.0 spec, a DescribeLayer request shall use the mandatory sld_version parameter (sld_version=1.1.0). See p.11 table 1.

landryb:
Fwiw - this hardcoded "/mapfishapp/" in the url, which is bad :)

fvanderbiest:
> Fwiw - this hardcoded "/mapfishapp/" in the url, which is bad :)

yes, we'll have to deal with this at some time ... see https://github.com/georchestra/georchestra/issues/227

PR #2175, opened 2018-09-21
Unhardcode /mapfishapp/ to enforce SLD 1.0.0 when the style comes fro…

landryb:
…m mapfishapp (cf #636)

match on ws/sld/geodoc instead, allows styler to work when
mapfishapp url is different than /mapfishapp/.


fvanderbiest:
Thanks for the fix Landry !

PR #2178, opened 2018-09-25
Remove a hardcoded /geoserver/wms and reuse GEOSERVER_WMS_URL

landryb:
Saw this one when poking at something else there, but i havent been able to trigger the codepaths that lead to this code being called. Also, it's a bit awkward to reuse GEOSERVER_WMS_URL which is supposed to be used only if OSM_AS_OVMAP is false, but oh well..


landryb:
To trigger it, one must load a layer into mfapp via POST

fvanderbiest:
That's it.
It would match by default `/geoserver/[ows|wms]` and you could edit it to also match  `/wxs/[ows|wms]`

PR #2180, opened 2018-09-26
Move console table creations to the right SQL file

landryb:
console-data should only insert sample data..

Unless hibernate in the console backend takes care of creating those ? then why `user_token` is different, and created in `04-console.sql` ?


fvanderbiest:
> Unless hibernate in the console backend takes care of creating those ?

That's the case indeed.

> then why `user_token` is different, and created in `04-console.sql` ?

Because the ORM does not handle this table at the moment.
Would be great though !

fvanderbiest:
Closing if you agree.

Issue #2176, opened 2018-09-25
[sec-proxy] problem with host header when target-mapping points to external URL


landryb:
With the following config in `targets-mapping.properties`
```
header=http://external.fqdn/banner/
```
the security proxy sends the request to `external.fqdn` webserver, but sets the `Host` header to the one received by the SP, because `localProxy` is true in https://github.com/georchestra/georchestra/blob/master/security-proxy/src/main/java/org/georchestra/security/Proxy.java#L667.
```
2018-09-25 09:39:15  security [DEBUG] Host header set to: localhost:8180 for proxy request.
2018-09-25 09:39:15 MainClientExec [DEBUG] Opening connection {}->http://tests8.dev.craig.fr:80
2018-09-25 09:39:15 DefaultHttpClientConnectionOperator [DEBUG] Connecting to tests8.dev.craig.fr/172.20.97.20:80
2018-09-25 09:39:15 DefaultHttpClientConnectionOperator [DEBUG] Connection established 10.0.0.8:45880<->172.20.97.20:80
2018-09-25 09:39:15 DefaultManagedHttpClientConnection [DEBUG] http-outgoing-0: set socket timeout to 1200000
2018-09-25 09:39:15 MainClientExec [DEBUG] Executing request GET /fr/banniere/ HTTP/1.1
2018-09-25 09:39:15 MainClientExec [DEBUG] Target auth state: UNCHALLENGED
2018-09-25 09:39:15 MainClientExec [DEBUG] Proxy auth state: UNCHALLENGED
2018-09-25 09:39:15 headers [DEBUG] http-outgoing-0 >> GET /fr/banniere/ HTTP/1.1
2018-09-25 09:39:15 headers [DEBUG] http-outgoing-0 >> user-agent: curl/7.52.1
2018-09-25 09:39:15 headers [DEBUG] http-outgoing-0 >> accept: */*
2018-09-25 09:39:15 headers [DEBUG] http-outgoing-0 >> x-forwarded-for: ::1
2018-09-25 09:39:15 headers [DEBUG] http-outgoing-0 >> x-forwarded-host: georchestra.dev.craig.fr
2018-09-25 09:39:15 headers [DEBUG] http-outgoing-0 >> x-forwarded-server: georchestra.dev.craig.fr
2018-09-25 09:39:15 headers [DEBUG] http-outgoing-0 >> connection: Keep-Alive
[...]

PR #2188, opened 2018-10-01
Remove 'Host' header override when localProxy is true (#2176)

landryb:
Was apparently a hack for geoserver 7 years ago, and causes wrong
behaviour when one of the SP targets points to an external FQDN

Might cause side effects, but works for me here, and unbreaks critical features (having the header served by a different FQDN, but through the SP)


pmauduit:
I have no objection to merge in master, but this would require some runtime tests (on sdi-stage.g.o ?)

Issue #2195, opened 2018-10-05
console - passwordRecovery fails when db was shut down then restarted


fvanderbiest:
/console/account/passwordRecovery?email=me%40provider.com then reports:

![screenshot from 2018-10-05 15-38-06](https://user-images.githubusercontent.com/265319/46538610-bb7d0000-c8b4-11e8-86e8-1e513580901b.png)
![screenshot from 2018-10-05 15-38-33](https://user-images.githubusercontent.com/265319/46538615-bfa91d80-c8b4-11e8-9099-4078530f808a.png)

fvanderbiest:
Restarting the console fixes the issue, but the expected behavior would be that the db connection pool is smart enough to recreate failing connections (use c3p0 ?)

fvanderbiest:
To be fixed on 16.12

cmangeat:
atlas, console-tpl and extractor app also use com.mchange.v2.c3p0.ComboPooledDataSource.

fvanderbiest:
...which should do the job IIUC ?

groldan:
So I couldn't reproduce this as-is, but this is what I found:
- it looks like the issue is reported against 16.12 but I couldn't run 16.12 (the docker build fails on that branch and there seems not to be published docker tags other than latest)
- On master, it's unlikely that a `GET` call to `/console/account/passwordRecovery?` (`PasswordRecoveryFormController.setupForm()`) would throw any JDBC related exception as it doesn't hit it. `POST` (`PasswordRecoveryFormController.generateToken()`) does a couple of db interactions (by means of `UserTokenDao.exists()/delete()/insertToken()`).
- More importantly, that stacktrace, unless there's a repeatable way to reproduce, is most likely unrelated to the request in question as it's *just a warning* issues by c3p0, most times asynchronously, indicating that the client code is still using a connection that the pool already discarded because it's thrown an exception before. In this case, at some point in a transactional JPA method, the operation failed becuase it was forcedly closed.
- Since there are commands (e.g. `AdvancedDelegationDao.findByRole()`) that don't close the connections, statements, and resultsets it uses, most probably the above stack trace happens when the c3p0 connection gets automatically closed.

That said, the only way I could reproduce something similar, which can very well happen on a live instance, was with some hacking to match the timing between a DAO command was assigned a connection and the time is gets executed, closing the connections in-between.
[...]

PR #2200, opened 2018-10-09
Console - db connection from pool tested before reuse

cmangeat:
… reuse it

relates to https://github.com/georchestra/georchestra/issues/2195


fvanderbiest:
If atlas, console and extractorapp also use `c3p0.ComboPooledDataSource` I would expect this PR to also modify the according configurations. Many thanks :-)

PR #2205, opened 2018-10-15
use c3p0 automaticTestTable to qualify connection from pool before use

cmangeat:


pmauduit:
I am just wondering where the table (in which schema of the db) will be created ... Hoping that it will end in the expected schema, to avoid having each webapp overlapping the same table.

PR #2208, opened 2018-10-16
GeoServer 2.12.5

fvanderbiest:
Branch `2.12.5-georchestra` created in https://github.com/georchestra/geoserver/tree/2.12.5-georchestra

To be tested...


pmauduit:
Nice catch for the classloader pollution with marlin ;-)

PR #2212, opened 2018-10-17
geoserver s3-geotiff profile

fvanderbiest:
More generally, how can we keep these profiles in sync more easily ?


pmauduit:
I cannot see any way of keeping the profiles in sync. The reference is the pom.xml where you found the profile, geoserver/webapp is another distinct pom project, there are no way of "integrating" a pom fragment into another pom

Issue #2209, opened 2018-10-16
Reduce docker image size


fvanderbiest:
Inspired from https://github.com/georchestra/geonetwork/pull/92

Issue #2171, opened 2018-09-20
docker - improve image generation


pmauduit:
To help customizing the official docker images, it would be interesting to provide the webapps in the docker container already unzipped in the `webapp/` directory. Looking at the current build process:

* https://github.com/georchestra/georchestra/blob/18.06/mapfishapp/src/docker/Dockerfile#L7
* https://github.com/georchestra/georchestra/blob/18.06/mapfishapp/pom.xml#L606-L617

I don't know if maven allows us to do this easily, but a directory in `target/` should contain the actual unzipped webapp. We could try to use it instead of using the war artifact.

fvanderbiest:
Duplicate of #2577

PR #2214, opened 2018-10-17
Reducing docker image size

fvanderbiest:
Proposal for #2209 and #2171 at the same time.

WIP - do not merge.
Right now, console `node_modules` are not excluded, which increases image size !


RemiDesgrange:
I tried to add a `.dockerignore` file but it doesn't work so far.

fvanderbiest:
Superseded by #2849

PR #2219, opened 2018-10-23
SP: change for ASYNC apache http client

cmangeat:
apache client does not implements "should" of https://tools.ietf.org/html/rfc2616#section-8.2.2
thus 414 error from backend server are igrnored as payload stream closing generate 500 error.

where as apache async http client does.

upgrade sp from java 1.7 to 1.8 in order to catch many qualified exception at the same time.

refactor: create http client builder at proxy bean initialisation.


fvanderbiest:
-1 on merging this to 18.06 as this may have unexpected side-effects.
+1 on merging it into master.

Thanks !

Issue #2217, opened 2018-10-19
console - protectedUser published by /console/private/roles


fvanderbiest:
Even when setting `protectedUser.uid1=geoserver_privileged_user` in `datadir/console/console.properties`, the `/console/private/roles` service exports the `geoserver_privileged_user` user:
```json
{
"description": "This group grants admin access to GeoServer",
"cn": "ADMINISTRATOR",
"users": [
"geoserver_privileged_user",
"testadmin"
],
"isFavorite": false
},
```
It should not be exported in the service.

PR #2222, opened 2018-10-26
console: filter protected user out of roles

cmangeat:


cmangeat:
relates to https://github.com/georchestra/georchestra/issues/2217

PR #2227, opened 2018-11-07
Send ldap attributes in cas replies

landryb:
Probably not to merge as-is - needs discussion. But generally i dislike putting all apps behind the SP and hacking them to auth users based on headers, when the said apps have support for CAS, so let's try to improve this. Most bits come from @pierrejego , thanks :)

With this PR i can connect a drupal8 (using CAS 3.0) and a nextcloud (using CAS 2.0) to my CAS, nextcloud needs the casServiceValidationSuccess.jsp chunk (diff is large but that's mostly reindentation, and upstream code for 3.0) so that attributes are appended to the reply as it's done in 3.0, without it the user is authenticated by nextcloud but i cant map cas attributes to nextcloud profile fields (email, displayname..).

For some reason groups don't work yet, `memberOf` seems empty (thus not added in the reply), i think this is because it's an OpenLDAP operational attribute, but @pierrejego told me he managed to make it work. Having LDAP groups sent this way would be a huge improvement.

Removing the `principalResolver` from `ldapPasswordHandler` is awkward but seems required per https://apereo.github.io/cas/4.2.x/installation/Configuring-Principal-Resolution.html#principalresolver-vs-authenticationhandler

Of course we could send more user attributes...

### DIFF
--- a/cas-server-webapp/src/main/webapp/WEB-INF/deployerConfigContext.xml
+++ b/cas-server-webapp/src/main/webapp/WEB-INF/deployerConfigContext.xml
@@ -61,7 +61,7 @@
                    | set its name to something other than its default name (typically the simple class name).
                    -->
                 <entry key-ref="proxyAuthenticationHandler" value-ref="principalResolver" />
-                <entry key-ref="ldapPasswordHandler" value-ref="principalResolver"/>
+                <entry key-ref="ldapPasswordHandler" value="#{null}"/>
             </map>
         </constructor-arg>
 
@@ -107,7 +107,15 @@
         <constructor-arg value="${ldap.authn.roleSearchFilter}" />
         <constructor-arg value="${ldap.authn.roleRoleAttribute}" />
         <constructor-arg value="${ldap.authn.pendingRoleName}" />
-
+        <property name="principalAttributeMap">
+          <map>
+            <entry key="mail" value="mail" />
+            <entry key="uid" value="uid" />
+            <entry key="memberOf" value="memberOf" />
+            <entry key="sn" value="sn" />
+            <entry key="givenName" value="givenName" />
+          </map>
+        </property>
     </bean>
 
     <!--
@@ -190,7 +198,17 @@
     <util:list id="registeredServicesList">
         <bean class="org.jasig.cas.services.RegexRegisteredService"
             p:id="0" p:name="HTTP and IMAP" p:description="Allows HTTP(S) and IMAP(S) protocols"
-            p:serviceId="^(https?|imaps?)://.*" p:evaluationOrder="10000001" />
+            p:serviceId="^(https?|imaps?)://.*" p:evaluationOrder="10000001">
+            <property name="allowedAttributes">
+              <list>
+                <value>uid</value>
+                <value>mail</value>
+                <value>memberOf</value>
+                <value>sn</value>
+                <value>givenName</value>
+              </list>
+            </property>
+        </bean>
         <!--
         Use the following definition instead of the above to further restrict access
         to services within your domain (including sub domains).
--- a/cas-server-webapp/src/main/webapp/WEB-INF/view/jsp/protocol/2.0/casServiceValidationSuccess.jsp
+++ b/cas-server-webapp/src/main/webapp/WEB-INF/view/jsp/protocol/2.0/casServiceValidationSuccess.jsp
@@ -19,20 +19,52 @@
 
 --%>
 <%@ page session="false" contentType="application/xml; charset=UTF-8" %>
+<%@ page import="java.util.Map.Entry" %>
 <%@ taglib prefix="c" uri="http://java.sun.com/jsp/jstl/core" %>
 <%@ taglib uri="http://java.sun.com/jsp/jstl/functions" prefix="fn" %>
 <cas:serviceResponse xmlns:cas='http://www.yale.edu/tp/cas'>
-	<cas:authenticationSuccess>
-		<cas:user>${fn:escapeXml(assertion.primaryAuthentication.principal.id)}</cas:user>
+    <cas:authenticationSuccess>
+        <cas:user>${fn:escapeXml(assertion.primaryAuthentication.principal.id)}</cas:user>
         <c:if test="${not empty pgtIou}">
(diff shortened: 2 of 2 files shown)


pierrejego:
Hello, I have just seen your message, I think when it worked, I had only on SV_USER Role in the overlay memberOf, but not sure at all. I did so many try to make it "not" work that I can not really be sure....

landryb:
You mean it worked only because you had only one value in memberOf ? that might explain why it doesnt work with multiple values...

landryb:
So i've found on https://gist.github.com/adamfranco/83dfbe784aab1162c478 a method that 'works' to flatten the group list via the `authenticator`:

```
<bean id="authenticator" class="org.ldaptive.auth.Authenticator"
      c:resolver-ref="pooledSearchDnResolver"
      c:handler-ref="pooledBindHandler" >
                <!-- Add the RecursiveEntryHandler to flatten the group hierarchy and-
                include ancestor groups in the MemberOf response values. -->
                <property name="entryResolver">
                        <bean class="org.ldaptive.auth.PooledSearchEntryResolver"
                                p:connectionFactory-ref="pooledLdapConnectionFactory">
                                <property name="searchEntryHandlers">
                                        <bean class="org.ldaptive.handler.RecursiveEntryHandler"
                                                p:searchAttribute="memberOf"
                                                p:mergeAttributes="memberOf"/>
                                </property>
                        </bean>
</property>
</bean>
```
but... it blows with my cas webapp configured to store tickets in memcached, because `kryo` fails to serialize those attributes when storing the ticket. if i temporarly disable memcached, the groups are added to the cas response, and nextcloud sees them:
```
=> CAS_Client::_readExtraAttributesCas20(DOMNodeList) [Client.php:3262]
  Found nested jasig style attributes [Client.php:3346]
  Attribute [uid] = breuil [Client.php:3352]
  Attribute [givenName] = Landry [Client.php:3352]
  Attribute [memberOf] = cn=GN_ADMIN,ou=roles,dc=craig,dc=fr [Client.php:3352]
  Attribute [memberOf] = cn=GN_REVIEWER,ou=roles,dc=craig,dc=fr [Client.php:3352]
```

landryb:
@fvanderbiest any comment/feedback on this feature ? i know CAS is a bit from the last decade, but still...

pmauduit:
>  i know CAS is a bit from the last decade, but still...

Well, it is used in production in several places actually: https://www.sfr.fr/cas/login?service=https%3A%2F%2Fwww.red-by-sfr.fr%2Faccueil%2Fj_spring_cas_security_check&theme=espaceclientred#redclicid=X_Menu_EspaceClient ;-)

cas-server-webapp/src/main/webapp/WEB-INF/deployerConfigContext.xml
```diff
@@ -61,7 +61,7 @@
                    | set its name to something other than its default name (typically the simple class name).
                    -->
                 <entry key-ref="proxyAuthenticationHandler" value-ref="principalResolver" />
-                <entry key-ref="ldapPasswordHandler" value-ref="principalResolver"/>
+                <entry key-ref="ldapPasswordHandler" value="#{null}"/>
```
pmauduit: Why removing the principalResolver here ?

landryb: Because iirc if i leave it, something doesnt work.. as i said in the first comment:

> Removing the principalResolver from ldapPasswordHandler is awkward but seems required per https://apereo.github.io/cas/4.2.x/installation/Configuring-Principal-Resolution.html#principalresolver-vs-authenticationhandler

pmauduit: ok, sorry for having missed the comment

landryb:
@pmauduit @fvanderbiest i'm still carrying this local modification fwiw.

fvanderbiest:
To be rebase onto upcoming CAS 6 (WIP in https://github.com/georchestra/georchestra-cas-server)

pmauduit:
> To be rebased onto upcoming CAS 6 (WIP in https://github.com/georchestra/georchestra-cas-server)

That will be complicated for several reasons:

* cas 6 is spring-boot based, and the configuration mechanisms are quite different from what we had since then
* cas 6 is also into its own seperated repository georchestra/georchestra-cas-server

I haven't had the opportunity to try the attribute release on cas6, but this should be pretty well documented and quite straightforward to configure: https://apereo.github.io/cas/6.1.x/integration/Attribute-Release.html

landryb:
Fwiw i've been able to achieve the same feature with cas6 from #2799, using only configuration in the datadir - no need to modify the CAS replies anymore.

* list the attributes that should be fetched from the ldap:
```
$grep givenN /etc/georchestra/cas/config/cas.properties 
cas.authn.ldap[0].principalAttributeList=sn,givenName,uid,mail,memberOf
```
* Using this cas service definition, specifying which attributes to send with which name:
```
$cat /etc/georchestra/cas/services/georchestra-1001.json
{
    "@class" : "org.apereo.cas.services.RegexRegisteredService",
    "serviceId" : "^https?://.*$",
    "name" : "georchestra",
    "id" : 1001,
    "attributeReleasePolicy" : {
        "@class" : "org.apereo.cas.services.ReturnMappedAttributeReleasePolicy",
        "allowedAttributes" : {
          "@class" : "java.util.TreeMap",
          "org": "file:/etc/georchestra/cas/filter-org.groovy",
          "cloud": "file:/etc/georchestra/cas/filter-cloud.groovy",
          "roles": "file:/etc/georchestra/cas/filter-roles.groovy",
          "uid": "uid",
          "mail": "mail",
          "sn": "sn",
          "givenName": "givenName"
        }
    },
    "evaluationOrder" : 10
}
```
in my usecase, the `org`, `roles` and `cloud` attributes are created by groovy scripts parsing/splitting the values of the `memberOf` ldap attribute. With that i've been able to auth users to drupal, and it should also work with nextcloud.

I still have to check what to do for cas to store tickets in memcache.

landryb:
yes, ppl can use what i posted above to achieve the same with only configurations - no code changes.

PR #2234, opened 2018-11-12
prepared statement and connection pool raw fix

cmangeat:
this to avoid "org.postgresql.util.PSQLException: This connection has been closed", from time to time, may need more.


fvanderbiest:
LGTM but asking for a second review by @pmauduit 
Thanks !

PR #2239, opened 2018-11-13
console - add the possibility to disable recaptcha at account creation

pmauduit:
By default the recaptcha integration is activated, but adding:

```
recaptcha.activated=false
```

Allows to deactivate the recaptcha integration.

Tests: runtime on previous flag set to false / true. With false, I was able to create my account without having to fill the recaptcha field.


landryb:
\o/ \o/ thanks @pmauduit - will make sure to test that shortly.

cmangeat:
(https://github.com/georchestra/georchestra/blob/master/console/src/main/webapp/WEB-INF/views/passwordRecoveryForm.jsp#L40)

landryb:
https://github.com/georchestra/georchestra/blob/master/console/src/main/webapp/WEB-INF/views/passwordRecoveryForm.jsp#L74 too :)

pmauduit:
hmmm seems that it is still in a WIP state ... sorry for the points I missed out.

pmauduit:
Ok reworked using the following strategy:

1. grep 'recaptcha/api.js' in every jsp files, got 'createAccountForm' and 'passwordRecoveryForm'.
2. grep for the 2 previous view names into every Java source files, got 'EditUserDetailsFormController.java' and 'NewAccountFormController.java' from the first one, "NewPasswordFormController.java' and 'PasswordRecoveryFormController.java' from the second one.
3. Ensure that everything is disabled (if / when statements in java / jsp code)
4. Last pass git grepping onto `RecaptchaUtils.validate()`, ensuring that there were conditional statements around.

Untested at runtime after the 2nd commit, @landryb care to have a look ?

console/src/main/webapp/WEB-INF/views/passwordRecoveryForm.jsp
```diff
@@ -90,7 +94,12 @@
 	<script type="text/javascript">
     /* Validate the form */
     function validate() {
+<c:when test="${recaptchaActivated}">
```
landryb: needs to be wrapped within `<c:choose>` & </c:choose>`

landryb:
Password recovery page blows:
```
** Root cause is: Illegal use of <when>-style tag without <choose> as its direct parent
javax.servlet.jsp.JspTagException: Illegal use of <when>-style tag without <choose> as its direct parent
```

then when/otherwise bits just needs to be wrapped within `<c:choose>` tags. With that fixed, account creation & password recovery pages work fine with recaptcha disabled.

console/src/main/webapp/WEB-INF/views/passwordRecoveryForm.jsp
pmauduit: Wooops sorry ...

fvanderbiest:
LGTM but missing var in https://github.com/georchestra/datadir/blob/master/console/console.properties

pmauduit:
@fvanderbiest  https://github.com/georchestra/datadir/pull/94

Good to go ?

fvanderbiest:
LGTM, you can merge, unless you expect someone else's review.

PR #2240, opened 2018-11-13
so to avoid filename to be encoded twice a different way (once ' ' to '%…

cmangeat:
…20', once no encoding 'type name')


fvanderbiest:
This is to fix an issue where shapefiles with a space in their filename cannot be read in zip file, thus not converted to GeoJSON in mapfishapp.

PR #2259, opened 2018-11-18
Fix 1885

severo:
Note that it's based on #2248 branch, in order to compile the mapfishapp module. If this is a problem, I remove the dependency.

### DIFF
--- a/mapfishapp/src/main/webapp/app/addons/goto/manifest.json
+++ b/mapfishapp/src/main/webapp/app/addons/goto/manifest.json
@@ -7,7 +7,8 @@
         "projections": [{
             "srs": "EPSG:4326", 
             "name": "WGS 84",
-            "labels": ["Longitude", "Latitude"]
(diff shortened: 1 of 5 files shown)


mapfishapp/src/main/webapp/app/addons/goto/manifest.json
```diff
@@ -7,7 +7,8 @@
         "projections": [{
             "srs": "EPSG:4326", 
             "name": "WGS 84",
-            "labels": ["Longitude", "Latitude"]
+            "labels": ["Longitude", "Latitude"],
+            "decimalPrecision": 10
```
fvanderbiest: 10 does not look like a relevant default value here.
6 would better suit the general use case IMO.

config.json should also be updated.

severo:
I close the PR, see https://github.com/georchestra/georchestra/pull/2281 instead.

Issue #2252, opened 2018-11-17
Build fails for geowebcache-webapp


severo:
**Describe the bug**

Following the steps in https://github.com/georchestra/georchestra/blob/master/docs/build.md (only avoiding `-Dserver=mytemplate` due to #2249), the build fails at project `geowebcache-webapp` with the following error message:

```
[ERROR] Failed to execute goal org.apache.maven.plugins:maven-war-plugin:2.2:war (default-war) on project geowebcache-webapp: Execution default-war of goal org.apache.maven.plugins:maven-war-plugin:2.2:war failed: basedir /mnt/ssd/fastdata/slesage/dev/4_contratos/c2c/georchestra/geowebcache-webapp/target/conf/geowebcache-webapp does not exist -> [Help 1]
org.apache.maven.lifecycle.LifecycleExecutionException: Failed to execute goal org.apache.maven.plugins:maven-war-plugin:2.2:war (default-war) on project geowebcache-webapp: Execution default-war of goal org.apache.maven.plugins:maven-war-plugin:2.2:war failed: basedir /mnt/ssd/fastdata/slesage/dev/4_contratos/c2c/georchestra/geowebcache-webapp/target/conf/geowebcache-webapp does not exist
    at org.apache.maven.lifecycle.internal.MojoExecutor.execute (MojoExecutor.java:215)
    at org.apache.maven.lifecycle.internal.MojoExecutor.execute (MojoExecutor.java:156)
    at org.apache.maven.lifecycle.internal.MojoExecutor.execute (MojoExecutor.java:148)
    at org.apache.maven.lifecycle.internal.LifecycleModuleBuilder.buildProject (LifecycleModuleBuilder.java:117)
    at org.apache.maven.lifecycle.internal.LifecycleModuleBuilder.buildProject (LifecycleModuleBuilder.java:81)
    at org.apache.maven.lifecycle.internal.builder.singlethreaded.SingleThreadedBuilder.build (SingleThreadedBuilder.java:56)
    at org.apache.maven.lifecycle.internal.LifecycleStarter.execute (LifecycleStarter.java:128)
    at org.apache.maven.DefaultMaven.doExecute (DefaultMaven.java:305)
    at org.apache.maven.DefaultMaven.doExecute (DefaultMaven.java:192)
    at org.apache.maven.DefaultMaven.execute (DefaultMaven.java:105)
    at org.apache.maven.cli.MavenCli.execute (MavenCli.java:956)
[...]

PR #2283, opened 2018-11-27
fixes #2028 (georchestra.datadir default value)

### LINKED ISSUE #2028: georchestra.datadir default value
What about having `georchestra.datadir` set by default to `/etc/georchestra` in our poms ?
```xml
<properties>
    <georchestra.datadir>/etc/georchestra</georchestra.datadir>
    ...
```
This would ease jetty running apps with the default usecase.

severo:
I created a commit per module, in order to ease future fixes if there is a need to.

## Modules where we modified pom.xml

Some modules directly call `georchestra.datadir` system property in the java code:

- [x] analytics
- [x] atlas
- [x] cas-server-webapp
- [x] console
- [x] security-proxy
- [x] geonetwork - PR proposed in the repository (https://github.com/georchestra/geonetwork/pull/99). We must merge that PR first in order to update the submodule commit hash.

Others depend on the commons module, passing it the `georchestra.datadir` system property (note that the previous list: analytics, atlas, console, security-proxy and cas-server-webapp, also depend on the commons module):

- [x] header
- [x] extractorapp
- [x] mapfishapp (note: we also set the `mapfish-print-config` property)
- [x] geowebcache-webapp (<strike>note: I could not test with jetty:run, due to https://github.com/georchestra/georchestra/issues/2252</strike> - fixed with 79adb5fa6edefda380a16f495462ea97aa35a51a)

## Modules left unmodified

Modules that use the `georchestra.datadir` system property, but that do not include the jetty-maven-plugin in the pom.xml are left as is, since the default system property is intended only for tests purposes using `mvn jetty:run`. That includes the following modules:

- [x] commons
- [x] geoserver

Other modules do not use the `georchestra.datadir` system property, so they do not need it to be set:

- [x] config
- [x] epsg-extension
- [x] geotools
- [x] ogc-server-statistics
- [x] root directory

Other do not have a `pom.xml`, so there is nothing to do:

- [x] docker
- [x] docs
- [x] ldap
- [x] migrations
- [x] postgresql
- [x] tests

### DIFF
--- a/analytics/pom.xml
+++ b/analytics/pom.xml
@@ -208,6 +208,12 @@
           <httpConnector>
             <port>8280</port>
           </httpConnector>
+          <systemProperties>
+            <systemProperty>
+              <name>georchestra.datadir</name>
+              <value>/etc/georchestra</value>
+          </systemProperty>
+        </systemProperties>
         </configuration>
       </plugin>
     </plugins>
--- a/atlas/pom.xml
+++ b/atlas/pom.xml
@@ -329,6 +329,12 @@
                     <httpConnector>
                         <port>8291</port>
                     </httpConnector>
+                    <systemProperties>
+                        <systemProperty>
+                            <name>georchestra.datadir</name>
+                            <value>/etc/georchestra</value>
+                        </systemProperty>
+                    </systemProperties>
                 </configuration>
             </plugin>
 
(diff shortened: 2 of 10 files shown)


severo:
Please review https://github.com/georchestra/geonetwork/pull/99 first, in order to update the submodule reference.

fvanderbiest:
Thanks.
Please update GN submodule, now that upstream PR is merged.

severo:
> Thanks.
> Please update GN submodule, now that upstream PR is merged.

OK. See https://github.com/georchestra/georchestra/pull/2344

Issue #2123, opened 2018-07-11
fix properties loading order/conflicts


landryb:
Generalisation of #2114, which applies potentially to the following files:
```
$git grep -l property-placeholder
analytics/src/main/webapp/WEB-INF/ws-servlet.xml
atlas/src/main/resources/camel/atlas_app.xml
cas-server-webapp/src/main/webapp/WEB-INF/spring-configuration/propertyFileConfigurer.xml
console/src/main/resources/META-INF/spring/applicationContext.xml
console/src/main/webapp/WEB-INF/spring/webmvc-config.xml
geowebcache-webapp/src/main/webapp/WEB-INF/geowebcache-servlet.xml
security-proxy/src/main/filtered-resources/WEB-INF/proxy-servlet.xml
security-proxy/src/main/webapp/WEB-INF/applicationContext-security.xml
security-proxy/src/test/resources/applicationContext-test.xml
```
in geonetwork submodule (some might be from/apply to georchestra, some not..):
```
core/src/main/resources/config-spring-geonetwork.xml
messaging/src/main/resources/config-spring-geonetwork-parent.xml
web/src/main/webResources/WEB-INF/config-db/defaultJdbcDataSource.xml
web/src/main/webResources/WEB-INF/config-db/postgis-alternate-datasource.xml
web/src/main/webResources/WEB-INF/config-db/postgres-postgis.xml
web/src/main/webResources/WEB-INF/config-spring-geonetwork.xml
web/src/main/webapp/WEB-INF/config-security/config-security.xml
workers/wfsfeature-harvester/src/main/resources/config-spring-geonetwork.xml
```
(none found in geoserver submodule)

When there are several `property-placeholder` entries in a file, the values they read might overwrite each other. On top of this, we want the following priority:
* `default.properties` (which has the lower weight/is overriden by values set in other files)
* `$webapp/WEB-INF/foo.properties` (which will disappear with the removal of config.jar in #1417)
* `${georchestra.datadir}/foo/foo.properties` which has the highest priority

landryb:
And the occurences of `ctx:property-placeholder` found in `${georchestra.datadir}/geonetwork/config`

landryb:
I think #2114 is not correctly fixed for some properties, because with 18.06 and the following config:
```
default.properties:publicUrl=https://ids.dev.craig.fr
security-proxy/security-proxy.properties:public.host=${publicUrl}
```
[...]

PR #2292, opened 2018-11-28
[WIP] Fixes 2123 (fix properties loading order/conflicts)

severo:
See https://github.com/georchestra/georchestra/issues/2123.

I created a commit per module, in order to ease future fixes if there is a need to.

Done:

- [x] atlas

Modules that do not take `default.properties` into account (see https://github.com/georchestra/georchestra/issues/2123#issuecomment-442622628):

- [x] analytics

Waiting:

- [ ] cas-server-webapp - it's complicated, because the [cas.properties](https://github.com/georchestra/georchestra/blob/master/cas-server-webapp/src/main/filtered-resources/WEB-INF/cas.properties) file is filtered using [maven.filter](https://github.com/georchestra/georchestra/blob/master/config/defaults/cas-server-webapp/maven.filter) from the config module. It's better to wait for #1417 to be solved first.
  One detail: it think the name of the directory "filtered-resources" is somewhat misleading, something like "resources-to-be-filtered" explains better what happens with these files.

To do:

- [ ] console - note that I don't touch the [applicationContext.xml](https://github.com/georchestra/georchestra/blob/master/console/src/main/resources/META-INF/spring/applicationContext.xml) file, see #2294
- [ ] security-proxy
- [ ] geonetwork
- [ ] header
- [ ] extractorapp
- [ ] mapfishapp
- [ ] geowebcache-webapp
- [ ] commons
- [ ] geoserver
- [ ] config
- [ ] epsg-extension
- [ ] geotools
- [ ] ogc-server-statistics
- [ ] root directory
- [ ] docker
- [ ] docs
- [ ] ldap
- [ ] migrations
- [ ] postgresql
- [ ] tests

### DIFF
--- a/analytics/src/main/webapp/WEB-INF/ws-servlet.xml
+++ b/analytics/src/main/webapp/WEB-INF/ws-servlet.xml
@@ -69,12 +69,9 @@
 
     <!-- Substitutes any ${...} variables in this (and loaded) spring configuration file
 				with values from the properties file -->
-    <context:property-placeholder location="/WEB-INF/analytics.properties"
-        ignore-resource-not-found="true" ignore-unresolvable="true" order="2" />
-
-    <context:property-placeholder
-        location="file:${georchestra.datadir}/analytics/analytics.properties"
-        ignore-resource-not-found="true" ignore-unresolvable="true" order="1" />
+    <context:property-placeholder location="/WEB-INF/analytics.properties,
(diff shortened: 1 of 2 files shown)


severo:
I will do one PR per module, as it seems to be the way to do, seeing https://github.com/georchestra/georchestra/issues/1417  for example.

PR #2298, opened 2018-11-29
atlas - fix properties loading order

severo:
See https://github.com/georchestra/georchestra/issues/2123.

Tests done (with `mvn jetty:run`, and playing with the `psql.url`
property):
- fails when none of the 3 files provides the property
- works when only one file provides the property
- follows the priority order:
  - least priority:   `${georchestra.datadir}/default.properties`
  - middle priority:  `/src/main/resources/atlas.properties`
  - highest priority: `${georchestra.datadir}/atlas/atlas.properties`

### DIFF
--- a/atlas/src/main/resources/camel/atlas_app.xml
+++ b/atlas/src/main/resources/camel/atlas_app.xml
@@ -21,15 +21,10 @@
     <jpa:repositories base-package="org.georchestra.atlas.repository" />
 
     <context:property-placeholder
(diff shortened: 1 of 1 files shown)


fvanderbiest:
> I set the priority according to [#2123 (comment)](https://github.com/georchestra/georchestra/issues/2123#issue-340266138), not to [#2123 (comment)](https://github.com/georchestra/georchestra/issues/2123#issuecomment-442931212). We must decide which order is the desired one.

I think it's OK. Thanks for mentionning it.

PR #2300, opened 2018-11-29
Fix postgis install & optimize docker image size

jeanpommier:
postgis seems to need the postgresql-10-postgis-2.4-scripts package to enable the create extension command (database init fails otherwise)
Bonus: the COPY --chown  reduces the size of the image


severo:
Maybe `--no-install-recommends` is what reduces the size of the image...

jeanpommier:
Yes, you're right. The chown too, but probably not much in that case.
Hum, I'm confused, I have mixed things up. Big, complicated day...
Anyway, at least, it reduces the image's size...

fvanderbiest:
Thanks !

PR #2309, opened 2018-12-04
Config readme format

severo:
I know that the directory will disappear, but meanwhile, a small fix that changes everything to understand the README, since a lot of the text was invisible due to format issues.

And I now realize that a lot of users of geOrchestra surely didn't notice this failure, and had problem to configure their geOrchestra instance due to this problem.


fvanderbiest:
Indeed, will soon be removed...
Anyway, thanks :-)

PR #2311, opened 2018-12-04
Remove unused file

severo:
I think that https://github.com/georchestra/georchestra/pull/2236 missed this change. All is working correctly after the deletion (with `mvn -Dgeorchestra.datadir=/etc/georchestra/ jetty:run`).


fvanderbiest:
Good point, thanks :-)

Issue #2210, opened 2018-10-16
CAS - Relicates of config.jar in the JSP files 


pmauduit:
**Bug description**

We can find some traces of the shared config here:
https://github.com/georchestra/georchestra/blob/master/cas-server-webapp/src/main/webapp/WEB-INF/view/jsp/default/ui/includes/top.jsp#L38-L42

**geOrchestra version or branch**

since 18.06, also present on master currently

**Expected behavior**

We should get rid of these shared.* variables previously defined in the old config.jar process. Maybe the concerned variables should end in the default.properties.

fvanderbiest:
linked with #1417

fvanderbiest:
Yes, we should fix all this on master branch.

severo:
Only four PR (#2366, #2367, #2368, #2369) to merge, and we can close the issue!

PR #2369, opened 2019-01-10
Remove obsolete properties from doc files

severo:
See #2210.

### DIFF
--- a/docs/setup/postgresql.md
+++ b/docs/setup/postgresql.md
@@ -35,7 +35,7 @@ psql -d georchestra -f postgresql/080-geofence.sql
 
 ## OGC statistics schema
 
-If the **security proxy** is deployed and ```OGC_STATISTICS``` is true in your setup ([true by default](https://github.com/georchestra/datadir/blob/18.06/analytics/js/GEOR_custom.js#L4)):
+If the **security proxy**' is deployed and its log4j uses the `org.georchestra.ogcservstatistics.log4j.OGCServicesAppender` as configured in the [datadir](https://github.com/georchestra/datadir/blob/docker-master/security-proxy/log4j/log4j.properties):
 ```
 psql -d georchestra -f postgresql/050-ogc-server-statistics.sql
 ```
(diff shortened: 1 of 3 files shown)


docs/setup/postgresql.md
```diff
@@ -35,7 +35,7 @@ psql -d georchestra -f postgresql/080-geofence.sql
 
 ## OGC statistics schema
 
-If the **security proxy** is deployed and ```shared.ogc.statistics.activated``` is true in your setup (false by default):
+If the **security proxy** is deployed and ```OGC_STATISTICS``` is true in your setup ([true by default](https://github.com/georchestra/datadir/blob/18.06/analytics/js/GEOR_custom.js#L4)):
```
fvanderbiest: rather: If SP's log4j uses the org.georchestra.ogcservstatistics.log4j.OGCServicesAppender as configured in https://github.com/georchestra/datadir/blob/docker-master/security-proxy/log4j/log4j.properties

This is more logical than refering to a client side property here (which is just a consequence).

ldap/README.md
```diff
@@ -34,7 +34,7 @@ The basic users:
  * ```testeditor``` has the USER & GN_EDITOR roles. The password is **testeditor**.
  * ```testadmin``` has the USER, GN_ADMIN, ADMINISTRATOR and MOD_* roles. The password is **testadmin**
  * ```testdelegatedadmin``` has the USER role. Is able to grant the EXTRACTORAPP & GN_EDITOR roles to members of the psc & c2c orgs. The password is **testdelegatedadmin**
- * ```geoserver_privileged_user``` is a required user. It is internally used by the extractorapp, mapfishapp & geofence modules. The default password is ```gerlsSnFd6SmM``` (you should change it, and update the ```shared.privileged.geoserver.pass``` option in your shared.maven.filters file).
+ * ```geoserver_privileged_user``` is a required user. It is internally used by the extractorapp, mapfishapp & geofence modules. The default password is ```gerlsSnFd6SmM``` (you should change it, and update the datadir as explain in its [README](https://github.com/georchestra/datadir/blob/18.06/README.md)).
```
fvanderbiest: as explained

docs/setup/postgresql.md
severo: Fixed, thanks

ldap/README.md
severo: Fixed, thanks

PR #2378, opened 2019-01-15
"en" default language everywhere

severo:
(even if the default results of GEONAMES are for France)


severo:
See #2378

severo:
See https://github.com/georchestra/datadir/pull/106

PR #2389, opened 2019-01-16
geonetwork - add a translation string to geonetwork

severo:


severo:
It has been reviewed in https://github.com/georchestra/geonetwork/pull/105, so I merge.

Issue #105, opened 2013-05-23
GN - Mapfishapp link for services


fvanderbiest:
From http://ns313112.ovh.net/geonetwork/apps/georchestra/?uuid=9e231a5c6a7b3b4a784b9a63899819bfc506adaa

Action : "view" => POSTed to mapfishapp:

```
{"services":[],"layers":[{"layername":"","metadataURL":"http://ns313112.ovh.net/geonetwork?id=113","owstype":"WMS","owsurl":"http://services.sandre.eaufrance.fr/geo/ouvrage?","title":"http://services.sandre.eaufrance.fr/geo/ouvrage? http://services.sandre.eaufrance.fr/geo/ouvrage?"}]}
```

Should be:

```
{"services":[{
    "owstype": "WMS",
    "owsurl": "http://services.sandre.eaufrance.fr/geo/ouvrage?"
}]}
```

See https://github.com/georchestra/georchestra/blob/master/mapfishapp/README.md#parameters

PR #2390, opened 2019-01-16
geonetwork - Add a translation string

severo:


severo:
It has been reviewed in #105, so I merge.

PR #2394, opened 2019-01-21
Homogenize maven-compiler-plugin settings

groldan:
maven-compiler-plugin was defined in several pom
files with different plugin version and Java version.

This patch defines the plugin version in the root pom's
pluginManagement section and its configuration in the
root pom's plugins section for all modules.


groldan:
would you merge it? cause I don't have enough privileges. TIA.

Issue #2305, opened 2018-12-04
PUT isfavorite not taken into account


fvanderbiest:
Branch 18.06

PUT https://fqdn.fr/console/private/roles/EL_TOTO 
```json
{"description":null,"cn":"EL_TOTO","users":[],"isFavorite":false,"originalID":"EL_TOTO"}
```

GET https://fqdn.fr/console/private/roles/EL_TOTO 
```json
{"description":null,"cn":"EL_TOTO","users":[],"isFavorite":true}
```

groldan:
Fixed on 18.06 by means of #2404
Awaiting merge of #2403 for a fix on master including integration testing

groldan:
Closing, fixed by #2403

PR #2404, opened 2019-01-23
Fix for #2305 on 18.06

groldan:
Fix a bug in console role management by which `isFavorite:false` won't be 

Setting a role as favorite works, but setting it back to false
doesn't. This patch fixes the issue. Note though that proper
testing for the fix requires some integration testing infrastructure
which is being worked out on the 18.12 branch as it would
be overkill to include on this one this close to a release.

See #2043


groldan:
> Isn't there an extra commit in this PR ?

Indeed, an intruder from #2392.

PR #2405, opened 2019-01-23
Harmonize spring and other important dependency versions

groldan:
Fixes 1329

Harmonize spring and other important dependency versions
across geoOrchestra modules (not on submodules i.e. geoserver
and geonetwork).

Depending on a released version of spring-boot-dependencies provides
a curated list of dependency versions for all modules.

See https://github.com/spring-projects/spring-boot/blob/v1.5.19.RELEASE/spring-boot-dependencies/pom.xml,
which is the last release of the 1.x series, defining dependencies for
Spring 4.3.22.RELEASE and related dependencies that are known to
work well with that spring version.

This is common practice when working on spring-boot projects.

Other dependency versions harmonized are:

- Spring-security: 4.2.11.RELEASE, by means of the parent pom.
- Postgresql: 9.4-1205-jdbc41

All additional dependencies (those not listed in the
spring-boot-dependeices pom) are added to the dependencyManagement
section in the root pom, and the individual module dependency
declarations have the version removed, so that a single version of each
dependency is ensured across all modules.

A special case is extractorapp, which overrides the gt.version property
declared in the root pom to stay at version 9.2, as it uses the gt-mif
module that seems to no longer exist.

Note: comments were removed from
/mapfishapp/ws/addons/annotation/config.json and
/mapfishapp/ws/addons/magnifier/config.json. The JSON
spec does not allow comments in JSON and if we let them
and strip them out would be hurting interoperability.


mapfishapp/src/test/java/org/georchestra/mapfishapp/ws/upload/UpLoadFileManagementGTImplTest.java
```diff
@@ -152,11 +154,11 @@ public void testKML22ExtendedData() throws Exception {
...
 	}
 
-    private String getJsonFieldValue(JSONObject properties, String field) {
+    private Object getJsonFieldValue(JSONObject properties, String field) {
 
-        String value;
+        Object value;
         try {
-            value = properties.getString(field);
+            value = properties.get(field);
```
cmangeat: there is an optString method with return null, or a string, not sure wether it is available or not in the targetted version, in 20180813 for sure (as in master).

mapfishapp/src/main/java/org/georchestra/mapfishapp/ws/classif/SLDClassifier.java
```diff
@@ -347,8 +347,8 @@ private FeatureTypeStyle createFeatureTypeStyle(I_FilterFactory filterFact, I_Sy
      * @throws DocServiceException 
      */
     @SuppressWarnings("unchecked")
-    private WFSContentDataStore connectToWFS(final URL wfsUrl) throws DocServiceException {   
-    	WFSContentDataStore wfs = null;
+    private WFSDataStore connectToWFS(final URL wfsUrl) throws DocServiceException {   
+    	WFSDataStore wfs = null;
```
cmangeat: May need side effects testing ?

mapfishapp/src/main/java/org/georchestra/mapfishapp/ws/AddonController.java
```diff
@@ -220,7 +220,8 @@ private JSONArray buildAddonSpecs(String path) {
                 continue;
             }
             try {
-                JSONArray parsed = new JSONArray(FileUtils.readFileToString(curConfig));
+                String contents = FileUtils.readFileToString(curConfig, "UTF-8");
+				JSONArray parsed = new JSONArray(contents);
```
cmangeat: This should make the linter unhappy (I agree, it should also have other way to copy an array in an array).

mapfishapp/src/test/java/org/georchestra/mapfishapp/ws/AddonControllerTest.java
```diff
@@ -20,6 +22,10 @@
 
 public class AddonControllerTest {
 
+	public static @BeforeClass void init(){
```
cmangeat: Where does theses tab chars come from ?  Don't know whether we have to use tab or whitespace.

mapfishapp/src/main/java/org/georchestra/mapfishapp/ws/classif/SLDClassifier.java
groldan: not sure I follow. From the old geotools version we were using to the 19.x series this patch upgrades to, the WFS-NG datastore graduated as the default one and `WFSContentDataStore` is no longer visible/does no longer exist. I've updated the code to use the public classes `WFSDataStoreFactory`/`WFSDataStore` which are the same ones they used to be but renamed, same parameters, etc.

mapfishapp/pom.xml
```diff
@@ -12,141 +12,96 @@
   <url>http://www.georchestra.org</url>
   <properties>
     <maven.test.skip>false</maven.test.skip>
-    <spring.version>3.1.0.RELEASE</spring.version>
+    <!-- Override parent version with 2.1.1-georchestra. Revisit. -->
```
fvanderbiest: Revisit when mapfishapp is completely rewritten, yes ;-D

fvanderbiest:
This is impressive !
Looks good to me.

Thanks Gabriel.

PR #2406, opened 2019-01-24
Update geoserver and last 18.06 commits

severo:


fvanderbiest:
When you merge 18.06 into master, no need to do a PR.
This can be silent.

severo:
> When you merge 18.06 into master, no need to do a PR.
> This can be silent.

OK

Issue #2391, opened 2019-01-18
maven - official wars are generated with a "-null" suffix


pmauduit:
After the config.jar removal, we probably need to adapt the way the official war packages are generated:
https://packages.georchestra.org/wars-master/

severo:
See https://github.com/georchestra/geonetwork/pull/108 and later https://github.com/georchestra/georchestra/pull/2416 that depend on it.

landryb:
As long as you dont break the versioning done for the debian pkgs in #1931, it was painful to get it right :)

severo:
> As long as you dont break the versioning done for the debian pkgs in #1931, it was painful to get it right :)

OK, it seems it was a long story. Let @pmauduit or @fvanderbiest give their opinion.

severo:
> what about adding the `<server>` property to the root pom instead of to each one

I didn't do it because the modules can be built by their own, from inside the directory. But we can change.

> and set the default value to `${project.version}`, so you get `atlas-18.12-SNAPTHOT` or `atlas-18.12` as appropriate for a snapshot or release version.

As said before, let listen to other opinions. To me, it would be OK.

PR #2416, opened 2019-01-26
Fix the suffix of the generated war files.

severo:
It was "-null", due to the removal of the config.jar that provided the
"server" property.

Fixes https://github.com/georchestra/georchestra/issues/2391.


severo:
See https://github.com/georchestra/geonetwork/pull/108 before merging.

severo:
Note that this way, the suffix may still be overridden as follows, for example:

```bash
$ mvn -Dserver=another-suffix install
```

fvanderbiest:
requires an update, thanks !

Issue #2068, opened 2018-06-01
factorize instanceName & language in default.properties


fvanderbiest:
https://github.com/georchestra/datadir/blob/17.12/security-proxy/security-proxy.properties#L49
https://github.com/georchestra/datadir/blob/17.12/mapfishapp/mapfishapp.properties#L2
https://github.com/georchestra/datadir/blob/17.12/ldapadmin/ldapadmin.properties#L4
https://github.com/georchestra/datadir/blob/17.12/geowebcache/geowebcache.properties#L4
https://github.com/georchestra/datadir/blob/17.12/cas/cas.properties#L3
https://github.com/georchestra/datadir/blob/17.12/analytics/analytics.properties#L6

on master branch ...

pmauduit:
Looks like recursive resolutions of properties file is not possible, currently the default docker composition is ending with the following trace at SP bootstrap:

```
proxy_1         | org.springframework.beans.factory.BeanCreationException: Error creating bean with name 'org.springframework.security.filterChains': Cannot resolve reference to bean 'org.springframework.security.web.DefaultSecurityFilterChain#0' while setting bean property 'sourceList' with key [0]; nested exception is org.springframework.beans.factory.BeanCreationException: Error creating bean with name 'org.springframework.security.web.DefaultSecurityFilterChain#0': Cannot create inner bean '(inner bean)#41d426b5' of type [org.springframework.security.web.authentication.logout.LogoutFilter] while setting constructor argument with key [3]; nested exception is org.springframework.beans.factory.BeanCreationException: Error creating bean with name '(inner bean)#41d426b5': Bean instantiation via constructor failed; nested exception is org.springframework.beans.BeanInstantiationException: Failed to instantiate [org.springframework.security.web.authentication.logout.LogoutFilter]: Constructor threw exception; nested exception is java.lang.IllegalArgumentException: ${publicUrl}/cas/logout?fromgeorchestra isn't a valid redirect URL
```

One workaround though, is to launch the SP container with `JAVA_OPTIONS` env variable set at least to `-DpublicUrl=http://blah/`.

fvanderbiest:
> One workaround though, is to launch the SP container with JAVA_OPTIONS env variable set at least to -DpublicUrl=http://blah/.

I like it a lot !
Though I would prefer a dedicated env var for the docker case. 

extractorapp, console, cas and atlas also use ${publicUrl}

pmauduit:
> I like it a lot !

[...]

PR #2430, opened 2019-01-28
viewer - simplify code

severo:
It will be the pattern to use for #2068, since it allows to access the properties files only once for all the JSP files, and reduces the required code.

### DIFF
--- a/mapfishapp/src/main/webapp/WEB-INF/jsp/index.jsp
+++ b/mapfishapp/src/main/webapp/WEB-INF/jsp/index.jsp
@@ -38,18 +38,6 @@
 Boolean anonymous = true;
 Boolean admin = false;
 
-String defaultLanguage = "en";
-
-try {
-  ApplicationContext ctx = RequestContextUtils.getWebApplicationContext(request);
-  if ((ctx.getBean(GeorchestraConfiguration.class) != null)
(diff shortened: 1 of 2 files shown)


severo:
Note that we could do the same for analytics, header, and other modules.

mapfishapp/src/main/webapp/WEB-INF/jsp/index.jsp
```diff
@@ -65,7 +47,7 @@ ResourceBundle bundle = org.georchestra.mapfishapp.ws.Utf8ResourceBundle.getBund
 String detectedLanguage = rLocale.getLanguage();
 String forcedLang = request.getParameter("lang");
 
-String lang = defaultLanguage;
+String lang = request.getParameter("defaultLanguage");
```
fvanderbiest: There's no such parameter to my knowledge

severo: It's not the name of a property from a properties file, it's a parameter passed in the Spring configuration file, see https://github.com/georchestra/georchestra/pull/2430/files#diff-c15e9e41a281b1d28fad1be82868b16eR60

fvanderbiest: okay !

PR #2434, opened 2019-01-29
Fix typos

severo:


severo:
Oups, it should have been on master

Issue #2226, opened 2018-10-31
Most docker-entrypoint.sh files are unused


fvanderbiest:
```
$ git grep ENTRYPOINT
atlas/src/docker/Dockerfile:ENTRYPOINT [ "/docker-entrypoint.sh" ]
ldap/Dockerfile:ENTRYPOINT [ "/docker-entrypoint.sh" ]

$ find . -name docker-entrypoint.sh
./geowebcache-webapp/src/docker/docker-entrypoint.sh
./geoserver/webapp/src/docker/docker-entrypoint.sh
./geoserver/webapp/target/docker/docker-entrypoint.sh
./geonetwork/web/src/docker/docker-entrypoint.sh
./geonetwork/web/target/docker/docker-entrypoint.sh
./ldap/docker/docker-entrypoint.sh
./extractorapp/src/docker/docker-entrypoint.sh
./mapfishapp/src/docker/docker-entrypoint.sh
./atlas/src/docker/docker-entrypoint.sh
```

severo:
See:

- #2438
- #2439 - be careful before merging that one (read the comments)

severo:
#2439 replaced by #2460

PR #2439, opened 2019-01-31
Remove unused files in GeoServer

severo:
Be careful - it does not change the current behaviour, but maybe the
00-geoserver-bootstrap script was intended to be run (but was not).

In that case, we instead should add something like

```
ENTRYPOINT [ "/docker-entrypoint.sh" ]
```

inside the Dockerfile.


severo:
See #2226

fvanderbiest:
You're completely right, it's meant to be run in container entrypoint.
I do not understand how this got removed at some point.

severo:
OK, I will add it

severo:
Changed to #2460

PR #2443, opened 2019-01-31
Fix dead links in mapfishapp

severo:


severo:
See also https://github.com/georchestra/georchestra/pull/2443.

PR #2446, opened 2019-02-01
analytics - extractorapp: replace the Ext.js git submodule by a webjar

groldan:
Webjars (https://www.webjars.org/) are client side JS
libraries packaged into a .jar file.

According to the documentation for servlet 3 apps
(https://www.webjars.org/documentation#servlet3) we just need
to add the required dependency.

This patch replaces the Ext.js git submodule by the extjs
4.1.1a jar dependency in both extractorapp and analytics,
saving ~500M of disk space in the developer's source code
checkout.

### DIFF
--- a/.gitmodules
+++ b/.gitmodules
@@ -4,12 +4,6 @@
 [submodule "mapfishapp/src/main/webapp/lib/externals/openlayers"]
 	path = mapfishapp/src/main/webapp/lib/externals/openlayers
 	url = https://github.com/georchestra/openlayers.git
-[submodule "extractorapp/src/main/webapp/admin/extjs"]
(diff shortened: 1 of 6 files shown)


groldan:
yes, generating a webjar is really simple and can be done with a maven project that saves static resources under `META-INF/resources/webjars/<library name>/<version>`.
As for using them, there's an locator dependency that can be added to resolve the installed version so you don't need to set it in the import paths (e.g. `src="extjs/ext.js` would automatically resolve to `extjs/<installed version>/ext.js`.

fvanderbiest:
Sounds very interesting Gabriel !
Please update the branch, there are conflicting files.

Issue #2448, opened 2019-02-01
[console] doesnt allow dot in login


landryb:
master with #2408 on top, the login field on `console/manager/#!/users/<uid>/infos` has the save button greyed out/disabled if the login field contains a dot or a dash, preventing an admin to rename a user using those chars - prenom.nom is often used, and prenom-compose.nom too so those chars should be allowed.

It seems underscore & numbers are allowed - didnt find in the code where the regex was to specify what was allowed, but istr there's been a ticket about that to harmonize the various checks done on login in the console.

landryb:
I'm not sure there should be a + after `[a-z]` either, we only want the login to *start* by a character.
```
>"prenom-compose.nom1".match(/^[a-z][\-\.\w]*$/)
Array [ "prenom-compose.nom1" ]
```

landryb:
Ah, serverside validation is in https://github.com/georchestra/georchestra/blob/master/console/src/main/java/org/georchestra/console/ws/newaccount/NewAccountFormController.java#L322 - and does indeed accept dot and dash:
```
Pattern regexp = Pattern.compile("[a-zA-Z][a-zA-Z0-9.-]*");
```
some harmonization is definitely due..

landryb:
I think the specification is 'letters, numbers, underscore, dash, dot' + 'starts with a letter' .. right @fvanderbiest ?

landryb:
And now i wonder if the regex in https://github.com/georchestra/georchestra/blob/18.06/console/src/main/webapp/WEB-INF/views/createAccountForm.jsp#L182 should be synchronized/fixed too. Right now it \W removed dashes, dots & underscores, which is not what we want i think..

landryb:
Ie testing `str = str.replace(/[^a-z0-9_.-]*/gi, '');` here, i'm a bit lost at what needs escaping when the js is inside jsp....

landryb:
Tested working fine, adding non-alnum chars to nom field correctly sanitizes the uid in the account creation form. Ie `nom-foo&_bar` does `nom-foo_bar`

fvanderbiest:
Many thanks Landry !

PR #2454, opened 2019-02-05
Harmonize uid check regexp - accept dot and dash in the frontend (#2448)

landryb:
while here properly escape dot and dash in the serverside validation.

i'm not 100% sure about escaping - at the end of a class.. but pretty sure dot needs to be escaped, otherwise the regexp reduces to `[a-z].*` i think...

### DIFF
--- a/console/src/main/webapp/manager/app/templates/userForm.tpl.html
+++ b/console/src/main/webapp/manager/app/templates/userForm.tpl.html
@@ -108,8 +108,8 @@
           {{model.uid}} (<a href="javascript:void(0)" ng-click="view.edit_login=!view.edit_login">change</a>)
         </div>
         <div class="col-sm-8" ng-if="view.edit_login">
-          <!-- Server validation should match this pattern: /^[a-z]+\w*$/ -->
-          <input ng-model="model.uid" class="form-control" id="uid" placeholder="{{'user.login' | translate}}" ng-required="required.uid" ng-pattern="'^[a-z]+\\\\w*$'">
+          <!-- Server validation is in java/org/georchestra/console/ws/newaccount/NewAccountFormController.java -->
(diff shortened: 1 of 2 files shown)


landryb:
And remember i'm no java expert either - so might be wrong on the dot escaping, but anyway the two regexps don't match - and that's the bug i want to squash here, accepting dot and dash :)

console/src/main/webapp/manager/app/templates/userForm.tpl.html
```diff
@@ -108,8 +108,8 @@
           {{model.uid}} (<a href="javascript:void(0)" ng-click="view.edit_login=!view.edit_login">change</a>)
         </div>
         <div class="col-sm-8" ng-if="view.edit_login">
-          <!-- Server validation should match this pattern: /^[a-z]+\w*$/ -->
-          <input ng-model="model.uid" class="form-control" id="uid" placeholder="{{'user.login' | translate}}" ng-required="required.uid" ng-pattern="'^[a-z]+\\\\w*$'">
+          <!-- Server validation is in java/org/georchestra/console/ws/newaccount/NewAccountFormController.java -->
+          <input ng-model="model.uid" class="form-control" id="uid" placeholder="{{'user.login' | translate}}" ng-required="required.uid" ng-pattern="'^[a-zA-Z][a-zA-Z0-9\\\\.\\\\-]*$'">
```
tonio:
```suggestion
          <input ng-model="model.uid" class="form-control" id="uid" placeholder="{{'user.login' | translate}}" ng-required="required.uid" ng-pattern="/^[a-z][a-z0-9\\\\.\\\\-]*$/i">
```
(untested but [should work](https://docs.angularjs.org/api/ng/directive/ngPattern))

landryb:
After a bit of testing on https://www.freeformatter.com/java-regex-tester.html it seems dot and dash don't need escaping in java. weird. my interpretation of http://tutorials.jenkov.com/java-regex/index.html#escaping-characters lead me to believe the opposite..

\w in js includes underscore, so underscore should be added to the serverside regexp anyway if it's accepted in the frontend.

fvanderbiest:
Thanks Landry

PR #2458, opened 2019-02-06
Accept letters, numbers, dot, dash & underscore in uid, strip everyth…

landryb:
…ing else (#2448)

(applies on top of #2454)


fvanderbiest:
Indeed !

Issue #2452, opened 2019-02-04
[console] display user org as a link


landryb:
On the `/console/manager/#!/browse/all/users` page it'd be great to have the user org displayed as a direct link to this org page (right now it's 'just text'), instead of having to : 
- click on the org tab
- filter the org list for the corresponding org
- click on the org link

the direct link would be much faster :)

fvanderbiest:
Agreed.
And it's probably immediate.

severo:
Not immediate, because /console/private/users/ only gives the name of the org, it doesn't give any ID:

```json
[
  {
    "uid":"testuser",
    "givenName":"Test",
    "sn":"USER",
    "org":"Project Steering Committee",
    "mail":"psc+testuser@georchestra.org",
    "pending":false
  },{
    "uid":"testreviewer",
    "givenName":"Test",
    "sn":"REVIEWER",
    "org":"Conseil régional d'Alsace",
    "mail":"psc+testreviewer@georchestra.org",
    "pending":false
  }...
```

landryb:
outch. good catch :) Well guess serverside needs to also return shortname...

severo:
Yes, I will add it to the REST service

severo:
See #2459.

PR #2459, opened 2019-02-06
Add link to org in the list of users in console

severo:
It requires to add a field (orgId) in the JSON provided at the "private/users"
endpoint, so: changes both in backend and frontend.


severo:
It fixes #2452.

fvanderbiest:
Nice.

PR #2478, opened 2019-02-07
Remove validation related PENDING role in CAS

severo:
Before bccbe8b1b9b00ef3e5c6b0802ed2a1ef4d6edeee, the pending users (new
LDAP users waiting for a validation of their account) were stored in
ou=users,dc=georchestra,dc=org, and were assigned the "PENDING" role.
CAS had to check if a user had this role or not before conceding
authentication (pending users are not allowed to authenticate).

But since that commit, pending users are now stored in
ou=pendingusers,dc=georchestra,dc=org, while CAS is looking
into ou=users,dc=georchestra,dc=org. So, there is no more need to check
if a user is pending or not.


severo:
See also: https://github.com/georchestra/datadir/pull/127

Issue #2411, opened 2019-01-25
CAS - GeorchestraLdapAuthenticationHandler does not need to be aware of pending users anymore


fvanderbiest:
After #2184 CAS does not have to be aware of the PENDING role anymore since it looks for users in `ou=users`.
This is done via
https://github.com/georchestra/datadir/blob/e5b354269e202dcf66cf5fe2f6b8f95ca4086241/cas/cas.properties#L28

As a result, some housekeeping is required in https://github.com/georchestra/datadir/blob/master/cas/cas.properties and also in CAS's `GeorchestraLdapAuthenticationHandler`

See also #581 which introduced the feature.

severo:
(I remove my assignment as it seems @cmangeat is already working on it)

severo:
See https://github.com/georchestra/georchestra/pull/2478 and https://github.com/georchestra/datadir/pull/127

severo:
I think we can close.

fvanderbiest:
indeed.

PR #2479, opened 2019-02-07
Remove mentions to obsolete PENDING role in console 

severo:


severo:
See https://github.com/georchestra/datadir/pull/128 and #2411

PR #2498, opened 2019-02-12
Remove the default properties file, and define default inside every placeholder

severo:


severo:
See related https://github.com/georchestra/datadir/pull/131

Issue #2499, opened 2019-02-12
Error in analytics (class not found)


severo:
To reproduce:

```
git checkout 18.12
cd analytics
mvn jetty:run
```

We get the following error:

```
org.springframework.beans.factory.BeanCreationException: Error creating bean with name 'waitForDb' defined in ServletContext resource [/WEB-INF/ws-servlet.xml]: Invocation of init method failed; nested exception is java.lang.NoClassDefFoundError: org/apache/commons/pool/KeyedObjectPoolFactory
```

severo:
Related to https://github.com/georchestra/georchestra/commit/ae8dc682657ded02e8c8f985d921b1cfb95f4575.

groldan:
Hey @severo can you check if you're not running with an outdated version of georchestra/commons? can't reproduce in 18.12 at 2edfa047fa525925b806d7ccd83e1ceed680e2ca, and indeed `commons`'s `WaitForDb` does not use any connection pool anymore. 

So I'm guessing you've built analytics without building commons previously, and the outdated commons jar might be trying to use dbcp which is no longer in analytics classpath?

severo:
Exactly, my bad...

Issue #2494, opened 2019-02-12
Find another way to pass the default language property to the JSP files in mapfishapp


severo:
The `defaultLanguage` attribute is set at:

https://github.com/georchestra/georchestra/blob/c8fcd26e4b932267effac04897aca86b44d4ad3e/mapfishapp/src/main/webapp/WEB-INF/ws-servlet.xml#L70

And accessed in index.jsp using `request.getParameter`:

https://github.com/georchestra/georchestra/blob/c8fcd26e4b932267effac04897aca86b44d4ad3e/mapfishapp/src/main/webapp/WEB-INF/jsp/index.jsp#L50

It would be better to avoid using that method that also allows to access the GET parameters.

severo:
See also #2495.

fvanderbiest:
> It would be better to avoid using that method that also allows to access the GET parameters.

Yes, that's what I meant with https://github.com/georchestra/georchestra/pull/2430#discussion_r254182956

severo:
Exactly, it's the way I found to make it work, but it does not feel like the correct manner to do it.

pmauduit:
You have request.getAttribute(String) (specific to Java / Servlet) and getParameters (pure HTTP): https://docs.oracle.com/javaee/6/api/javax/servlet/ServletRequest.html#getAttribute(java.lang.String)

severo:
See #2528.

PR #2500, opened 2019-02-12
Analytics harmonize properties

severo:


severo:
- testing is blocked by https://github.com/georchestra/georchestra/issues/2499
- note that I used the same pattern as in #2494 to get access to the language property from inside a JSP file
- see related https://github.com/georchestra/datadir/pull/132

Issue #2496, opened 2019-02-12
Harmonize properties files management


severo:
I create a dedicated issue to keep track of the work to harmonize the management of properties files in the geOrchestra modules.

The work to be done:
- remove the "resources" properties files like [mapfishapp.properties](https://github.com/georchestra/georchestra/blob/18.12/mapfishapp/src/main/resources/mapfishapp.properties)
- ensure both `default.properties` and the application-specific properties file from the [datadir](https://github.com/georchestra/datadir) are accessed by `<context:property-placeholder>`
- replace the use of GeorchestraConfiguration's [`getProperty`](https://github.com/georchestra/georchestra/blob/c8fcd26e4b932267effac04897aca86b44d4ad3e/commons/src/main/java/org/georchestra/commons/configuration/GeorchestraConfiguration.java#L132) method by a property passed to a bean using a Spring placeholder (see [spring-context.xsd](http://www.springframework.org/schema/context/spring-context.xsd)).
- set a default value to every Spring placeholder.
- ensure all the parameters used in the Spring configuration files are defined in the datadir's properties files (eg, the following is not defined in the datadir)

    https://github.com/georchestra/georchestra/blob/c8fcd26e4b932267effac04897aca86b44d4ad3e/mapfishapp/src/main/webapp/WEB-INF/ws-servlet.xml#L71

Modules processed:

- [x] mapfishapp - #2497 and https://github.com/georchestra/datadir/pull/130
- [x] extractorapp - #2498 and https://github.com/georchestra/datadir/pull/131, also #2505
- [x] header - #2503
- [x] console - #2508, https://github.com/georchestra/datadir/pull/133
- [x] atlas - # 2509 and https://github.com/georchestra/datadir/pull/134
- [x] security-proxy - #2510 and https://github.com/georchestra/datadir/pull/135
- [x] cas - #2512
- [x] geoserver - nothing to do
- [x] geonetwork - nothing to do - but see https://github.com/georchestra/georchestra/issues/2379#issuecomment-465998817
- [x] geowebcache - review #2525 
- [x] analytics - #2500 and https://github.com/georchestra/datadir/pull/132, blocked by #2507 - review #2526 and [georchestra/datadir#136](https://github.com/georchestra/datadir/pull/136)

See also:

- [x] comment all the properties in the datadir and add the corresponding comments to explain what the properties are for - https://github.com/georchestra/datadir/issues/129 - review https://github.com/georchestra/datadir/pull/138
[...]

PR #2503, opened 2019-02-13
Use Spring placeholders to access datadir properties

severo:
instead of the GeorchestraConfiguration class.

### DIFF
--- a/header/src/main/webapp/WEB-INF/ws-servlet.xml
+++ b/header/src/main/webapp/WEB-INF/ws-servlet.xml
@@ -15,10 +15,20 @@
         <property name="viewClass" value="org.springframework.web.servlet.view.JstlView"></property>
         <property name="prefix" value="/WEB-INF/jsp/"></property>
         <property name="suffix" value=".jsp"></property>
+        <property name="attributes">
+            <map>
+                <entry key="defaultLanguage" value="${language:en}" />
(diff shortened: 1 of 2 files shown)


severo:
Note that I used the same pattern as in #2494 to get access to the language property from inside a JSP file

severo:
See #2496

header/src/main/webapp/WEB-INF/ws-servlet.xml
```diff
@@ -15,10 +15,20 @@
...
+            </map>
+        </property>
     </bean>
 
     <bean id="georchestraConfiguration" class="org.georchestra.commons.configuration.GeorchestraConfiguration">
         <constructor-arg value="header" />
     </bean>
 
+    <context:property-placeholder
+        location="file:${georchestra.datadir}/default.properties, file:${georchestra.datadir}/mapfishapp/mapfishapp.properties"
```
fvanderbiest: you probably mean header.properties ;-)

severo: AHHHHH

fvanderbiest: You're allowed to `git push --force` in the branch ;-)

severo: Yes. Done.

Issue #2379, opened 2019-01-15
Question: What to do with Spring placeholders vs GeorchestraConfiguration?


severo:
Currently some modules only use the [GeorchestraConfiguration](https://github.com/georchestra/georchestra/blob/master/commons/src/main/java/org/georchestra/commons/configuration/GeorchestraConfiguration.java) class to get the [datadir](https://github.com/georchestra/datadir) properties: mapfishapp, header, extractorapp
Other use it, but also Spring placeholders (see an [analytics module Spring configuration file](https://github.com/georchestra/georchestra/blob/master/analytics/src/main/webapp/WEB-INF/ws-servlet.xml#L77) for instance): analytics, console, etc.

The problems with that situation:
- in both cases we have to take care of the priority order between the properties files (mainly [default.properties](https://github.com/georchestra/datadir/blob/master/default.properties) and the module properties files, for example [console/console.properties](https://github.com/georchestra/datadir/blob/master/console/console.properties))
- complexity for the developers
- maybe we waste resources (double access and processing)

What would be the best situation, for you?

1. all with GeorchestraConfiguration?
2. all with placeholders?
3. all with another mechanism of Spring (which one)?
4. all with another mechanism of Java (which one)?
5. keep the current situation?

severo:
Note that there is a related issue:

How to manage the case when the datadir is not accessible, or when some properties are not present in the files of the datadir? Currently, we sometimes hardcode the default value in the code (see [index.jsp](https://github.com/georchestra/georchestra/blob/master/mapfishapp/src/main/webapp/WEB-INF/jsp/index.jsp#L52), we sometimes do not specify a default value, we sometime load a properties files from the code (see [security-proxy.properties](https://github.com/georchestra/georchestra/blob/master/security-proxy/src/main/webapp/WEB-INF/security-proxy.properties) for example)...

1. Hardcode the default value each time we access a property?
2. Hardcode the default value, but in a dedicated properties file?
3. Launch an exception if the value is not specified?
4. Another solution?

severo:
If the best situation is 1, or 4 -> we could use Java methods and the `@Bean` annotation as specified in https://docs.spring.io/spring-javaconfig/docs/1.0.0.m3/reference/html/creating-bean-definitions.html to create the beans.

severo:
If I understand well, we would add a new "module" in geOrchestra, that would serve all the other modules (mapfishapp, extractorapp, console, etc.) their configuration, via an http connection, at runtime?
[...]

PR #2508, opened 2019-02-13
Console - harmonize properties

severo:
Note that we removed two classes, and that we used `@Value` in some classes to access the datadir properties files, since they use the `@Controller` annotation, and do not appear in the Spring XML configuration file.

If using `@Value` is a good idea or not could be discussed (see https://github.com/georchestra/georchestra/issues/2379#issuecomment-463246745), but I did it to stick at most to #2496, and remove all occurrences of the `getProperty` method of the `GeorchestraConfiguration` class.


severo:
Oups, build failed.... Will fix it tomorrow

severo:
I had forgot to launch the integration tests (with `mvn -Pit verify`). Maybe they should be launched by default? What do you say?

To fix, I:
- used git bisect to find the guilty commit - see https://delicious-insights.com/fr/articles/git-bisect/
- edit the history to insert the fix inside the guilty commit - see https://stackoverflow.com/questions/8824971/how-to-amend-older-git-commit/18150592#18150592

fvanderbiest:
> I had forgotten to launch the integration tests (with `mvn -Pit verify`). Maybe they should be launched by default? What do you say?

Already done via 7051fec3e6cdb33d80f79cdc54b444f7819d691d ?

severo:
Yes, thanks to this, I have been warned of the error.
I mentioned when launching locally `mvn jetty:run`, or `mvn install`, for example

fvanderbiest:
There are still places where `GeorchestraConfiguration` is used. Is it expected ?

severo:
Yes, I'm only removing the occurrences where the `getProperty` method of the `GeorchestraConfiguration` is used, ie. removing the use of that class to access default.properties and the `console/console.properties` file (and similar one for the other applications).

But there are other methods that allow to access files in the datadir, or load other properties files... I let them for the moment. Maybe #2379 will lead to also remove these others uses.

PR #2509, opened 2019-02-14
Atlas harmonize properties

severo:


severo:
See also https://github.com/georchestra/datadir/pull/134.
And #2496.

Issue #2146, opened 2018-08-29
console - manage more Orgs fields


fvanderbiest:
For the Geo2France project, orgs miss these fields:
 * a description text
 * a logo URL
 * a website URL

In forms:
 * the description field might be long text, which requires a textarea field
 * both URLs should be regexp-validated

They are not mandatory by default, but the platform admin may decide that they are required, via the `requiredFields` [console property](https://github.com/georchestra/datadir/blob/d1721391786989c3e6065180c62c653ba68f3a67/console/console.properties#L27-L28).

They should be editable by a SUPERUSER via the manager `/console/manager/#!/org/xxxx/infos`

These fields should be editable when registering a new Org via the public `/console/account/new` page.

They should be displayed to the current user on his page `/console/account/userdetails`, next to the other Org-related information

fvanderbiest:
Currently they're made of a `groupOfMembers` and an `organization` :
https://github.com/georchestra/georchestra/blob/50dcb976d8b8e17da4a79564b8546e974c4c5132/ldap/georchestra.ldif#L257-L273

`groupOfMembers` provides a `description` field which is currently used to store a list of areas.
As a result, we would be able to use the `organization`'s description field (see http://www.zytrax.com/books/ldap/ape/#organization)

Storing the logo will probably require an external storage, since only `inetOrgPerson` are allowed to store `jpegPhoto`

fvanderbiest:
> Storing the logo will probably require an external storage

Discussion whether "picture uploading" be part of the feature or not. In the latter case, the logo is a url to an external ressource not managed by the console webapp.

gryckelynck:
URL seems great and enough.
Many applications and services can be used to store images (Pydio, Dropbox, Nextcloud.. or simple FTP)

landryb:
Fwiw, iff you properly exploit geonetwork, you have a contact for orgs in which you store the logo... and the file end up in geonetwork datadir iirc, with a public url.

fvanderbiest:
Good point, thanks Landry.

fvanderbiest:
I cannot find proper fields to store logo and website urls: 
```
 objectclass ( 2.5.6.4 NAME 'organization' SUP top STRUCTURAL
  MUST o
  MAY ( userPassword $ searchGuide $ seeAlso $ businessCategory $
   x121Address $ registeredAddress $ destinationIndicator $
   preferredDeliveryMethod $ telexNumber $ teletexTerminalIdentifier $
[...]

PR #2511, opened 2019-02-14
LDAP - changes to store more Org fields

fvanderbiest:
Related with #2146


ldap/georchestra.ldif
```diff
@@ -221,12 +222,75 @@ member: uid=testuser,ou=users,dc=georchestra,dc=org
...
 
+dn: uid=psc,ou=orgs,dc=georchestra,dc=org
+objectClass: top
+objectClass: person
+objectClass: organizationalPerson
+objectClass: inetOrgPerson
+cn: psc
+sn: psc
+uid: psc
+jpegPhoto:: /9j/4AAQSkZJRgABAQAAAQABAAD/2wBDAAgGBQgMFBofBgYHCg0dHhwHBwgMFB0j
```
pmauduit: So you can add a `jpegPhoto` attribute onto a `inetOrgPerson` object, but it seems that the original request was on the organizations objects, right ? (either on the `organization` or the `groupOfMembers` object)

fvanderbiest: There's no `jpegPhoto` attribute on `organizations` or the `groupOfMembers` objects.
The trick here is to reference a `inetOrgPerson` instance via the `seeAlso` attribute of the `groupOfMembers` instance

pmauduit: So, one new object for managing the organization, that would have some consequences also on the console code :(

fvanderbiest: True, this is planned.

Issue #2462, opened 2019-02-06
SP - org.springframework.security.web.firewall.RequestRejectedException


fvanderbiest:
Navigating the docker composition leads to https://georchestra.mydomain.org/geoserver/web/;jsessionid=node012l7tnwtustnl5gb6patggaor2.node0?0 which fails with

```
org.springframework.security.web.firewall.RequestRejectedException: The request was rejected because the URL contained a potentially malicious String ";"
	at org.springframework.security.web.firewall.StrictHttpFirewall.rejectedBlacklistedUrls(StrictHttpFirewall.java:265)
	at org.springframework.security.web.firewall.StrictHttpFirewall.getFirewalledRequest(StrictHttpFirewall.java:245)
	at org.springframework.security.web.FilterChainProxy.doFilterInternal(FilterChainProxy.java:193)
	at org.springframework.security.web.FilterChainProxy.doFilter(FilterChainProxy.java:177)
	at org.springframework.web.filter.DelegatingFilterProxy.invokeDelegate(DelegatingFilterProxy.java:347)
	at org.springframework.web.filter.DelegatingFilterProxy.doFilter(DelegatingFilterProxy.java:263)
	at org.eclipse.jetty.servlet.ServletHandler$CachedChain.doFilter(ServletHandler.java:1642)
	at org.georchestra.security.UTF8Filter.doFilter(UTF8Filter.java:33)
```

pmauduit:
servlet 3.0 allows to deactivate passing the jsessionid in the url (servlet containers do that when they detect they cannot use cookies for this). This requires to add the following in each `web.xml` files:

```xml
<session-config>
     <tracking-mode>COOKIE</tracking-mode>
</session-config>
```

groldan:
Related GeoServer issue: https://osgeo-org.atlassian.net/browse/GEOS-8913
 and  patch: https://github.com/geoserver/geoserver/pull/3355

groldan:
Note to self: `curl -L` = follow redirects, `curl -c` = save cookies to a file and reuse

At the geoserver side (exposed docker-compose's 8080 port) I get this:
```
$ curl -v http://localhost:8080/geoserver/web/ > /dev/null
< HTTP/1.1 302 Found
< Set-Cookie: JSESSIONID=node0vrc1uvel4zfji54t7qf5mnpt0.node0;Path=/geoserver;Secure;HttpOnly
[...]

PR #2516, opened 2019-02-19
security-proxy: handle response headers before handling redirects

groldan:
Fix for https://github.com/georchestra/georchestra/issues/2462 , in tandem with https://github.com/georchestra/geoserver/pull/17

When geoserver returns a redirect, proxy was returning
before handling response headers, which prevented the
JSESSIONID cookie to be propagated, resulting in too
many redirects.

### DIFF
--- a/security-proxy/src/main/java/org/georchestra/security/Proxy.java
+++ b/security-proxy/src/main/java/org/georchestra/security/Proxy.java
@@ -708,10 +708,14 @@ public class Proxy {
                 }
             }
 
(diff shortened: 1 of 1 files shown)


fvanderbiest:
Nice catch.

fvanderbiest:
Might be interesting to backport to 16.12 and 18.06 as well.
Ping @pmauduit

pmauduit:
I'm ok with the idea of porting it back, the code should not have evolved that much.

Issue #2519, opened 2019-02-19
extractorapp - PropertyAccessException 1: org.springframework.beans.MethodInvocationException: Property 'publicUrl' threw exception


fvanderbiest:
Witnessed with the docker composition with an up to date data dir
```
Caused by: 
org.springframework.beans.factory.BeanCreationException: Error creating bean with name 'extractor-controller' defined in ServletContext resource [/WEB-INF/ws-servlet.xml]: Error setting property values; nested exception is org.springframework.beans.PropertyBatchUpdateException; nested PropertyAccessExceptions (1) are:|PropertyAccessException 1: org.springframework.beans.MethodInvocationException: Property 'publicUrl' threw exception; nested exception is java.lang.NullPointerException
	at org.springframework.beans.factory.support.AbstractAutowireCapableBeanFactory.applyPropertyValues(AbstractAutowireCapableBeanFactory.java:1571)
[...]
	at org.eclipse.jetty.start.Main.start(Main.java:490)
	at org.eclipse.jetty.start.Main.main(Main.java:77)
Caused by: 
org.springframework.beans.PropertyBatchUpdateException; nested PropertyAccessExceptions (1) are:|PropertyAccessException 1: org.springframework.beans.MethodInvocationException: Property 'publicUrl' threw exception; nested exception is java.lang.NullPointerException
	at org.springframework.beans.AbstractPropertyAccessor.setPropertyValues(AbstractPropertyAccessor.java:121)
```

severo:
Also thrown with:

```
cd extractorapp
mvn jetty:run
```

severo:
See #2524.

PR #2524, opened 2019-02-21
Extractorapp - Fix exception when setting publicUrl property

severo:
Bug introduced by https://github.com/georchestra/georchestra/commit/156e342b0260d3ea79bc9de059027a2a247d071a#diff-cb3d8cd657ccafb05e41f9f2cf21a7d8L103.

Note: here we rename the setPublicUrl method since in reality it sets the secureHost and the servletUrl properties, not the publicUrl property.


severo:
See #2519.

PR #2525, opened 2019-02-21
geowebcache - harmonize properties files management

severo:
- removes dependency to the GeorchestraConfiguration class
- deletes the hardcoded properties file
- use placeholders in Spring XML configuration file
- sets defaults for the placeholders


severo:
See #2496

Issue #2507, opened 2019-02-13
analytics - why not using a simpler comma separated list for the excluded users?


severo:
See 
https://github.com/georchestra/georchestra/blob/2387e6f374a0adf56f6d896a5e8a23e31ec14308/analytics/src/main/java/org/georchestra/analytics/StatisticsController.java#L697-L706

and

https://github.com/georchestra/datadir/blob/e6e08d2247174cb8a1a4158a8a84c258eaf4a839/analytics/analytics.properties#L12-L15

Note that it blocks passing the list as a parameter through Spring placeholders (see #2496).

fvanderbiest:
If you can replace the current config with a comma separated list, I'm fine with it

severo:
Note that it seems the comma separated list is mapped directly to an Array: https://www.baeldung.com/spring-value-defaults

severo:
See also

https://github.com/georchestra/datadir/blob/2ce6e4d6eb0cd3fb2b45baf29cec0e1a6efd8f0f/security-proxy/removed-xforwarded-headers.properties#L1-L3

and where it is accessed:

https://github.com/georchestra/georchestra/blob/2387e6f374a0adf56f6d896a5e8a23e31ec14308/security-proxy/src/main/java/org/georchestra/security/RemoveXForwardedHeaders.java#L64-L76

severo:
I suspect there are more occurrences of this pattern, that could be replaced with comma separated list. It would require scanning all the code.

severo:
See #2526, and https://github.com/georchestra/datadir/pull/136.

If merged (please also test by your own before to be sure this does not break), we could look for other occurrences of that pattern and remove them all.

severo:
See also:

- [ ] https://github.com/georchestra/datadir/blob/18.12/security-proxy/removed-xforwarded-headers.properties
- [ ] https://github.com/georchestra/datadir/blob/18.12/security-proxy/trusted-users.properties
- [ ] https://github.com/georchestra/datadir/blob/18.12/security-proxy/user-agents.properties

PR #2526, opened 2019-02-21
analytics - Use only one comma-separated list of excluded users as property

severo:
Instead of looking for properties with a suffix (excludedUser1,
excludedUser2, etc.)
Note the use of the ConversionServiceFactoryBean to handle the
conversion from comma separated list to a Collection, here Set<String> (more
at: https://stackoverflow.com/a/29970335/7351594)

### DIFF
--- a/analytics/src/main/java/org/georchestra/analytics/StatisticsController.java
+++ b/analytics/src/main/java/org/georchestra/analytics/StatisticsController.java
@@ -37,7 +37,6 @@ import javax.servlet.http.HttpServletResponse;
 import javax.sql.DataSource;
 
 import org.georchestra.analytics.util.QueryBuilder;
-import org.georchestra.commons.configuration.GeorchestraConfiguration;
 import org.joda.time.DateTime;
 import org.joda.time.DateTimeZone;
 import org.joda.time.Duration;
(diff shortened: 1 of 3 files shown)


severo:
See #2507 and https://github.com/georchestra/datadir/pull/136.

Please test well before merging to be sure it does not break. Maybe I could add a test, but I'm not sure how to implement it (Spring related). Any hint is welcome.

fvanderbiest:
Live tested, datadir modified with `excludedUsers=testadmin`, the requests still shows testadmin stats

```
$ curl --silent 'https://georchestra.mydomain.org/analytics/ws/distinctUsers' -H 'content-type: application/json;charset=UTF-8' -u testadmin:testadmin --data-binary '{"service":"distinctUsers","startDate":"2019-03-14","endDate":"2019-03-15"}' --insecure | jq .results[0]
{
  "organization": "Project Steering Committee",
  "nb_requests": 95,
  "user": "testadmin"
}
```

severo:
> Needs more work

It was... a typo! But you spotted it with your tests, thanks! I have been able to test correctly and check the property has the expected effect, with property set to 0, 1 or 2 comma separated usernames.

severo:
I checked, there is no other occurrences of the same typo (a ',' instead of a ':' to separate the property name from its default value)

PR #2528, opened 2019-02-25
use getAttribute method in JSP to specifically get request attribute

severo:
and not a possible HTTP request parameter.


severo:
Fixes #2494

PR #2531, opened 2019-02-25
Remove getProperty methods from commons

severo:
The properties files from the datadir are loaded through the Spring
mechanisms, and use the Spring XML configuration files placeholders.
The getProperty methods are no more used.

See https://github.com/georchestra/georchestra/issues/2496 for the
motivation.


severo:
See #2496.

Be careful: only merge after #2526!

PR #2540, opened 2019-03-07
LDAP - parametrize slapd log level in docker

severo:
Allow to configure the log level of slapd (LDPA server) - see http://www.zytrax.com/books/ldap/ch6/#loglevel

### DIFF
--- a/ldap/Dockerfile
+++ b/ldap/Dockerfile
@@ -40,7 +40,7 @@ VOLUME ["/etc/ldap", "/var/lib/ldap"]
 
 ENTRYPOINT [ "/docker-entrypoint.sh" ]
 
-CMD [ "sh", "-c", "exec slapd -d 32768 -u ${RUN_AS_UID} -g ${RUN_AS_GID}" ]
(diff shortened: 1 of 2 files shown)


severo:
See also https://github.com/georchestra/georchestra/pull/2540.

ldap/Dockerfile
```diff
@@ -40,7 +40,7 @@ VOLUME ["/etc/ldap", "/var/lib/ldap"]
 
 ENTRYPOINT [ "/docker-entrypoint.sh" ]
 
-CMD [ "sh", "-c", "exec slapd -d 32768 -u ${RUN_AS_UID} -g ${RUN_AS_GID}" ]
+CMD [ "sh", "-c", "exec slapd -d ${SLAPD_LOG_LEVEL:-32768} -u ${RUN_AS_UID} -g ${RUN_AS_GID}" ]
```
fvanderbiest: IIRC this new env var has to be defined in an ENV statement before (but docker is nice to us and does not bark)

severo: No, simply, if the SLAPD_LOG_LEVEL variable has been set, it's used, otherwise the default value (32768) is used as before

PR #2542, opened 2019-03-11
proxy - skip BasicAuth filter if user has already been authenticated

severo:


severo:
Note that we could also consider the comment in https://github.com/georchestra/georchestra/commit/f050704d0f5ac752a65f9fcb5b601a2331ec82bf:

```
* What about dropping BasicAuthByUserAgent class ? I thought we agreed
  on using the Basic-Auth if present in any cases. This would allow us
  to drop the code and rely on upstream spring-security class.
```

fvanderbiest:
> What about dropping BasicAuthByUserAgent class ? 

I think several customers are using it actively (eg for QGIS pro users)

severo:
> I think several customers are using it actively (eg for QGIS pro users)

Sure, but the idea is to allow any User Agent to authenticate with Basic Auth, not to remove the possibility to authenticate with QGis.

PR #2545, opened 2019-03-12
Update 'description' values to match samples in foncier plugin (branch 18.06)

jeanpommier:
While still matching with https://www.geopicardie.fr/public/communes_simplified.json used in organizations' area of activity


fvanderbiest:
thanks !

Issue #2149, opened 2018-08-30
GDPR compliance - track user consent to platform terms of use (including data treatments)


fvanderbiest:
On `/console/account/new`, add a checkbox "agree to platform terms" (provides a configurable http link to platform terms - writing such a document is out of geOrchestra scope)  
Keeping track of this consent can probably be achieved by storing in LDAP a timestamp for this consent.

severo:
Base OID for geOrchestra project: 1.3.6.1.4.1.53611 (see https://www.iana.org/assignments/enterprise-numbers/enterprise-numbers, http://www.oid-info.com/get/1.3.6.1.4.1.53611)

severo:
See also https://github.com/georchestra/georchestra.github.io/pull/86.

fvanderbiest:
Seems like the PR broke master

https://travis-ci.org/georchestra/georchestra/builds/517165502
2019-04-08 09:37:47,236 [main] WARN  org.springframework.web.context.support.GenericWebApplicationContext - Exception encountered during context initialization - cancelling refresh attempt: org.springframework.beans.factory.UnsatisfiedDependencyException: Error creating bean with name 'newAccountFormController': Unsatisfied dependency expressed through field 'privacyPolicyAgreementUrl'; nested exception is org.springframework.beans.factory.NoSuchBeanDefinitionException: No qualifying bean of type 'java.lang.String' available: expected at least 1 bean which qualifies as autowire candidate. Dependency annotations: {@org.springframework.beans.factory.annotation.Autowired(required=true)}

severo:
Right. I forgot to launch `mvn -Pit verify`...

Should be fixed with https://github.com/georchestra/georchestra/pull/2589

severo:
The REST API allows to modify the date, but currently I think the web application is the only client, so if we put it read-only, I think we could also remove the possibility to modify the field from the API.
What would be the way to modify the data in case it's needed? Only modifying in hard in the database?

fvanderbiest:
> One remark : on page /console/manager/#!/users/XXX/infos, the field should be read-only because one could accidentally modify the value.

Adding `disabled` attribute to the input field.

fvanderbiest:
See https://github.com/georchestra/georchestra/commit/46a10f1db92922237a4056baf2ac15e4af7f7af6

PR #2546, opened 2019-03-13
GDPR - track user consent

severo:
Note: we must first [get an OID](http://pen.iana.org/pen/PenApplication.page) for the geOrchestra project, and set it in the code, before merging. The request have been made, just waiting for the registry.


severo:
See #2149.

console/src/main/java/org/georchestra/console/ws/backoffice/users/UsersController.java
```diff
@@ -604,6 +606,14 @@ private Account modifyAccount(Account account, ServletInputStream inputStream) t
 				account.setShadowExpire((new SimpleDateFormat("yyyy-MM-dd")).parse(shadowExpire));
 		}
 
+		String privacyPolicyAgreementDate = RequestUtil.getFieldValue(json, UserSchema.PRIVACY_POLICY_AGREEMENT_DATE_KEY);
```
cmangeat: Maybe should accoun dao impl choose what to do with empty date... In my option opinion, field should be let blank in dto. Here logic is if "", then set it to null, and at dao side if null delete it.

severo: Yes, it would be better

console/src/main/java/org/georchestra/console/ws/backoffice/users/UsersController.java
```diff
@@ -676,6 +686,16 @@ private Account createAccountFromRequestBody(ServletInputStream is) throws Illeg
 			}
 		}
 
+		String privacyPolicyAgreementDate = RequestUtil.getFieldValue(json, UserSchema.PRIVACY_POLICY_AGREEMENT_DATE_KEY);
```
cmangeat: Parse errors should be discarded at validation time ?

severo: I don't understand, sorry. What is the proposal?

cmangeat: Json should be first validated, I think it is, looking for example for mandatory field, is it ?
In this case we should also validate that date format are ok so to avoid later parse exception ?
One should also try to reuse same code in account controller for update and modify service (building an account from a json ?).

Just saying that small refactor steps are welcomed in my opinion.

severo: OK, I will work on it.

cmangeat: This sais I am probably messing things with OrgControler as 

https://github.com/georchestra/georchestra/pull/2546/files/4fb587e287128730054495a5effaf504a9c9f842#diff-0ceccf060c18478dcaeeef338384f688L596

states for example that if org is messing for json, let it unchanged.

severo:
OID -> https://github.com/georchestra/georchestra.github.io/pull/86

console/src/main/java/org/georchestra/console/ws/backoffice/users/UsersController.java
severo: > In this case we should also validate that date format are ok so to avoid later parse exception ?

I think it's what the exceptions are for. The two expected behaviors are:

- empty field ("") -> we do not set the privacyPolicyAgreementDate field in the Account object
- some value -> it should be formatted as expected ('2007-12-03') by LocalData.parse(), or an exception would be thrown because the JSON is malformed.

(note that I copied and adapted the code above, for shadowExpire)

severo: > One should also try to reuse same code in account controller for update and modify service (building an account from a json ?).

Yes, I agree.

severo: > an exception would be thrown because the JSON is malformed

I wonder if we are managing the errors correctly in that REST API (I don't think so). I imagine that in https://github.com/georchestra/georchestra/blob/master/console/src/main/java/org/georchestra/console/ws/backoffice/users/UsersController.java#L324 and https://github.com/georchestra/georchestra/blob/master/console/src/main/java/org/georchestra/console/ws/backoffice/users/UsersController.java#L392 endpoints, we should manage the case where the JSON is malformed, and return the corresponding HTTP error code, instead of failing with an exception on the server side.

console/src/main/java/org/georchestra/console/ws/backoffice/users/UsersController.java
severo: Hmmm, I think we should keep it this way: if the field is not present in JSON, the Account field is let unmodified. But if the JSON contains the field, specifically set to empty value, I imagine it's logical to understand that the Account field should be emptied (obviously, this behavior should be written in the API doc... to be written in the same style than https://app.swaggerhub.com/apis-docs/groldan/extractorapp/1.0.0).
If we don't do that way, we remove the possibility to the user to empty that field.

fvanderbiest:
Docker image `georchestra/ldap:latest` updated

PR #2548, opened 2019-03-18
seems that init methods have disappeared

cmangeat:


fvanderbiest:
They also disappeared on 18.12 => this PR should be rebased on top of 18.12

PR #2554, opened 2019-03-19
Proxy streaming when using async client 18.12

cmangeat:
see https://github.com/georchestra/georchestra/pull/2547.

### DIFF
--- a/security-proxy/src/main/java/org/georchestra/security/Proxy.java
+++ b/security-proxy/src/main/java/org/georchestra/security/Proxy.java
@@ -20,7 +20,6 @@
 package org.georchestra.security;
 
 import com.google.common.annotations.VisibleForTesting;
-import com.google.common.io.Closer;
 import org.apache.commons.lang3.StringUtils;
 import org.apache.commons.logging.Log;
 import org.apache.commons.logging.LogFactory;
@@ -52,7 +51,6 @@ import org.apache.http.impl.conn.SystemDefaultRoutePlanner;
 import org.apache.http.impl.nio.client.CloseableHttpAsyncClient;
 import org.apache.http.impl.nio.client.HttpAsyncClientBuilder;
 import org.apache.http.impl.nio.client.HttpAsyncClients;
-import org.apache.http.message.BasicHttpRequest;
 import org.apache.http.message.BasicNameValuePair;
 import org.apache.http.nio.ContentDecoder;
 import org.apache.http.nio.IOControl;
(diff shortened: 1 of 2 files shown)


security-proxy/src/main/java/org/georchestra/security/Proxy.java
```diff
@@ -787,7 +738,58 @@ private Header extractHeaderSetCookie(HttpResponse proxiedResponse) {
...
+                HttpEntity streamEntity = new InputStreamEntity(pis, contentType);
+                httpResponse.setEntity(streamEntity);
+                future.complete(httpResponse);
+            }
+
+            protected void onContentReceived(ContentDecoder decoder, IOControl ioctrl) throws IOException {
+                int bytesRead = decoder.read(this.bbuf);
+                if (bytesRead > 0) {
+                    this.bbuf.flip();
+                    channel.write(this.bbuf);
```
groldan: Beware the output channel have written less than bbuf.remaining bytes, I guess we should make it something like ```while(this.bbuf.hasRemaining()) {
                    	channel.write(this.bbuf);
                    }```

security-proxy/src/main/java/org/georchestra/security/Proxy.java
```diff
@@ -787,7 +738,58 @@ private Header extractHeaderSetCookie(HttpResponse proxiedResponse) {
 
     @VisibleForTesting
     protected HttpResponse executeHttpRequest(CloseableHttpAsyncClient httpclient, HttpRequestBase proxyingRequest) throws IOException, TimeoutException, ExecutionException, InterruptedException {
-        Future<HttpResponse> future = httpclient.execute(proxyingRequest, null);
+        CompletableFuture<HttpResponse> future = new CompletableFuture<HttpResponse>();
+
+        PipedOutputStream pos = new PipedOutputStream();
```
groldan: Mape `pos`, `pis`, and `channel` members of the `consumer` inner class? Also add `@Override` annotations to `onEntityEnclosed`, `onContentReceived`, and `releaseResources`?

security-proxy/src/main/java/org/georchestra/security/Proxy.java
```diff
@@ -251,16 +260,13 @@ public void init() throws Exception {
 
         // Proxy permissions not set by datadir
         if (proxyPermissionsFile != null && proxyPermissions == null) {
-            Closer closer = Closer.create();
-            try {
+            try (Closer closer = Closer.create()) {
```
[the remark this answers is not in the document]
cmangeat: thank you !

security-proxy/src/main/java/org/georchestra/security/Proxy.java
cmangeat: thank you !

security-proxy/src/main/java/org/georchestra/security/Proxy.java
cmangeat: thank you !

PR #2557, opened 2019-03-20
gwc - remove useless RUN cmd in dockerfile

fvanderbiest:
This was previously used to install `libjai-core-java` & `libjai-imageio-core-java` but these packages are not available anymore.


fvanderbiest:
BTW @groldan do you know how the gwc community deals with it now that JAI is not available anymore ?

PR #2558, opened 2019-03-20
Use default properties from datadir's default.properties file

severo:
It's required because all the application properties will be commented by default in the datadir. See https://github.com/georchestra/datadir/pull/138 and https://github.com/georchestra/datadir/pull/137.


fvanderbiest:
`publicUrl` being always defined, as all other properties from `default.properties`, there's no need specifying a default value like `https://georchestra.mydomain.org` in `${publicUrl:https://georchestra.mydomain.org}` IMO

Same holds true for other vars from `default.properties`

Issue #2555, opened 2019-03-20
gwc standalone - do not store the geowebcache.xml config file next to the tile cache


fvanderbiest:
Backup strategies are not the same for a text file and a dir filled with lots of tiles...

The doc mentions: "location of the configuration file is set in WEB-INF/geowebcache-core-context.xml. If you wish to move the configuration file to a different location, open geowebcache-core-context.xml and look for the <bean> tag referencing gwcXmlConfig:
Uncomment the bottom <constructor-arg value= ...> tag and specify the location where you would like to place geowebcache.xml."

fvanderbiest:
Same was done in https://github.com/PDOK/geowebcache-docker/blob/master/generate-tomcat-dockerfile.sh#L30-L32

fvanderbiest:
This requires having an unzipped geowebcache war in the docker image.

fvanderbiest:
... and I need a helping hand from our maven gurus here ;-)

severo:
Fixed by https://github.com/georchestra/georchestra/pull/2560 and https://github.com/georchestra/datadir/pull/144

PR #2560, opened 2019-03-22
gwc - allow to configure the configuration file directory

severo:
Now, by default, the configuration file will be located in the /tmp/gwc_config/
directory, but it can be overriden by the configurationDir property in
the datadir.

If the configuration directory does not exist, it is created at runtime
(if allowed).
If the configuration file does not exist, it is also created at runtime (if
allowed), as a copy of the template file located in the geowebcache
cache dir (geowebcache.xml).

It's meant to separate data file from configuration files (see
https://github.com/georchestra/georchestra/issues/2555).

### DIFF
--- a/geowebcache-webapp/src/main/webapp/WEB-INF/geowebcache-servlet.xml
+++ b/geowebcache-webapp/src/main/webapp/WEB-INF/geowebcache-servlet.xml
@@ -40,12 +40,18 @@ http://www.springframework.org/schema/context http://www.springframework.org/sch
         <constructor-arg ref="gwcTLDispatcher"/>
         <constructor-arg ref="gwcGridSetBroker"/>
         <constructor-arg ref="gwcStorageBroker"/>
-        <constructor-arg ref="gwcXmlConfig"/>
+        <constructor-arg ref="gwcXmlConfigGeorchestra"/>
         <constructor-arg ref="gwcRuntimeStats"/>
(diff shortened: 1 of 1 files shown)


severo:
See https://github.com/georchestra/georchestra/pull/2560/commits/41961de98c05f83476c386bbd017d9dde86d3ba1 and #2555.

geowebcache-webapp/src/main/webapp/WEB-INF/geowebcache-servlet.xml
```diff
@@ -40,14 +40,20 @@ http://www.springframework.org/schema/context http://www.springframework.org/sch
...
         <property name="defaultStorageFinder" ref="gwcDefaultStorageFinder"/>
         <property name="instanceName" value="${instanceName}"/>
         <property name="headerUrl" value="${headerUrl}"/>
         <property name="headerHeight" value="${headerHeight}"/>
     </bean>
 
+    <!-- The location of a static configuration file for GeoWebCache. -->
+    <bean id="gwcXmlConfigGeorchestra" class="org.geowebcache.config.XMLConfiguration">
+        <constructor-arg ref="gwcAppCtx" />
+        <constructor-arg value="${configurationDir:/tmp/gwc_config}" />
```
fvanderbiest: `/etc/georchestra/geowebcache` ?

PR #2567, opened 2019-03-25
SP - following spring-security 3.2.10 bump

pmauduit:
The CustomAuthenticationProvider class allows to first try a bind on an
Active Directory, then (if failing) trying another time on the
geOrchestra LDAP server. It is mainly used in the Rennes-Métropole
setup. Following the spring-security version update, the getObject()
method is expecting to return spring-based objects and not the ones from
the official Java API anymore. We then need to configure a specific
factory for this kind of objects so that spring-security based classes
are used.

Tests: runtime tested on Rennes-Métropole


severo:
Small and brilliant commit

Issue #2569, opened 2019-03-28
broken header in cas


landryb:
headerUrl/headerHeight dont have a default value in https://github.com/georchestra/georchestra/blob/18.12/cas-server-webapp/src/main/webapp/WEB-INF/view/jsp/default/ui/includes/top.jsp#L47 sam as #2566 @severo ?

landryb:
Oh, and having headerHeight set to 60 in default.properties doesnt propagate to this JSP it seems.. with headerHeight commented out in cas.properties.

severo:
Normally the JSP templates get these placeholders (that coincidentally are named the same as the default properties, but could be named differently) from https://github.com/georchestra/georchestra/blob/18.12/cas-server-webapp/src/main/webapp/WEB-INF/cas-servlet.xml#L75-L76.

Did you experience a problem with these placeholders?

landryb:
Well, right now the resulting html has 'src=""' and 'height:px' so they have no value. The only spot i've set those is in default.properties... and i see that cas-servlet.xml has no default value for those anyway ?

severo:
Exactly, the logic is to have mandatory properties in default.properties, and not to set a default value for them when they are called.
But something occurs that prevent the JSP to access the properties. I will try to reproduce.

landryb:
Setting headerHeight/headerUrl in cas.properties doesnt seem to propagate either, i still get ```<iframe src="" style="width:100%;height:px;border:none;overflow:hidden;" scrolling="no" frameborder="0"></iframe>``` on the cas login page, ie header not shown at all..

landryb:
I dont think it matters, but we use our own cas theme, ie not `default` but `WEB-INF/view/jsp/craig`, but i see that the `urlBasedViewResolver` in `cas-servlet.xml` seems to properly include that path.

severo:
I reproduce. I'm investigating

severo:
This is the only module that uses ResourceBundleViewResolver, instead of UrlBasedViewResolver (there is a UrlBasedViewResolver bean in CAS, but i's not used). And ResourceBundleViewResolver does not seem to allow passing arguments to the JSP from the context.
First solution: revert the changes in JSP, and get the getProperty() method back in commons module.
Second solution: use the UrlBasedViewResolver instead of ResourceBundleViewResolver.

landryb:
@pmauduit as the resident spring/java expert, any hint ?

severo:
[...]

PR #2578, opened 2019-04-02
Fix cas properties declaration in JSP.

RemiDesgrange:
The new version is less readable than  `${myVar}` but much better than
the solution in 18.06 version.

### DIFF
--- a/cas-server-webapp/src/main/webapp/WEB-INF/view/jsp/default/ui/casLoginView.jsp
+++ b/cas-server-webapp/src/main/webapp/WEB-INF/view/jsp/default/ui/casLoginView.jsp
@@ -20,7 +20,7 @@
 --%>
 <jsp:directive.include file="includes/top.jsp" />
 
-<spring:eval expression="@propertyConfigurer.getProperty('consoleContextpath')" var="consoleContextpath"/>
+<spring:eval expression="@propertyConfigurer.getProperty('console.contextpath') ?: '/console'"  var="consoleContextpath"/>
 <spring:eval expression="@propertyConfigurer.getProperty('loginTicket')" var="loginTicket"/>
 <spring:eval expression="@propertyConfigurer.getProperty('flowExecutionKey')" var="flowExecutionKey"/>
 
@@ -66,7 +66,8 @@
     </section>
 
     <section class="row">
-      <a href="${consoleContextpath}/account/new"><spring:message code="screen.welcome.link.newaccount" /></a> - <a href="${consoleContextpath}/account/passwordRecovery"><spring:message code="screen.welcome.link.forgot" /></a>
+      <a href="${consoleContextpath}/account/new"><spring:message code="screen.welcome.link.newaccount" /></a>
+      - <a href="${consoleContextpath}/account/passwordRecovery"><spring:message code="screen.welcome.link.forgot" /></a>
     </section>
 
   </form:form>
--- a/cas-server-webapp/src/main/webapp/WEB-INF/view/jsp/default/ui/casLoginView.jsp
+++ b/cas-server-webapp/src/main/webapp/WEB-INF/view/jsp/default/ui/casLoginView.jsp
@@ -20,6 +20,10 @@
 --%>
 <jsp:directive.include file="includes/top.jsp" />
 
+<spring:eval expression="@propertyConfigurer.getProperty('consoleContextpath')" var="consoleContextpath"/>
+<spring:eval expression="@propertyConfigurer.getProperty('loginTicket')" var="loginTicket"/>
+<spring:eval expression="@propertyConfigurer.getProperty('flowExecutionKey')" var="flowExecutionKey"/>
+
 <div class="box" id="login">
   <form:form method="post" id="fm1" commandName="${commandName}" htmlEscape="true">
 
--- a/Makefile
+++ b/Makefile
@@ -38,6 +38,9 @@ docker-build-geoserver-geofence: docker-pull-jetty
 docker-build-proxy: build-deps docker-pull-jetty
 	mvn clean package docker:build -Pdocker -DskipTests --pl security-proxy
 
+docker-build-cas: build-deps docker-pull-jetty
+	mvn clean package docker:build -Pdocker -DskipTests --pl cas-server-webapp
+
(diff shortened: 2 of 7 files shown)


landryb:
If this is for #2569, don't forget that there are other instances of `${instanceName}` and that `${consoleContextPath}` & `${publicUrl}` also needs handling. Will test your PR .. thanks!

landryb:
Fwiw, that seems to work here at runtime, tested with `headerHeight` & `headerUrl` defined in `cas.properties`, then commented out & only defined in `defaults.properties`, the values seems properly propagated - `instanceName` too.

Minor nit but for cleanlyness, the reindenting done in the first chunk of `propertyFileConfigurer.xml` is unrelated/unneeded.

So the principle looks right, so if you can also fix `consoleContextpath` in https://github.com/georchestra/georchestra/blob/19.04/cas-server-webapp/src/main/webapp/WEB-INF/view/jsp/default/ui/casLoginView.jsp#L65, `publicUrl` in https://github.com/georchestra/georchestra/blob/19.04/cas-server-webapp/src/main/webapp/WEB-INF/view/jsp/default/ui/casGenericSuccess.jsp#L24 and `instanceName` in https://github.com/georchestra/georchestra/blob/19.04/cas-server-webapp/src/main/webapp/WEB-INF/view/jsp/default/ui/includes/top-noheader.jsp#L35 (hope i havent forgotten any) i think that'll be all green for me to merge

RemiDesgrange:
I found a clearer solution. testing it and pushing asap

landryb:
That indeed looks clearer, nice !

cas-server-webapp/src/main/webapp/WEB-INF/view/jsp/default/ui/casLoginView.jsp
```diff
@@ -20,6 +20,10 @@
 --%>
 <jsp:directive.include file="includes/top.jsp" />
 
+<spring:eval expression="@propertyConfigurer.getProperty('consoleContextpath')" var="consoleContextpath"/>
+<spring:eval expression="@propertyConfigurer.getProperty('loginTicket')" var="loginTicket"/>
+<spring:eval expression="@propertyConfigurer.getProperty('flowExecutionKey')" var="flowExecutionKey"/>
+
```
landryb: Are you sure loginTicket & flowExecutionKey need special treatment here ? to me only the vars from https://github.com/georchestra/georchestra/blob/19.04/cas-server-webapp/src/main/webapp/WEB-INF/cas-servlet.xml#L70 do...

landryb:
Something feels wrong about `${consoleContextPath}` replacement, here i dont define a value in default.properties nor cas/cas.properties (where its defined as `cas/cas.properties:#console.contextpath=/console`), there's a default value in `cas-servlet.xml`
```
<entry key="consoleContextPath" value="${console.contextpath:/console}" />
```
but adding
```
<spring:eval expression="@propertyConfigurer.getProperty('consoleContextpath')" var="consoleContextpath"/>
```
to `casLoginView.jsp` doesnt seem enough, there's still an empty value for `consoleContextpath` and the two links displayed on the cas login page aren't prefixed by `/console` as expected by the defaults.

landryb:
Uncommenting the default value in `cas/cas.properties` isnt enough to fix it, one needs to change the `casLoginView.jsp` line for:
```
<spring:eval expression="@propertyConfigurer.getProperty('console.contextpath')" var="consoleContextpath"/>
``` (note the right property name)

so imo, if we go that route the confusing default from `cas-servlet.xml` should be dropped, and the value should be uncommented from `cas.properties` so that it's properly set.

**Comment by RemiDesgrange** [comment]
Totally agree about `contextpath`

**Comment by RemiDesgrange** [comment]
Ok @landryb I fixed it for 19.04. but the whole config loading in cas need to be rethink, there's stuff everywhere.

**Comment by landryb** [comment]
Tested it here, seems to work fine (commented out values set in `cas.properties`, so `headerHeight/instanceName/headerUrl` properly come from `defaults.properties`), but it'd be better if someone else also tests it at runtime :) @fvanderbiest ?.

I wouldnt add the `loginTicket/flowExecutionKey` vars though...

**Comment by RemiDesgrange** [comment]
WDYT, ATM, this PR seems really fragile... I'm working on another solution but I'm fighting against spring, and, for now, it wins :-/

**Comment by groldan** [comment]
I think instead of doing `expression="@propertyConfigurer.getProperty(...)`  you can do  `expression="@environment.getProperty(...)` so it gets whatever value was already resolved by spring, being it a default or provided value

**Comment by landryb** [comment]
> I think instead of doing `expression="@propertyConfigurer.getProperty(...)` you can do `expression="@environment.getProperty(...)` so it gets whatever value was already resolved by spring, being it a default or provided value

which means there's no need for the `context:property-placeholder` replacement by a `propertyfactory`?

**Comment by RemiDesgrange** [comment]
> which means there's no need for the `context:property-placeholder` replacement by a `propertyfactory`?

no :cry: If you remove `context:property-placeholder` then it will crash. @groldan I think I already test `@environment` and it didn't work well but not sure, I'll test again. In any case, this stuff is more of a fix. We should fund a config refactor.

**Comment by severo** [comment]
Find or fund ? Both, I imagine

**Comment by RemiDesgrange** [comment]
@landryb The problem I'm facing right now, is that the properties file loader doesn't resolve var in properties file. Ex: 
```
publicurl=${schema}://{domainName}
```
display in html : `${schema}://{domainName}`

There are solutions but none of then worked ATM. I tried using `PropertySourcesPlaceholderConfigurer` [doc](https://docs.spring.io/spring/docs/current/javadoc-api/org/springframework/context/support/PropertySourcesPlaceholderConfigurer.html) but it didn't work out.

landryb:
i like the *make it work* commits, which clearly conveys the frustration of spending hours on an issue, and finally giving up for a "dammit, will be good enough" fix ;)

severo:
Yes, I understand it the same way :) But, I think if you have spent various days on an issue, the commit message is a good occasion to give more details on what has changed, and maybe what has been tried, didn't work, why it was so hard, etc.

RemiDesgrange:
Too bad that françois merged it. For this kind of stuff I like to squash before commit.  Gitlab has this usefull feature, and think this is great !

PR #2583, opened 2019-04-04
Try to make dependency to slf4j explicit

RemiDesgrange:


fvanderbiest:
Rebase on top of 19.04 + PR to https://github.com/georchestra/datadir/blob/19.04/atlas/logback/logback.xml please.

atlas/src/main/resources/camel/atlas_app.xml
```diff
@@ -87,6 +87,10 @@
         <constructor-arg value="atlas" />
     </bean>
 
+    <bean id="georLogger" class="org.slf4j.LoggerFactory" factory-method="getLogger" xmlns="http://www.springframework.org/schema/beans">
+        <constructor-arg value="org.georchestra.georLogger" />
```
pmauduit: The main concern here is that each time the georLogger will be used, it will match the `org.georchestra.georLogger` package, which might not be intended.

Anyway, since it seems to be used only here in the XML file, that is not a big deal though. I'd have seen something like `org.georchestra.atlas.camel` which would have been more meaningfull.

atlas/src/main/java/org/georchestra/atlas/AtlasMailComponent.java
```diff
@@ -5,7 +5,6 @@
 import org.apache.camel.Handler;
 import org.apache.velocity.app.VelocityEngine;
 import org.georchestra.commons.configuration.GeorchestraConfiguration;
-import org.jfree.util.Log;
```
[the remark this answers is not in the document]
RemiDesgrange: ? I suppressed the dependcies to jfreecharts. There where a typo in the code that use `Log` class of jfreecharts instead of log object.

atlas/src/main/resources/camel/atlas_app.xml
RemiDesgrange: IIRC, you want me to rename the id to `org.georchestra.atlas.camel.logger` ?

atlas/src/main/java/org/georchestra/atlas/AtlasMailComponent.java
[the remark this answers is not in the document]
pmauduit: I read the diff in the opposite way, sorry.

atlas/src/main/resources/camel/atlas_app.xml
pmauduit: something like that, instead of georLogger

Issue #2588, opened 2019-04-08
BasicResourcePool - Bad pool size config


fvanderbiest:
2019-04-08 09:37:43,230 [main] WARN  com.mchange.v2.resourcepool.BasicResourcePool - Bad pool size config, start 1 < min 2. Using 2 as start.

severo:
See https://github.com/georchestra/georchestra/pull/2590 and https://github.com/georchestra/datadir/pull/158.

PR #2590, opened 2019-04-08
All - min/initial postgresql pool size from 1 to 2

severo:

PR #2591, opened 2019-04-09
All - min/initial postgresql pool size from 1 to 2

severo:


severo:
Replaces https://github.com/georchestra/georchestra/pull/2590. Fixes #2588. See also georchestra/datadir#159

PR #2597, opened 2019-04-12
Make properties loading in jsp cleaner

RemiDesgrange:
Following the recent PR #2578 and #2594 adding a little more config
to make it cleaner to load properties in jsp.


severo:
Nice!

RemiDesgrange:
need stash then ok :-)

landryb:
I assume it's been tested at runtime ? ;)

Issue #2231, opened 2018-11-12
epsg-extension - Execution default-test of goal org.apache.maven.plugins:maven-surefire-plugin:2.12.4:test failed


fvanderbiest:
mvn -P-all,mapfishapp install fails on "Custom Extensions to EPSG authority factory" 

[ERROR] Failed to execute goal org.apache.maven.plugins:maven-surefire-plugin:2.12.4:test (default-test) on project epsg-extension: Execution default-test of goal org.apache.maven.plugins:maven-surefire-plugin:2.12.4:test failed: The forked VM terminated without saying properly goodbye. VM crash or System.exit called ? -> [Help 1]

```
$ mvn --version
Apache Maven 3.5.2
Maven home: /usr/share/maven
Java version: 1.8.0_181, vendor: Oracle Corporation
Java home: /usr/lib/jvm/java-8-openjdk-amd64/jre
Default locale: fr_FR, platform encoding: UTF-8
OS name: "linux", version: "4.15.0-36-generic", arch: "amd64", family: "unix"
```

fvanderbiest:
-DskipTests is a workaround but I'm surprised...

pmauduit:
Note that you will have to skip tests everywhere, the debian bug pointed above proposes to add an option to the maven-surefire-plugin (`<useSystemClassLoader>false</useSystemClassLoader>`)

landryb:
Issue still present as of now, and annoying, as it breaks the buildbot ci using stretch - since it uses the Makefile for epsg-extensions, and the Makefile doesnt set -DskipTests. I had to downgrade jdk...

groldan:
Superseded by #2600, should be closed as won't fix

PR #2600, opened 2019-04-12
mapfishapp/extractorapp: drop unused epsg-extension module

### LINKED ISSUE #2453: Drop epsg_extension in favor of vanilla geoserver user projections support?
It seems that the `epsg-extension` module was added back in [May 2011](https://github.com/georchestra/georchestra/blame/18.06/epsg-extension/src/main/java/org/geotools/referencing/factory/epsg/CustomCodes.java) right before GeoServer added support for custom CRS definitions in `<data dir>/user_projections/epsg.properties` in [June](https://github.com/geoserver/geoserver/blame/44bacafcf1352bfe0b1836262df686fd58036585/src/main/src/main/java/org/vfny/geoserver/crs/GeoserverCustomWKTFactory.java) of the same year.

Neither have changed over the years, so I wonder if we shouldn't just get rid of the custom `epsg_extension` module in favor of vanilla geoserver support for the same functionality.

Only thing to consider would be upgrading the config of anyone using the `-DCUSTOM_EPSG_FILE` system property by `-Duser.projections.file` or just saving the custom projections in `<data dir>/user_projections/epsg.properties`.

@fvanderbiest @pmauduit comments?

groldan:
Fixes #2453


groldan:
Fixes #2231

fvanderbiest:
Thanks !

PR #2601, opened 2019-04-16
send confirmation mail to user on account validation (Fixes #2598)

### LINKED ISSUE #2598: [console] send confirmation email upon account validation
when moderatedSignup is true, it'd be nice to still send a confirmation email to the user so that he knows his account has been validated. Might be a new template, different from newaccount-was-created-template.txt or could reuse it.

landryb:
The logic could also go into https://github.com/georchestra/georchestra/blob/master/console/src/main/java/org/georchestra/console/ds/AccountDaoImpl.java#L211 - feedback welcome.

If merged, will do a PR on the datadir to amend https://github.com/georchestra/datadir/blob/19.04/console/templates/README.md#console-email-templates to reflect the new behaviour & https://github.com/georchestra/datadir/blob/19.04/console/templates/newaccount-requires-moderation-template.txt#L5 (add email - for the latter, the logic is that the user email might give a clue what org just registered..)

Tested behaving fine on 19.04.

### DIFF
--- a/console/src/main/java/org/georchestra/console/mailservice/EmailFactory.java
+++ b/console/src/main/java/org/georchestra/console/mailservice/EmailFactory.java
@@ -144,6 +144,7 @@ public class EmailFactory {
 				this.instanceName);
 		email.set("name", userName);
 		email.set("uid", uid);
+		email.set("email", userEmail);
 		email.send();
 	}
 
--- a/console/src/main/java/org/georchestra/console/mailservice/EmailFactory.java
+++ b/console/src/main/java/org/georchestra/console/mailservice/EmailFactory.java
@@ -93,7 +93,7 @@ public class EmailFactory {
 	}
 
 	/**
-	 * e-mail to the user to inform the account requires the moderator's singnup
+	 * e-mail to the user to inform the account requires the moderator's validation
 	 */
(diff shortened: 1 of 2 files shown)


fvanderbiest:
Sounds good.
One might question the re-use of https://github.com/georchestra/datadir/blob/19.04/console/templates/newaccount-was-created-template.txt in this context, but why not.
I think I would have prefered a different email template & method, saying "Your account on {publicUrl}/ has been validated !" rather than "Your account on {publicUrl}/ has been successfully created !"

landryb:
In most cases, ppl rewrite those templates anyway... :)

fvanderbiest:
It's not about rewriting it but having a different one, because, IIUC your proposal, the same template will be used in two different cases (moderated signup & unmoderated signup), which might be misleading.

landryb:
to me it's just a matter of clearly explaining it in https://github.com/georchestra/datadir/blob/19.04/console/templates/README.md#console-email-templates

fvanderbiest:
> to me it's just a matter of clearly explaining it in https://github.com/georchestra/datadir/blob/19.04/console/templates/README.md#console-email-templates

OK, please submit a PR on datadir now that it's merged.

Issue #2605, opened 2019-04-19
geoserver: upgrade to latest stable version 2.15.0


groldan:
We might want to upgrade to the latest geoserver before the version we're using (2.12.5) gets too old, since it's actually EOL already.

Also, 2.15.0 is the first one that runs on Java 11 which opens the door to migrate to a docker container that supports cgroups (i.e. respecting CPU and memory limits).

RemiDesgrange:
This is a great idea.Can we list all the benefits of migration in PR ?

fvanderbiest:
Was done with #2606

PR #2606, opened 2019-04-19
Upgrade to Geoserver 2.15.0

groldan:
Upgrades `geoserver/webapp` to 2.15.0 as of #2605.
Note: I don't have write access to georchestra/geoserver to try this out you'll need to manually fetch the following branch:
```
cd geoserver/geoserver-submodule
git remote add groldan git@github.com:groldan/geoserver.git
git fetch groldan
git reset groldan/2.15.0-georchestra --hard
```


RemiDesgrange:
Thank for this. As you are pointing in #2605 there are many stuff improved in geoserver 2.15 :
* java11 support that imply
  * better container support (jvm is aware of cgroups and namespaces)
  * availability of ZGC, which can lead to significant memory reduction (not sure, but we can try)
 * JAI-EXT by default
* ...

geoserver/webapp/pom.xml
```diff
@@ -169,42 +169,6 @@
...
-      <artifactId>spring-tx</artifactId>
-      <version>${gs.spring.version}</version>
-    </dependency>
-    <!-- xml-apis is needed by gt-svg in v1.3.04, but the GS webapp upstream
-         uses v1.4.01 -->
-    <dependency>
-      <groupId>xml-apis</groupId>
-      <artifactId>xml-apis</artifactId>
-      <version>1.4.01</version>
-    </dependency>
```
pmauduit: Why removing the dependencies ? Are they now implicit by using the spring boot parent pom ?

geoserver/lib/2.15.0/gt-s3-geotiff-21.0.jar.readme
```diff
@@ -0,0 +1,8 @@
+This is gt-s3-getotiff-21-SNAPHOT meant to
+replace gt-s3-geotiff-21.0.jar in 
+geoserver/WEB-INF/lib, with fixes
+that didn't make it to the official 21.0 
+GeoTools release from Tue Feb 2019, whilst
+the patch is from Mon Mar 2019 at commit
+c73b886b1, and hence will be available in
+the next 21.1 GeoTools release.
```
pmauduit: I'm not very fond of adding binaries to git, but It's fine. Do you confirm that we could remove them later on ? (once GS will have upgraded the GT version, and we will stick with the new one) ?

fvanderbiest:
Hi Gabriel, what's the status of this PR ? Is it draft or ready ?

> I don't have write access to georchestra/geoserver

Weird, since you're a commiter ...

geoserver/lib/2.15.0/gt-s3-geotiff-21.0.jar.readme
groldan: yeah, we just need to upgrade to a version that comes with the patched s3-geotiff.jar. Which, as of today, should be 15.1, I'll give it a try without the extra binaries and update the PR

PR #2612, opened 2019-05-06
SP - let the possibility to have a "moved *" HTTP response + a content

pmauduit:
The motivation behind this PR is that if a 301 || 302 redirect is sent by an downstream server with a content, the content is not copied (because of the "return" statement), but the header "Content-Length" is. This triggers an actual 500 error from the Jetty hosting the security-proxy with the message "insufficient content written".

Only returning in case of unable to extract the Location header (sendError()), so that we let the possibility to write the content of the downstream http response.

Tests: runtime only as the modified code is pretty hard to instrument :(


fvanderbiest:
Do you think it makes sense having this patch in 19.04 too ?

pmauduit:
yes it does, sorry for not having targetted the right branche

fvanderbiest:
OK: 21201ec1a

PR #2614, opened 2019-05-07
SP - let the possibility to have a "moved *" HTTP response + a content

pmauduit:
Same as PR 2612, but on 19.04


fvanderbiest:
I cherry-picked the commit in between

PR #2624, opened 2019-05-17
changing make file to reflect git branch on building

RemiDesgrange:
Adding this maven option will also work with fixed version (ex: 19.04,
19.06 etc..)

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -20,36 +20,36 @@ docker-build-gn3: docker-pull-jetty
 	cd geonetwork; \
 	mvn -DskipTests clean install; \
 	cd web; \
-	mvn -P docker -DskipTests package docker:build -DdockerImageTags=latest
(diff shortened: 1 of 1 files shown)


RemiDesgrange:
Yess that was my intitial thought :-)

pmauduit:
Did you try it ? since I cannot find any occurence of `dockerImageTags` in the pom.xml, I wonder how it will substitute:
https://github.com/georchestra/georchestra/blob/master/mapfishapp/pom.xml#L544

I'd have put a dockerImageTag property which by default would have taken project.version, if you want to keep the possibility to override it (anyway, it's still overrideable using -DdockerImageName and fully giving a custom image name + tag).

RemiDesgrange:
Yup I tested it ;-)

pmauduit:
Ok, I was not aware of the possibility by the spotify docker maven plugin to override the tag this way, sorry about that.

Issue #2633, opened 2019-05-29
Missing gs-printing maven artifact


jusabatier:
I tried to build geoserver  (georchestra 19.04) with the printing module, but I get an error : 

Downloading from artifactory-georchestra: https://packages.georchestra.org/artifactory/maven/org/geoserver/community/gs-printing/2.12.5/gs-pri... 
The POM for org.geoserver.community:gs-printing:jar:2.12.5 is missing, no dependency information available.

Is it possible to add it on the repo ?

jusabatier:
It appear that gs-printing is no more in org.geoserver.community but moved to org.geoserver.extension

So it have to be updated in the webapp's pom.xml

PR #2634, opened 2019-05-29
gs-printing move to org.geoserver.extension

jusabatier:
Solve https://github.com/georchestra/georchestra/issues/2633

The gs-printing module moved from community to extension namespace.


pmauduit:
gs-printing v2.12 seems available on the georchestra's artifactory (and upstream on boundlessgeo maven repo), so it should not harm merging it. thanks

PR #2636, opened 2019-06-14
security-proxy: Allow semicolons in URLs

groldan:
Configure org.springframework.security.web.firewall.StrictHttpFirewall"
with allowSemicolon = true to avoid errors
like 'the request was rejected because the URL contained a potentially malicious String ";"'

";" in URL's come in the form http://<domain>/<path>;jsessionid=xxxx
In general, proxied applications are discouraged to disclose jsessionid this way,
and shall use a cookie instead.
Some applications like CKAN though can't be modified to use cookies, hence this configuration.

### DIFF
--- a/security-proxy/src/main/webapp/WEB-INF/applicationContext-security.xml
+++ b/security-proxy/src/main/webapp/WEB-INF/applicationContext-security.xml
@@ -16,14 +16,14 @@
 
 	<bean id="allowSemicolonHttpFirewall" class="org.springframework.security.web.firewall.StrictHttpFirewall">
 		<!-- 
-			Set allowSemicolon = true to avoid errors 
+			Allows configuring the allowSemicolon StrictHttpFirewall property to avoid errors 
 		    like 'the request was rejected because the URL contained a potentially malicious String ";"'
 		    ";" in URL's come in the form http://<domain>/<path>;jsessionid=xxxx OR in static resource 
 		    URLs (e.g to reference bundles of JS files (minified on the fly)). 
 		    In general, proxied applications are discouraged to disclose jsessionid this way, and shall use a cookie instead.
 		    Some applications like though can't be modified to use cookies, hence this configuration.
 		 --> 
-	  	<property name="allowSemicolon" value="true"/> 
+	  	<property name="allowSemicolon" value="${allowSemicolon:false}"/> 
 	</bean> 
 	<s:http-firewall ref="allowSemicolonHttpFirewall"/>
 
--- a/security-proxy/src/main/webapp/WEB-INF/applicationContext-security.xml
+++ b/security-proxy/src/main/webapp/WEB-INF/applicationContext-security.xml
@@ -14,6 +14,19 @@
     <context:property-placeholder location="file:${georchestra.datadir}/default.properties, file:${georchestra.datadir}/security-proxy/security-proxy.properties"
         ignore-resource-not-found="true" ignore-unresolvable="true" />
 
+	<bean id="allowSemicolonHttpFirewall" class="org.springframework.security.web.firewall.StrictHttpFirewall">
+		<!-- 
+			Set allowSemicolon = true to avoid errors 
+		    like 'the request was rejected because the URL contained a potentially malicious String ";"'
+		    ";" in URL's come in the form http://<domain>/<path>;jsessionid=xxxx OR in static resource 
+		    URLs (e.g to reference bundles of JS files (minified on the fly)). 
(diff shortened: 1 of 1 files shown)


security-proxy/src/main/webapp/WEB-INF/applicationContext-security.xml
```diff
@@ -14,6 +14,19 @@
     <context:property-placeholder location="file:${georchestra.datadir}/default.properties, file:${georchestra.datadir}/security-proxy/security-proxy.properties"
         ignore-resource-not-found="true" ignore-unresolvable="true" />
 
+	<bean id="allowSemicolonHttpFirewall" class="org.springframework.security.web.firewall.StrictHttpFirewall">
+		<!-- 
+			Set allowSemicolon = true to avoid errors 
+		    like 'the request was rejected because the URL contained a potentially malicious String ";"'
+		    See <https://jira.camptocamp.com/browse/GSHDF-95>
```
fvanderbiest: This URL is private. IMO we should remove it.

security-proxy/src/main/webapp/WEB-INF/applicationContext-security.xml
```diff
@@ -14,6 +14,19 @@
...
         ignore-resource-not-found="true" ignore-unresolvable="true" />
 
+	<bean id="allowSemicolonHttpFirewall" class="org.springframework.security.web.firewall.StrictHttpFirewall">
+		<!-- 
+			Set allowSemicolon = true to avoid errors 
+		    like 'the request was rejected because the URL contained a potentially malicious String ";"'
+		    See <https://jira.camptocamp.com/browse/GSHDF-95> 
+		    ";" in URL's come in the form http://<domain>/<path>;jsessionid=xxxx
+		    In general, proxied applications are discouraged to disclose jsessionid this way, and shall use a cookie instead.
+		    Some applications like CKAN though can't be modified to use cookies, hence this configuration.
```
fvanderbiest: CKAN does not use semicolons to disclose jsessionid, but to reference bundles of JS files (minified on the fly)

security-proxy/src/main/webapp/WEB-INF/applicationContext-security.xml
```diff
@@ -14,6 +14,19 @@
...
+		 --> 
+	  	<property name="allowSemicolon" value="true"/>
```
fvanderbiest: I would find it useful to have this setting configurable from the datadir, with a default value set to `false`.
Platforms using CKAN would have to switch it to true.

groldan: sounds doable, let me see

groldan: See 26cfc9a6b064828a8eba1dca4fdc54727e741aff and https://github.com/georchestra/datadir/pull/162

fvanderbiest:
Excellent, thanks !

pmauduit:
This should also be ported back to 19.04

landryb:
Would it be possible to also allow `//` in urls too while here ? this is a regression from 18.06..

pmauduit:
> Would it be possible to also allow `//` in urls too while here ? this is a regression from 18.06..

where have you encountered such urls ?

landryb:
https://github.com/georchestra/cadastrapp/commit/db5b00871c555efba978a90f60f3944ef38c2e98 & https://github.com/georchestra/cadastrapp/commit/42a01a13b4fea09c0550fc6788a0014d36d1181c from https://github.com/georchestra/cadastrapp/issues/445 among others, but this can cause issues for generated links in geoserver or geonetwork (ie interlinks between md & layer, when the url is generated by some code -> this sometimes has `//`)

I know i had to fix/remove some slashes here and there in the geor datadir to account for this change

pmauduit:
Ok, it's in the path, not in the arguments, so it would be somewhere else

landryb:
> > Would it be possible to also allow `//` in urls too while here ? this is a regression from 18.06..
> 
> where have you encountered such urls ?

i was having broken extent thumbnails on a metadata view, because the url generated by geonetwork looked like https://ids.craig.fr/geocat/srv//fre/region.getmap.png?... -> double-slash triggered a 500 code with the sec-proxy firewall.

Looking at upstream repo and after digging a good while to find out where the html was generated, if we want to fix it on 19.04 geonetwork/core-geonetwork@05836b4777cd213096576875233ed838f9d9969d should be cherry-picked on top of `georchestra-gn3.4-19.04` branch. Yes, i'm still fighting with this horror...

PR #2642, opened 2019-06-24
Remove X-Frame-Options http header

RemiDesgrange:
This will permit geochestra app to be embed in other app.

Note that X-Frame-Options is _not_ removed from cas, which we
explicitely don't to append.


fvanderbiest:
useless PR, since the fix has already been applied to oldest supported branch (18.06)

PR #2652, opened 2019-07-15
[console] Allow multiselection of users

tonio:
see https://jira.camptocamp.com/browse/GSHDF-29

### DIFF
--- a/console/src/main/webapp/manager/app/assets/lang/de.json
+++ b/console/src/main/webapp/manager/app/assets/lang/de.json
@@ -192,6 +192,10 @@
   "users.email"           : "E-Mail",
   "users.pending"         : "***TODO***",
   "users.pending_desc"    : "***TODO***",
+  "users.filter_selected" : "**** TODO / translate ****",
+
+  "sel.all"               : "**** TODO / translate ****",
+  "sel.none"              : "**** TODO / translate ****",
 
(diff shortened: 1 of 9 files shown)


fvanderbiest:
@tonio conflicts to fix

Would be nice if @Gaetanbrl could have a look too.

Gaetanbrl:
How work filter and selection together ? 

If i use filter and click on "select all", i have all users in my selection.
image

That could be be very powerfull to use filter and multiselection together in this particular case and others.

Gaetanbrl:
Additionally, if you select 2 users with manual selection, write something to filter users and check "display selection only"  you have no result.

image

tonio:
> That could be be very powerfull to use filter and multiselection together in this particular case and others.

Good catch, will do.

> Additionally, if you select 2 users with manual selection, write something to filter users and check "display selection only" you have no result.

This one looks correct to me. You're displaying selection, and filter into it…

Gaetanbrl:
> You're displaying selection, and filter into it…
To understand, could you explain me what is the selection display by this checkbox ?
image

tonio:
You display only selected(checked) users.

Gaetanbrl:
Thanks, i was confused because in this case i see the number of selected users, my filter selection (independant of manual selection) and nothing if i check "display only selection" checkbox.

PR #2655, opened 2019-07-24
Support geojson in mapfishapp

RemiDesgrange:
@groldan or @fvanderbiest If you look at the test mapfishapp/src/test/java/org/georchestra/mapfishapp/ws/upload/GeotoolsFeatureReaderTest.java#L450 and you launch the test, you'll see it fail. But it work for example with EPSG:2154. I have no Idea :-(


fvanderbiest:
It should target `master`, it's a feature.

Probably also misses a better comment in https://github.com/georchestra/georchestra/blob/master/postgresql/020-mapfishapp.sql#L8

@groldan to the rescue for the tests ?

Issue #2278, opened 2018-11-23
[console] improve/fix user export as csv/vcf


landryb:
Somewhat of a hidden feature it seems, one can export a list of uids as VCF/CSV with https://github.com/georchestra/georchestra/blob/master/console/src/main/java/org/georchestra/console/ws/backoffice/users/UsersExport.java#L76 . Except i struggled a bit to make it work, as the code declares it's a POST, but fails if you just use POST with json data:
```
curl -v -H "Content-Type: application/json" -X POST -d '{"users":["testadmin"]}' -u testadmin:testadmin https://georchestra.dev.craig.fr/console/private/users.csv

GlobalExceptionHandler [ERROR] org.springframework.web.bind.MissingServletRequestParameterException: Required String parameter 'users' is not present
```

i figured out one needs to pass the users arg urlencoded as a GET queryParam to make it work but still specify it's a POST (while the arg is not taken from the postdata), which is somewhat awkward:
```
curl -XPOST -v -H "Content-Type: application/json" -d '{"users":["testadmin"]}' -u testadmin:testadmin https://georchestra.dev.craig.fr/console/private/users.csv?users

org.json.JSONException: A JSONArray text must start with '[' at character 0
```
the version that works:
```
curl -H "Content-Type: application/json" -XPOST -u testadmin:testadmin https://georchestra.dev.craig.fr/console/private/users.csv?users=%5Btestadmin%5D
```
can the code be fixed so that it properly takes the params from the POST data ? otherwise the query should be declared as GET..

On top of this, two ideas:
  * adding a link from the user page on the console allowing to export the given user as CSV/VCF (i suppose that was one of the original intents of the given code ?)
  * if no param is given, export the complete list...

And, while at it, could be nice to have the same for the orgs.. but that's probably ways more code :)

landryb:
@fvanderbiest @tonio: API-related, should the call be a GET or a POST ? To me it seems more natural to use a GET, unless the intent is to send a huge list of usernames in the queryparams, where in that case POST makes sense..

Or have GET take no params and export the complete list, and POST only taking usernames as param ?

Gaetanbrl:
Is this two link good for you ? 

![image](https://user-images.githubusercontent.com/16317988/61796086-04427d00-ae25-11e9-97ba-f8a830b82e52.png)

fvanderbiest:
[...]

PR #2660, opened 2019-07-25
Add links to export all/filtered users to csv/vcf with POST

Gaetanbrl:
Issue #2278

* Only use POST request without URL params.

This PR seems to be in redundancy with [the Geo2france backlog](https://github.com/orgs/georchestra/projects/7#card-15166711). 
So, you could close or approve this work according to the upcoming progress.

_Ask manual reviewers (no reviewers panel available) @pierrejego @landryb_

### DIFF
--- a/console/src/main/java/org/georchestra/console/ws/backoffice/users/UsersExport.java
+++ b/console/src/main/java/org/georchestra/console/ws/backoffice/users/UsersExport.java
@@ -32,9 +32,9 @@ import org.springframework.security.access.AccessDeniedException;
 import org.springframework.security.core.Authentication;
 import org.springframework.security.core.context.SecurityContextHolder;
 import org.springframework.stereotype.Controller;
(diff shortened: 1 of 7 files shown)


landryb:
dunno which pr should be rebased on top of which but what's in this one works fine.

fvanderbiest:
> IMHO this should be rebased on top of #2652, use ids from the selection & implement action in the selection dropdown.

Agreed: having a manual selection of users seems a more powerful approach.
Closing this one.

fvanderbiest:
Sorry Gaetan. Hope you don't mind.

Gaetanbrl:
> Agreed: having a manual selection of users seems a more powerful approach

No problem @fvanderbiest.
But for the bothe are powerfull. The filter field help users to select a large number of users that could be more slow with checkbox selection (click by click). 
Manual selection is really usefull if you want to select only some users (for exemple, inside a filter's results or inside role members).

PR #2670, opened 2019-08-01
Correct backend to serve csv and vcf.

RemiDesgrange:
Only VCF work ATM.

Also correct ldapScheme.

@groldan @cmangeat Can you try CSV export and tell me what piece I'm missing here ?

I also wasn't able to make the JSON parsing via spring. According to "the internet" having :
```java
public String getUsersAsCsv(@RequestParam List<String> rawUsers) 
```
Should be possible but I cannot make it working.


cmangeat:
For csv, seems there are no need for converters, just changing produce into:
```produces = "text/comma-separated-values;charset=utf-8"```.

(I think that when url ends with ".csv", a resolution is attempted and lead to this).

PR #2675, opened 2019-08-14
Change the api endpoint to avoid clashing with url in spring

RemiDesgrange:
previous endpoint `private/users.csv` was fragile. I could clash with [this endpoint](https://github.com/georchestra/georchestra/blob/master/console/src/main/java/org/georchestra/console/ws/backoffice/users/UsersController.java#L213)

### DIFF
--- a/console/src/main/java/org/georchestra/console/ws/backoffice/users/UsersExport.java
+++ b/console/src/main/java/org/georchestra/console/ws/backoffice/users/UsersExport.java
@@ -97,7 +97,7 @@ public class UsersExport {
     }
 
     @PostMapping(value = "/private/export/users.csv", consumes = MediaType.APPLICATION_JSON_VALUE,
-            produces = "text/comma-separated-values; charset=utf-8")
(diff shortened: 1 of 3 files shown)


fvanderbiest:
Can you explain the role of the ContentNegotiationManager here ?
Does it fix the export ? Thanks.

RemiDesgrange:
Default behavior of spring MVC is to infer the "Content-Type" of a response, based on the extension in the URL. 
Ex: 
* csv
* pdf
* etc...

Spring was trying to infer CSV auto-magically, but it doesn't work for us, so we are telling spring to _not_ try to guess anything and let us do the job manually.

fvanderbiest:
Was it because of `text/comma-separated-values` ?
Did you also test exporting `text/csv; charset=utf-8` without contentNegotiationManager ?

RemiDesgrange:
Yes, we tested without `contentNegotiationManager`and it was working, but not behind the SP (or any other proxy)

PR #2676, opened 2019-08-15
Upgrade geotools to 21.2 and remove the geotools submodule

groldan:
## Upgrade to GeoTools 21.2 (latest release)

* Main geotools version upgraded from 19.4 (latest with JTS from vividsolutions) to the latest release. Changes mostly related to refactor `com.vividsolutions.jts` to `org.locationtech.jts`.
* geowebcache-webapp: declares `<gt.version>14.1</gt.version>` to match the geotools version used by geowebcache `1.8.1` and avoid possible runtime clashes.
* extractorapp:  removes support for exporting to MIF/TAB formats, since they're not supported in GeoTools anymore.
* mapfishapp: migration without issues

## Remove GeoTools submodule:
    
The GeoTools submodule was stuck at version 9.2 for extractorapp to support exporting to MapInfo MIF file format. With the upgrade of extractorapp and the removal of support for MIF, the geotools submodule is no longer needed.

PR #2677, opened 2019-08-16
mapfishapp: Fix support for uploading KML files and convert to GeoJSON for display

groldan:
This PR depends on #2676 and the following GeoTools PR to be backported to 21.x: geotools/geotools/pull/2523

Fixes several issues with KML upload to mapfishapp and getting a GeoJSON response:

### Response content-type: 
The API returns application/json as response content type, except if the request explicitly asks for text/html, which is the case for the Ext.js submitted form when uploading a file. In this case text/html is returned to avoid an error/warning on the Javascript client side.
    
### KML file parsing improvements (KmlFeatureSource):
* Parsing to FeatureCollection made streaming, it was loading all the Features in memory.
* Added on-the fly reprojection through FeatureCollection decorator
* Force the default geometry attribute to be the "Geometry" attribute, the GeoTools parser returns Features whose schema's default geometry attribute is "LookAt", since it occurs before "Geometry",
and results in no geometry being encoded to GeoJSON, which takes the default geometry as the GeoJSON "geometry" property.
* Remove the "Style" attribute through a feature collection decorator, the GeoTools parser sets it to a FeatureTypeImpl instance which gets encoded to GeoJSON as FetureTypeImpl.toString(), which makes no sense.

Several other fixes and improvements, check the PR's commit list.


groldan:
GeoTools backport PR: geotools/geotools/pull/2523. Once travis finishes I'll merge it, then once Jenkins publishes the artifacts this PR should build fine.

fvanderbiest:
Excellent !
Thanks a bunch Gabriel.

Can you rebase on top of master, so we don't ship commits from #2676 ?

groldan:
Yep, that's the idea, I'll do, thanks for the review

groldan:
@fvanderbiest rebased.

PR #2694, opened 2019-08-23
Automated Java code formatting and verification

groldan:
### Apply formatter-maven-plugin formatting to all projects.

Add the `net.revelc.code.formatter:formatter-maven-plugin` and configure it to be inherited by georchestra projects (no geoserver/geonetwork submodules).

The plugin is configured to format only java source code files using the `.mvn/formatter.xml` eclipse format definition. 
This file can be directly imported from eclipse IDE and from IntelliJ IDEA's [Eclipse Code Formatter](https://plugins.jetbrains.com/plugin/6546-eclipse-code-formatter) plugin.

When compiling a project that includes the formatter-maven-plugin, the default action is to format the source code to adhere to the defined code style.

The other option is to instruct the plugin to check for correct formatting and fail the build.
    
This behavior is controlled by the fmt.action property, like in `mvn compile -Dfmt.action=format` or `-Dfmt.action=validate`. By default a project build will run the `format` action, so nothing special needs to be done in order to ensure the proper format of the code other than making sure to run the build before committing/pushing to github.

On a CI build server (Jenkins/Travis-CI), the `-Dfmt.action=validate` should be used, to ensure a committer did his/her due diligence before pushing to the upstream git repository, which is already the case for Travis-CI.
    
Finally, the plugin can be disabled with the `-Dfmt.skip=true` maven property.


landryb:
Not a big fan of big PRs reformatting code, as it often invalidates/creates useless conflicts with all pending PRs. Oh well.. has to be done someday, i guess.

pmauduit:
> (maybe squash when merge)

I think it will be better to keep the reformatting with no changes in a separate commit (for history and understanding purposes).

fvanderbiest:
Thanks Gabriel, awesome.

Issue #2685, opened 2019-08-21
[Console] filtering by org in analytics doesn't work


RemiDesgrange:
When you go to the analytics tab in the console, filtering by role doesn't work.

maybe add a new line over here : https://github.com/georchestra/georchestra/blob/19.04/analytics/src/main/java/org/georchestra/analytics/StatisticsController.java#L313

fvanderbiest:
```
curl 'https://username:password@www.geo2france.fr/analytics/ws/layersUsage.json' \
-H 'Content-Type: application/json;charset=UTF-8' \
--data-binary '{"service":"layersUsage.json","startDate":"2019-07-23","endDate":"2019-08-23","role":"GT_TELECOM","limit":10}'
```
produces indeed the same output as
```
curl 'https://username:password@www.geo2france.fr/analytics/ws/layersUsage.json' \
-H 'Content-Type: application/json;charset=UTF-8' \
--data-binary '{"service":"layersUsage.json","startDate":"2019-07-23","endDate":"2019-08-23","limit":10}'
```

BTW, why do we use POST to query this service ? Would be more logical to use GET instead ...

fvanderbiest:
`18.06`, `19.04` and `master` are affected.

RemiDesgrange:
maybe because of the data passed. But these data could be passed as query string easily

PR #2698, opened 2019-08-29
analytics - fixing filtering by roles

pmauduit:
See #2685 ;

Note: this PR introduces some integration tests on the analytics webapp, inspired from the ones introduced in 19.04 for the console, but:

* Integration tests follow the same conventions as the regular default maven failsafe plugin (code into src/test instead of src/it)
* The introduced IT makes use of the exact same spring beans configuration as the one used in the webapp in production (ws-servlet.xml)
* We still have to activate the "it" profile to make use of the postgresql docker image.

Tests:

* `mvn test` ok
* `mvn verify -Pit` ok

Credits goes to @groldan for the help !

### DIFF
--- a/analytics/pom.xml
+++ b/analytics/pom.xml
@@ -126,7 +126,7 @@
     <dependency>
       <groupId>junit</groupId>
       <artifactId>junit</artifactId>
-      <version>4.8</version>
+      <version>4.9</version>
       <scope>test</scope>
     </dependency>
     <dependency>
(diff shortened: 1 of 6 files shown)


analytics/pom.xml
```diff
@@ -1,599 +1,704 @@
 <?xml version="1.0"?>
```
pmauduit: I did not expect such a reindentation on the pom, sorry about that. Maybe we shall stick with 2 spaces on xml files ?

fvanderbiest:
When merged, this will require to merge 18.06 into 19.04 into master ...

groldan:
> When merged, this will require to merge 18.06 into 19.04 into master ...

The correct command would then be:
```
git checkout 19.04
git merge --strategy-option ours 18.06
``
You'll get a number of files deleted in 19.04, so do:
```
git mergetool
```
And select the `d` option when prompted for what to do with each deleted file, to keep'em deleted
now `git merge --continue` and voilà

**Comment by pmauduit** [comment]
> It'd be good if you can squash all commits in the PR into a single one and provide a descriptive commit message of what it encompasses functionality wise.

Done

File: analytics/src/test/java/org/georchestra/analytics/StatisticsControllerIT.java [hunk 2]
```diff
@@ -0,0 +1,84 @@
...
+    private @Value("${dlJdbcUrlOGC}") String dlJdbcUrlOGC;
+
+    private @Autowired BasicDataSource ds;
+
+    protected final Log logger = LogFactory.getLog(getClass().getPackage().getName());
+
+    private static boolean dataLoaded = false;
+
+    public @Before void before() throws Exception {
+        if (!dataLoaded) {
```
pmauduit: I think @groldan provided a better approach while working on GPDR:

https://github.com/georchestra/georchestra/pull/2613/files#diff-0cdf168c2c66425c514a83589eeddd9dR62

The annotation comes from com.github.database.rider and should do the work better than I can do in this before() method

fvanderbiest: IIUC, you do not consider this one is ready for merge ?

pmauduit: I was wondering if it was worth to integrate this database-rider dependency (on a test scope) or not. In the current state, it works, it would be fancier with it.

pmauduit: I did some rapid tests with database-rider (the `DataSet` annotation), as well as the `Sql` from spring-tests, but with no luck, I cannot get them work to load my fixtures.

pmauduit: trying harder I succeeded \o/ I'll update my PR actually

fvanderbiest:
> > When merged, this will require to merge 18.06 into 19.04 into master ...

Thanks Gabriel. I successfully merged 18.06 into 19.04.
I gave up merging 19.04 into master, there are too many conflicts that I was not able to resolve.
Asking for help here...

PR #2699, opened 2019-08-30
GDPR - UI

tonio:
Fixes [GSHDF-155](https://jira.camptocamp.com/browse/GSHDF-155) & [GSHDF-214](https://jira.camptocamp.com/browse/GSHDF-214)

Screenshot from 2019-08-30 12-18-24


console/src/main/webapp/account/css/console.css
```diff
@@ -26,3 +26,6 @@
 .gdpr {
   margin-top: 2em;
 }
+.org-logo {
+  max-width: 100%;
```
fvanderbiest: what about adding a `max-height` in case the image is flipped ?

PR #2702, opened 2019-09-05
mapfishapp - boolean config option to prevent layer download

fvanderbiest:
SDIs expect their users to use services rather than static files ...


fvanderbiest:
See also https://github.com/georchestra/datadir/commit/644d1dcc432d8f3c119a48ea3b2d2522256ca434

Issue #2450, opened 2019-02-04
[console] display role description when hovering on the role labels


landryb:
Nice to have and probably easy to do - on the user page, in the role tab, the 'raw' list of georchestra & other roles are displayed - it would be great to display the role description LDAP field when hovering over the role name so that a user admin knows that role 'FOO' means 'gives access to FOO, BAR and BAZ depending on BLAH' - otherwise you get to create a super long role name which needs to carry all this knowledge..

Would also be nice on the role/favorites list displayed in the left pane on the user list.

fvanderbiest:
Agreed, interesting evolution.

Gaetanbrl:
Currently, some descriptions are translate in [lang files](https://github.com/georchestra/georchestra/blob/19.04/console/src/main/webapp/manager/app/assets/lang/fr.json#L10).

- What about the others description read in english from the response as JSON ? 

- And what do you thing about this simple tooltip design ? 
![image](https://user-images.githubusercontent.com/16317988/61472120-1be0b800-a984-11e9-8992-24b8bebd4ecc.png)

landryb:
ok for the tooltip design - and for the tooltip source, it should just be the description field in the json response. The value should come from ldap in all cases, no ? I find it strange that the description field cannot be modified for the georchestra roles (GN*, etc) - @fvanderbiest was it on purpose (i see it was done in 6370fda846) ? I can understand not being able to change the role name, but the description..

The ones coming from the translation files are only used when hovering the 'administrator/delegated administrator' badge on the upper right part of the console. It's not related to a role name.

Gaetanbrl:
> The value should come from ldap in all cases, no ?
I tend to say yes, from the JSON file.

Gaetanbrl:
@landryb could we validate this modification to realize PR and open new issue about translation / LDAP field ?

landryb:
there's no issue with translations afaict, the value to display comes from LDAP, that's all.

A separate issue would be making the description modifiable for the privileged roles.

landryb:
closed by #2703

PR #2703, opened 2019-09-09
display role description when hovering on the role labels

Gaetanbrl:
Link to #2450
Ask manual reviewers (no reviewers panel) @pierrejego @landryb

### DIFF
--- a/console/src/main/webapp/manager/app/components/user/user.es6
+++ b/console/src/main/webapp/manager/app/components/user/user.es6
@@ -66,6 +66,18 @@ class UserController {
     return (this.delegation.orgs.length !== 0) && (this.delegation.roles.length !== 0)
   }
 
+  // search each choosen span elements and set title manually
+  // to display roles description on hover
+  setTitle () {
+    if (this.systemRoles) {
+      let selectEls = $('li.search-choice')
+      $(selectEls).each(i => {
+        let text = selectEls[i].firstChild.innerHTML
+        selectEls[i].firstChild.setAttribute('title', this.systemRoles[text])
+      })
+    }
+  }
+
   bindRoles () {
     const TMP_ROLE = 'TEMPORARY'
 
@@ -351,6 +363,8 @@ class UserController {
         }
         if (this.user.adminRoles) {
           previousRoles = roles
+          // to manually display roles description on roles multi select elements
+          this.setTitle()
           return roles
         } else {
           return previousRoles
--- a/console/src/main/webapp/manager/app/components/user/user.es6
+++ b/console/src/main/webapp/manager/app/components/user/user.es6
@@ -70,13 +70,10 @@ class UserController {
   // to display roles description on hover
   setTitles () {
     if (this.roleDescriptions) {
-      let selectEls = document.querySelectorAll('li.search-choice span')
-      selectEls.forEach(span => {
-        let text = span.innerHTML
-        if (this.roleDescriptions[text]) {
-          span.setAttribute('title', this.roleDescriptions[text])
-        }
-      })
+      [].forEach.call(
+        document.querySelectorAll('li.search-choice span'),
+        span => span.setAttribute('title', this.roleDescriptions[span.innerHTML])
+      )
     }
   }
 
--- a/console/src/main/webapp/manager/app/components/user/user.es6
+++ b/console/src/main/webapp/manager/app/components/user/user.es6
@@ -72,6 +72,7 @@ class UserController {
     // Load role infos for every tab (for confirmation)
     let Role = this.$injector.get('Role')
     this.roles = Role.query()
+    this.systemRoles = Role.query()
     Role.query(roles => { this.allroles = roles.map(r => r.cn) })
 
     this.user.$promise.then(() => {
--- a/console/src/main/webapp/manager/app/components/user/user.es6
+++ b/console/src/main/webapp/manager/app/components/user/user.es6
@@ -68,12 +68,14 @@ class UserController {
 
   // search each choosen span elements and set title manually
   // to display roles description on hover
-  setTitle () {
-    if (this.systemRoles) {
-      let selectEls = $('li.search-choice')
-      $(selectEls).each(i => {
-        let text = selectEls[i].firstChild.innerHTML
-        selectEls[i].firstChild.setAttribute('title', this.systemRoles[text])
+  setTitles () {
+    if (this.roleDescriptions) {
+      let selectEls = document.querySelectorAll('li.search-choice span')
+      selectEls.forEach(span => {
+        let text = span.innerHTML
+        if (this.roleDescriptions[text]) {
+          span.setAttribute('title', this.roleDescriptions[text])
+        }
       })
     }
   }
@@ -87,9 +89,9 @@ class UserController {
     Role.query(roles => {
       this.allroles = roles.map(r => r.cn)
       // get roles informations to get description from template
-      this.systemRoles = {}
+      this.roleDescriptions = {}
       roles.map(r => {
-        this.systemRoles[r.cn] = r.description
+        this.roleDescriptions[r.cn] = r.description
       })
     })
     this.user.$promise.then(() => {
@@ -364,7 +366,7 @@ class UserController {
         if (this.user.adminRoles) {
           previousRoles = roles
           // to manually display roles description on roles multi select elements
-          this.setTitle()
+          this.setTitles()
           return roles
         } else {
           return previousRoles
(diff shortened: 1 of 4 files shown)


landryb:
cant test those days as my build server is dying, but i trust you it does what's expected

console/src/main/webapp/manager/app/components/user/user.es6
```diff
@@ -72,6 +72,7 @@ class UserController {
     // Load role infos for every tab (for confirmation)
     let Role = this.$injector.get('Role')
     this.roles = Role.query()
+    this.systemRoles = Role.query()
```
tonio: There's already a duplicate call to `Role.query()`, I don't think it's a good idea to add one more…

console/src/main/webapp/manager/app/components/user/user.tpl.html
```diff
@@ -50,7 +50,7 @@ <h4>
         <h4 translate>role.system</h4>
         <div class="form-group clearfix">
           <div class="col-sm-6" ng-repeat="role in user.adminRoles" ng-if="user.allroles.indexOf(role) !== -1 && role !== 'ORGADMIN'">
-            <div class="checkbox">
+            <div class="checkbox" ng-repeat="r in user.systemRoles" ng-if="role === r.cn" title="{{r.description}}">
```
tonio: This loop inside loop looks sub-optimal to me…

console/src/main/webapp/manager/app/components/user/user.es6
Gaetanbrl: From template, user.roles doest not contain description, only CN. You suggest to clone instead of call Role.query twice ?

console/src/main/webapp/manager/app/components/user/user.tpl.html
Gaetanbrl: What would you suggest ?

console/src/main/webapp/manager/app/components/user/user.es6
tonio: There should be a single call (current code shouldn't have 2) to `Role.query` and get what you want in a callback, see https://github.com/georchestra/georchestra/blob/19b1c9fc0b688b1e6bf7791dde2162f902db1fe6/console/src/main/webapp/manager/app/components/user/user.es6#L76

console/src/main/webapp/manager/app/components/user/user.tpl.html
tonio: In the callback pointed above, store role or description in a dict with `cn` as key to avoid another loop here.

landryb:
great, works fine except for two (i suppose forgotten) cases:
- no description displayed for 'other roles' on the role selection tab of a user page, works for georchestra roles
- no description displayed if a 'other role' is set as favorite (and thus is listed on the left on the 'all users page') and one hovers on the role name on the same page. works for georchestra roles

Gaetanbrl:
I will check that now @landryb

Gaetanbrl:
>no description displayed for 'other roles' on the role selection tab of a user page, works for georchestra roles

@landryb i'm disturb by the plugin use to select 'other roles'.
Indeed, the plugin havn't options to set hover or tooltip.

First solution (not easiest) is to set a `change `event and catch the parent div with a JS function. In this div, with the same JS function, we could get roles from `<li>`, match with roles description, and set the hover for each `<li>` ...

@tonio maybe know you know a better solution with this plugin ?

console/src/main/webapp/manager/app/components/user/user.tpl.html
Gaetanbrl: Thanks. That's was my first idea, i thought it was more clean to realize it into template. But your advice is more simple to read and effective.

Gaetanbrl:
[This simple solution](https://github.com/georchestra/georchestra/pull/2703/commits/2fb602c58614f86c1d8782c26676ffe007be3953#diff-ae48973307aac36019f3391cc05a521eR76)  with `Jquery` work.

RemiDesgrange:
> [This simple solution](https://github.com/georchestra/georchestra/pull/2703/commits/2fb602c58614f86c1d8782c26676ffe007be3953#diff-ae48973307aac36019f3391cc05a521eR76) with `Jquery` work.

Can you use `document`

Gaetanbrl:
@RemiDesgrange why should i use document ?

RemiDesgrange:
I deleted the comment, I also closed by error. Forget about my comment :-)

landryb:
@fvanderbiest if this jquery solution is not acceptable, what should be used ? Otherwise, is the PR itself can be merged ?

fvanderbiest:
> if this jquery solution is not acceptable, what should be used ?

I'm sorry I have no idea.

landryb:
Then can this PR can be reviewed/merged as is ?

Gaetanbrl:
The package.json contain angular-chosen : 

> Angular Chosen directive is an AngularJS Directive that brings the Chosen jQuery in a Angular way

But [angular-chosen](http://adityasharat.github.io/angular-chosen/) needs jQuery : 
```
<!-- include the js -->
<script src="path/to/jquery.min.js"></script>
```
So, i'm not an expert with Angular and i'm a little bit lost with you'r advise to do not use Jquery here.

If you prefer, simple JS could replace jQuery ?

RemiDesgrange:
jQuery is a thing of the past. Now that vanilla js API (almost) work, we should strive not use jquery in new code and move legacy code away from it when we have the possibility to do so. If you can come with a solution without jQuery, it's great :-)

console/src/main/webapp/manager/app/components/user/user.es6
```diff
@@ -66,14 +66,32 @@ class UserController {
     return (this.delegation.orgs.length !== 0) && (this.delegation.roles.length !== 0)
   }
 
+  // search each choosen span elements and set title manually
+  // to display roles description on hover
+  setTitle () {
```
tonio:
```suggestion
  setTitles () {
```

console/src/main/webapp/manager/app/components/user/user.es6
```diff
@@ -66,14 +66,32 @@ class UserController {
...
+    if (this.systemRoles) {
+      let selectEls = $('li.search-choice')
+      $(selectEls).each(i => {
```
tonio: no need for `$` it's already a jquery object.
`.search-choice span` selector would avoid to fetch first child.

console/src/main/webapp/manager/app/components/user/user.es6
```diff
@@ -66,14 +66,32 @@ class UserController {
...
     let Role = this.$injector.get('Role')
     this.roles = Role.query()
-    Role.query(roles => { this.allroles = roles.map(r => r.cn) })
-
+    Role.query(roles => {
+      this.allroles = roles.map(r => r.cn)
+      // get roles informations to get description from template
+      this.systemRoles = {}
+      roles.map(r => {
+        this.systemRoles[r.cn] = r.description
```
tonio: either rename it to `roleDescriptions` or store the whole role rather than description.

Gaetanbrl:
@RemiDesgrange @tonio thanks for the review.

> If you can come with a solution without jQuery, it's great

Yes it's probably possible if i have no choice, but could you explain me why ? Just to know why jQuery is to avoid the next time.

tonio:
As long as jQuery is an angularjs dependency, it's not so chocking to use it, imho. (it's just usually useless nowadays with modern browsers…)

Gaetanbrl:
@fvanderbiest @landryb a new commit was realized to replace jQuery by querySelector.
`let selectEls = document.querySelectorAll('li.search-choice span')`

Thank you again to all reviewers.

console/src/main/webapp/manager/app/components/user/user.es6
```diff
@@ -68,12 +68,14 @@ class UserController {
...
-  setTitle () {
-    if (this.systemRoles) {
-      let selectEls = $('li.search-choice')
-      $(selectEls).each(i => {
-        let text = selectEls[i].firstChild.innerHTML
-        selectEls[i].firstChild.setAttribute('title', this.systemRoles[text])
+  setTitles () {
+    if (this.roleDescriptions) {
+      let selectEls = document.querySelectorAll('li.search-choice span')
+      selectEls.forEach(span => {
```
tonio: Now that you don't use jQuery selectEls is a nodeList. `forEach` on nodeList is [not implement in IE](https://developer.mozilla.org/fr/docs/Web/API/NodeList/forEach#Compatibilit%C3%A9_des_navigateurs).

You can use  `[].forEach.call` to be IE-proof.
```javascript
[].forEach.call(
  document.querySelectorAll('li.search-choice span'),
  span => span.setAttribute('title', this.roleDescriptions[text.innerHTML])
)
```

Gaetanbrl: > orEach on nodeList is not implement in IE

+1 for jQuery ;)

landryb:
Thanks @tonio for the feedback & review :)

PR #2706, opened 2019-09-11
Console - Fixing GDPR download w/ geonetwork

pmauduit:
Dates in GeoNetwork's metadata table are stored as text / varchar fields, and follow the iso8601 spec (mainly a 'T' instead of a space to separate the date from the hours, they might be suffixed by a timezone identifier, but looking at the metadata table on geo2france-dev, it did not seem to be the case).

The current logic in the code was to try to parse the invert (space instead of 'T') for dates coming from GeoNetwork.

Tests:
* added a utest to specifically parse using the GEONETWORK_DATE_FORMAT constant.
* utests ok (and adapted consequently).
* Runtime tested directly by injecting the class onto dev.geo2france.fr


pmauduit:
integration tests are failing though

PR #2707, opened 2019-09-11
[WIP] console - relocate the GDPR-related endpoints

pmauduit:
Some endpoints are protected by the security-proxy, making them unavailable for regular users, yet these users should be able to access them (to retrieve their own data).

Note:

* I found curious that there are no place in the JS code to be modified ? Only the JSP one ?
* I might have missed something: does the controller code has to check if the user doing the request is actually allowed to do so ? It seems to be the case for delete, I'm not quite sure for download ...

Tests:

* `mvn clean test` ok
* `mvn clean verify -Pit` ok


fvanderbiest:
See also https://github.com/georchestra/datadir/pull/170

PR #2708, opened 2019-09-16
console - fixing serialization of cities in Org

pmauduit:
This PR aims to fix the JSON format returned for the organizations:

Hitting  `<georchestra>/console/private/orgs/<myorg>` 

returns (in case of multivalued cities) a JSON entry as follows:

```
 "cities":["62270,62178"]
```

Where the UI actually expects:

```
 "cities":["62270","62178"]
```

Tests: 
* Added utests to test the Org Attribute Mapper
* mvn test OK
* mvn verify OK


fvanderbiest:
Thanks !
Was it OK on 19.04 finally ?

Issue #2544, opened 2019-03-11
ldap - memberof.ldif seems to be defined two times


severo:
See:

1. https://github.com/georchestra/georchestra/blob/master/ldap/memberof.ldif
2. https://github.com/georchestra/georchestra/blob/master/ldap/docker/modules/memberof.ldif.

Only the first one seems to be useful:

- the second one is copied to `/etc/ldap.dist/modules`: https://github.com/georchestra/georchestra/blob/master/ldap/Dockerfile#L19
- the first one is copied to `/tmp`: https://github.com/georchestra/georchestra/blob/master/ldap/docker/docker-entrypoint.d/01-populate#L34-L35
- `/etc/ldap.dist` is copied to `/etc/ldap`: https://github.com/georchestra/georchestra/blob/master/ldap/docker/docker-entrypoint.d/00-init#L33
- if defined in `SLAPD_ADDITIONAL_MODULES` env var (it's not in the docker-compose.yml: https://github.com/georchestra/georchestra/blob/master/ldap/docker-compose.yml#L15), the second one is imported in slapd: https://github.com/georchestra/georchestra/blob/master/ldap/docker/docker-entrypoint.d/00-init#L93
- finally the first one is imported in slapd: https://github.com/georchestra/georchestra/blob/master/ldap/docker/docker-entrypoint.d/01-populate#L34-L35

I think we should remove the second one. To be confirmed.

severo:
Also:

why changing the backend with `sed` at https://github.com/georchestra/georchestra/blob/master/ldap/docker/docker-entrypoint.d/01-populate#L34. It would be better to change it in the versioned file https://github.com/georchestra/georchestra/blob/master/ldap/memberof.ldif

fvanderbiest:
Closed by #2715

Issue #1835, opened 2017-12-04
no pkill in ldap container


MaxiReglisse:
georchestra_ldap_1 container crashes at startup, with the following message : 

    /docker-entrypoint.d/00-populate: line 54: pkill: command not found
    Waiting for LDAP to stop...Waiting: OK
    run-parts: executing /docker-entrypoint.d/01-indexes
    LDAP indexed
    5a257144 @(#) $OpenLDAP: slapd  (Aug 10 2017 19:12:46) $
	Debian OpenLDAP Maintainers <pkg-openldap-devel@lists.alioth.debian.org>
    5a257144 daemon: bind(6) failed errno=98 (Address already in use)
    5a257144 hdb_db_open: database "dc=georchestra,dc=org": database already in use.
    5a257144 backend_startup_one (type=hdb, suffix="dc=georchestra,dc=org"): bi_db_open failed! (-1)
    5a257144 slapd stopped.

i don't no why this pkill command is missing... but if you add the package procps in Dockerfile, the problem is solved !

étonnant, non ?

pmauduit:
> étonnant, non ?

the ldap docker image should derive from the official debian stretch one, where the `procps` package is no longer installed by default. I already figured out this when trying to launch `ps` in a running container.

The package can be easily added here: https://github.com/georchestra/georchestra/blob/16.12/ldap/Dockerfile#L8

pmauduit:
We did basically the same PR ;-)

MaxiReglisse:
cool !

PR #2715, opened 2019-09-23
LDAP - reorganize files for docker image

fvanderbiest:
This change allows to only `COPY docker-root /` in the dockerfile.
The drawback (which is not really one) is that symlincks have to be used to not duplicate LDIF files.

This also emphasizes the fact that we had two `memberof.ldif` files (see #2544), one of which was unused.

Regarding the inclusion of `ldap-utils` & `procps`:
 * `procps` is required, see #1835 
 * `ldap-utils` is a nice-to-have which only adds 1Mb on a 150Mb image. Leaving it.


landryb:
Fwiw this part of the PR for 3/4 files seems weird: https://github.com/georchestra/georchestra/pull/2715/commits/213a08530e183b713ae14b989a6408e2a1e79590#diff-0baf0f56c9b48799dd931323cd72377c

an ldif file containing a path to another ldif file ? that's sure to confuse users.... or was it supposed to be a symlink ?

fvanderbiest:
Right, it's a symlink.

landryb:
ah well, "interesting". Then people downloading files from the github web UI are confused :)

fvanderbiest:
I tried hard to find the best solution avoiding duplication and complexity, but it seems like there is no silver bullet here

landryb:
Fwiw, as a side effect, i just realized this completely broke the ansible playbook deployments, so far beyond repair.  The 19.04 branch is also broken by e3e4dc082ca but maybe i can fix this one 'easily'..

i'm considering just marking the ansible repo as abandoned, since changes in georchestra are done without (apparently) testing anything outside docker deployment... there's more and more deviation/discrepancy between the 'follow the traditional docs' and "i dont give a f*** i just run docker-compose-yolo". Yes, i'm saddened by this sorry state :)

fvanderbiest:
Thanks Landry for the head-up.
As you can imagine, we do not break things on purpose, it's just a matter of available time.
Before each release I try to check / fix / help with the ansible deployment on my professional time.

But you're right, there's an open question about who's maintaining it & with which resources & for what purpose ...

PR #2718, opened 2019-09-25
Update openldap.md

bendeg:
Step "geOrchestra users and groups" :

first add openssh.ldif, otherwise ERROR 21 when adding testadmin user (objectClass "sshPublicKey" is unknown)


pmauduit:
thanks !

PR #2721, opened 2019-09-26
Better Org referents impl

RemiDesgrange:
A better Org referent impl, much more cleaner with more test.


console/src/main/java/org/georchestra/console/ws/editorgdetails/EditOrgDetailsFormController.java
```diff
@@ -24,6 +24,7 @@
 import org.georchestra.console.dto.orgs.OrgExt;
 import org.georchestra.console.ws.utils.Validation;
 import org.springframework.beans.factory.annotation.Autowired;
+import org.springframework.security.access.annotation.Secured;
```
cmangeat: Of no use ?

PR #2724, opened 2019-09-27
Console - enhance email template by adding orgs to new users

RemiDesgrange:
Add org of the user when new user register (see also [datadir](https://github.com/georchestra/datadir/pull/171)). Having Org in your template is nor mandatory. If it's present it will be populated accordingly

### DIFF
--- a/console/src/main/java/org/georchestra/console/mailservice/EmailFactory.java
+++ b/console/src/main/java/org/georchestra/console/mailservice/EmailFactory.java
@@ -112,7 +112,7 @@ public class EmailFactory {
         email.set("name", userName);
         email.set("uid", uid);
         email.set("email", userEmail);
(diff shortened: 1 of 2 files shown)


console/src/main/java/org/georchestra/console/mailservice/EmailFactory.java
```diff
@@ -150,6 +150,9 @@ public void sendNewAccountNotificationEmail(ServletContext servletContext, List<
         email.set("name", userName);
         email.set("uid", uid);
         email.set("email", userEmail);
+        if (userOrg == null) {
+            userOrg = "None";
```
fvanderbiest: I'd rather have an empty string here, since there is no email template i18n

PR #2726, opened 2019-09-27
Add ldapPublicKey to the list of objectclass when mapping an ldap use…

### LINKED ISSUE #2725: #2713 broke saving user in console
Since the addition of the ssh pubkey attr, saving an user in the console blows, cf trace in https://github.com/georchestra/georchestra/pull/2713#issuecomment-534095703

Now looking a bit at the code, i think the `ldapPublicKey` objectclass should be added to https://github.com/georchestra/georchestra/blob/master/console/src/main/java/org/georchestra/console/ds/AccountDaoImpl.java#L423 - testing it here fixes the crash.

landryb:
…r (fixes #2725)

otherwise, saving an user blows with [LDAP: error code 65 - attribute 'sshPublicKey' not allowed]

### DIFF
--- a/console/src/main/java/org/georchestra/console/ds/AccountDaoImpl.java
+++ b/console/src/main/java/org/georchestra/console/ds/AccountDaoImpl.java
@@ -421,7 +421,7 @@ public final class AccountDaoImpl implements AccountDao {
     private void mapToContext(Account account, DirContextOperations context) {
 
         context.setAttributeValues("objectclass", new String[] { "top", "person", "organizationalPerson",
-                "inetOrgPerson", "shadowAccount", "georchestraUser" });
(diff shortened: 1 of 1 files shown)


pmauduit:
The fix is a bit more complicated than adding every users the objectclass 'ldapPublicKey', because as stated here: https://github.com/georchestra/georchestra/issues/2725#issuecomment-535843059, this requires to define at least one sshPublicKey attribute on the user.

I reworked the PR a bit, for now utests & integration tests are ok (as long as we use the current georchestra/ldap:latest docker image !), but it will still require some runtime tests.

pmauduit:
...

I cannot edit the current PR so that it points to the official repo instead of @landryb 's one ...

PR #2727, opened 2019-09-27
hide orgs details when user have no orgs

RemiDesgrange:


console/src/main/webapp/WEB-INF/views/editUserDetailsForm.jsp
```diff
@@ -173,7 +173,7 @@
     <!-- /angularjs app dependencies -->
 
     <div class="col-lg-5 col-lg-offset-1" ng-app="manager" ng-strict-di ng-controller="StandaloneController">
-      <fieldset>
+      <fieldset ng-if="Object.hasOwnProperty(org)">
```
tonio: [WAT](https://i.kym-cdn.com/photos/images/original/000/173/580/Wat.jpg)

That's not how you're supposed to use this method. `ng-if="org"` or `ng-if="org.id"` should be enough (not sure of which, please check in StandaloneController in app.es6 what's in).

RemiDesgrange: I tried `ng-if="org"` and it didn't worked. Sorry, I'm normally using real language :trollface: 

Nevertheless, I found this `Object.hasOwnProperty` code dirty, that's why I asked for review. Thanks for the tips

PR #2731, opened 2019-10-02
Change UI when user is referent. 

RemiDesgrange:
Improve UI (maybe UX ? not an expert :sweat_smile: )


fvanderbiest:
A screenshot before / after would help us see what was done here :-)
Thanks.

PR #2732, opened 2019-10-02
Console - adding SSH key support to the object mapper.

pmauduit:
This PR is the following of this one:

https://github.com/georchestra/georchestra/pull/2726

Tests:
* utests ok
* IT ok
* Still runtime tests to do

### DIFF
--- a/console/src/main/java/org/georchestra/console/dto/AccountImpl.java
+++ b/console/src/main/java/org/georchestra/console/dto/AccountImpl.java
@@ -21,7 +21,9 @@ package org.georchestra.console.dto;
 
 import java.io.Serializable;
 import java.time.LocalDate;
+import java.util.Arrays;
 import java.util.Date;
+import java.util.List;
 import java.util.Objects;
 
 import javax.naming.InvalidNameException;
@@ -130,6 +132,9 @@ public class AccountImpl implements Serializable, Account {
     @JsonProperty(UserSchema.CONTEXT_KEY)
     private String context;
 
+    @JsonProperty(UserSchema.SSH_KEY)
+    private String[] sshKeys;
+
     // Organization from ou=orgs,dc=georchestra,dc=org
     // Json export is defined on the getter getOrg()
     private String org;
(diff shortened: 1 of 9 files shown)


console/src/main/java/org/georchestra/console/dto/AccountImpl.java
```diff
@@ -148,7 +153,7 @@ public String toString() {
                 + ", roomNumber='" + roomNumber + '\'' + ", stateOrProvince='" + stateOrProvince + '\''
                 + ", homePostalAddress='" + homePostalAddress + '\'' + ", shadowExpire='" + shadowExpire + '\''
                 + ", privacyPolicyAgreementDate='" + privacyPolicyAgreementDate + '\'' + ", context='" + context + '\''
-                + ", org='" + org + '\'' + '}';
+                + ", org='" + org + '\'' + ", sshKeys='" + sshKeys + "'}";
```
groldan: looks like this should be `Arrays.toString(sshKeys)`

pmauduit:
Runtime tested: OK

Utests are failing now, they were ok when I PRed, still to figure out why.

console/src/main/java/org/georchestra/console/dto/AccountImpl.java
pmauduit: I thought it would be ok to leave Java do the job, but I think you're right, it would be more explicit to have the toString()

landryb:
Also tested here, can save properties on a user, and add/remove users from an org. So that seems to fix #2725 for me.

fvanderbiest:
Thanks !
Please rebase.

pmauduit:
Ok rebased (waiting for the checks above to pass though ...)

pmauduit:
> (waiting for the checks above to pass though ...)

It passes (and I checked locally before force-pushing anyway)

pmauduit:
> Also tested here, can save properties on a user, and add/remove users from an org. So that seems to fix #2725 for me.

Thanks for having runtime-tested it btw

console/src/main/java/org/georchestra/console/dto/AccountImpl.java
groldan: oh, the issue is that String[].toString() witll give you garbage

groldan:
LGTM

Issue #2737, opened 2019-10-04
console - Wrong character encoding - Export users as CSV


Gaetanbrl:
With some accents or other typicals french characters, we get wrong representation.

An example with a fictive data :  donnÃ©

![image](https://user-images.githubusercontent.com/16317988/66225399-a1008100-e6d8-11e9-911e-1e2aba9126d1.png)

Gaetanbrl:
Encoding seems good [here](https://github.com/georchestra/georchestra/blob/master/console/src/main/webapp/manager/app/components/users/users.es6#L99).

So, i think we lose characters from this line and we need to specify 'utf-8' like that : 

`new Blob([csvString], { type: 'text/csv; charset=utf-8' }),`

Gaetanbrl:
This seems to works fine : 
`const blob = new Blob(['\ufeff', result.data], { type: mimetype })`

Gaetanbrl:
I move here a good question asked by @fvanderbiest :
> OK, I'd like to collect other people's opinion here before deciding to merge it or not.
@landryb @pmauduit @RemiDesgrange @cmangeat Do you think it's a good idea to create invalid UTF-8 documents to please Excel and ease user experience ?

landryb:
@fvanderbiest technically, a document starting with BOM is a valid UTF-8 document, no ? As much as i hate excel, it's what people use.... and we should strive to ease their experience, even if it means adding gross hacks.

landryb:
That's why i usually run away screaming when faced with encoding issues :) Seems BOM usage is discouraged, but if that's the common denominator to workaround most issues..

fvanderbiest:
OK, as long as LibreOffice is able to cope with it also ...

PR #2742, opened 2019-10-07
Insert BOM to encode CSV correctly

Gaetanbrl:
This PR specify the encoding to display specials characters (é,à,ù,...) for **user(s) export only**.

Link to #2737 .

### DIFF
--- a/console/src/main/webapp/manager/app/components/users/users.es6
+++ b/console/src/main/webapp/manager/app/components/users/users.es6
@@ -96,7 +96,7 @@ class UsersController {
         default:
           mimetype = `text/${fileType}`
       }
-      const blob = new Blob([result.data], { type: mimetype })
+      const blob = new Blob(['\ufeff', result.data], { type: mimetype })
       const a = document.createElement('a')
       a.href = window.URL.createObjectURL(blob)
       a.target = '_blank'


fvanderbiest:
BOMs are a nightmare ...
https://fr.wikipedia.org/wiki/Indicateur_d%27ordre_des_octets#UTF-8

What if the user manually selects UTF-8 when importing the file in LibreOffice?
I guess it fixes the issue.

Gaetanbrl:
> What if the user manually selects UTF-8 when importing the file in LibreOffice?

Not easy to do for each user with Excel, that's suppose to use LibreOffice only.

fvanderbiest:
> Excel

Are there people still using it ?
Just kidding ;-)

OK, I'd like to collect other people's opinion here before deciding to merge it or not.
@landryb @pmauduit @RemiDesgrange @cmangeat Do you think it's a good idea to create non valide UTF-8 documents to please Excel and ease user experience ?

fvanderbiest:
See also https://stackoverflow.com/questions/2223882/whats-the-difference-between-utf-8-and-utf-8-without-bom

Gaetanbrl:
> BOMs are a nightmare ...
You're right, it's maybe not the best way...

RemiDesgrange:
I'm noy using windows, but normally, if you send the right headers (`Content-Type: text/csv; charset=utf-8` it should work ? I never set the BOM when downloading files, and Excel user were not disturbed :thinking: There should have a better way

console/src/main/webapp/manager/app/components/users/users.es6
```diff
@@ -96,7 +96,7 @@ class UsersController {
         default:
           mimetype = `text/${fileType}`
       }
-      const blob = new Blob([result.data], { type: mimetype })
+      const blob = new Blob(['\ufeff', result.data], { type: mimetype })
```
RemiDesgrange: maybe 
```
const blob = new Blob(['\ufeff', result.data], { type: `mimetype; charset=utf-8` })
```

Gaetanbrl: Sorry but  `text/csv; charset=utf-8` don't work.

fvanderbiest:
If we all agree it's not a good idea, let's close this one.

landryb:
I will always agree on 

> BOMs are a nightmare ... 

and actually would expand it to 

> encodings are a nightmare

fvanderbiest:
Directly from the trenches... I can confirm that it's a nightmare for Excel users to open the file.
So I'm +1 to add the BOM, I finally made my mind.

PR #2744, opened 2019-10-08
root pom - renaming sentry profile

pmauduit:
The main motivation is that using log4j to send logs to sentry might not
be the best way of integrating sentry. We want to just send the
exceptions actually, sentry does not need to get the whole logs.

The caveat of this approach is that it requires to define an extra bean
to the spring XML configuration (see
https://docs.sentry.io/clients/java/integrations/#spring). Using the
log4j1 module only needed to modify the log4j configuration.

Also moving to the new version of sentry-spring (1.7.27)

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -798,12 +798,12 @@
   </reporting>
   <profiles>
     <profile>
-      <id>sentry-log4j</id>
+      <id>sentry-spring</id>
       <dependencies>
         <dependency>
           <groupId>io.sentry</groupId>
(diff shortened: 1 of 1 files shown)


RemiDesgrange:
As a side notes, we can say that if you don't set the `SENTRY_DSN` in your env var or pass it as Java option, sentry is not activated.

fvanderbiest:
LGTM, thanks !

pmauduit:
FYI I ported back the PR to 18.06 & 19.04

Issue #2730, opened 2019-09-30
Export org/role as CSV


Gaetanbrl:
We can export users as CSV/VCF from #2278 by @landryb.
This issue is about org/role export as CSV.

Gaetanbrl:
@landryb is it correct for you (org as csv) ?

![image](https://user-images.githubusercontent.com/16317988/66225802-a0b4b580-e6d9-11e9-84c9-1f85b9571fbd.png)

landryb:
Yup, except it lacks the description field showing the area codes.

RemiDesgrange:
@Gaetanbrl encoding problem ?

Gaetanbrl:
@RemiDesgrange #2737 - same for user

Gaetanbrl:
@landryb what is the maximum number of area codes per user ?

landryb:
no limit, in my case 4100 but an sdi could be on france and have 36k...

Gaetanbrl:
Are you sure to display all area code in the CSV ?

landryb:
that's the most useful information to me. And actually, it might be more useful/usable in calc/office if displayed as one line per combination of values instead of concatenating multivalued attributes, ie 20 lignes for an org that has 1 member and an area of 20 codes, 10 lines for an org with 2 members and an area of 5 codes, etc....
@fvanderbiest what do you think ?

fvanderbiest:
> it might be more useful/usable in calc/office if displayed as one line per combination of values instead of concatenating multivalued attributes, ie 20 lignes for an org that has 1 member and an area of 20 codes, 10 lines for an org with 2 members and an area of 5 codes, etc....

Seems weird...
I'd rather expect something like this (eventually with quotes, etc):
```csv
# id;name;members;featureIdList...
c2c;Camptocamp;Remi,François,Pierre,Antoine,Christophe;73000,38500,73370
...
```
Please note that @RemiDesgrange is currently working on replacing the description field by a featureId field (see #2638)

Also : please consider using commons-csv as we did with #2733

landryb:
> Please note that @RemiDesgrange is currently working on replacing the description field by a featureId field (see #2638)

well it would be nice to know what's the actual plan (ie how it will be laid in ldap, where does the featureId come from, etc etc), because many things rely on it here. And nobody likes spending time on something that will have to be dumped in 3 months :)

> Seems weird...
> I'd rather expect something like this (eventually with quotes, etc):
> 
> ```
[...]

PR #2746, opened 2019-10-08
console: add organization info to CSV users export

groldan:
Add the following mappings from outlook CSV format
to Account:

`Web Page`:     -> Org.url
`Notes`:        -> Org.description
`Home City`:    -> Account.locality
`Company`:      -> Org.name
`Business City`:-> Org.cities


landryb:
A bit sad that it totally ignored the discussion in #2730 but whatever. I should rejoice that in some way orgs/cities are available in some kind of csv export..

PR #2771, opened 2019-10-15
Security-proxy - rewrite cookie management

pmauduit:
Cookie management was kind of undocumented in the SP, and required a rework, because the code was unreadable at least to me:

When hitting a security-proxified webapp, some cookies can be retrieved, but they should not be transmitted "as is" to the client without some checks:

* We need to extract the jsessionid and avoid transmitting it to the client, but instead keeping it in the user's session (as a hashmap, indexed by the cookie path => JSESSIONID=value) for further requests to the same webapp by the same user. Transmitting it to the client could invalidate the previous JSESSIONID used by the SP (e.g. if no path limits the scope of the cookie. This does not seem to be the case on the georchestra webapps I could test, but since this was a precaution in the current version of the code, I preferred to keep it). Here is what this PR aims to introduce as changes:

* The other cookies should not harm and can be probably transmitted "as is", but an extra precaution is taken so that the scope (Path=/...) is limited to the concerned proxified webapp.

* Onto the SP, one can configure the cookie sent for keeping the session with the HttpOnly flag, so that we make sure that any JS code cannot mangle (remove / edit) the session client-side (see https://developer.mozilla.org/en-US/docs/Web/HTTP/Cookies#Secure_and_HttpOnly_cookies)

* I took the opportunity to use slf4j instead of commons-logging / log4j1 as it simplifies the log.*() calls and string formatting into the `HeadersManagementStrategy` class.

Instead of parsing cookies using regex and/or String.split(), I have preferred introducing the use of the HttpCookie class https://docs.oracle.com/javase/7/docs/api/java/net/HttpCookie.html. The main regret I have in regards to this class is that there is no way to get back the "Set-Cookie: ..." value, which would have been useful in our case. That is why I allowed myself to mangle the cookie path using a regex ... I did not take the time to check, maybe Spring has a better class for this.

Tests:
* utests added
* mainly runtime tested onto dev.geo2france.fr, but would definitely require more thorough tests, especially in regards to using other servlet containers (e.g. tomcat).

### DIFF
--- a/security-proxy/src/main/java/org/georchestra/security/HeadersManagementStrategy.java
+++ b/security-proxy/src/main/java/org/georchestra/security/HeadersManagementStrategy.java
@@ -24,14 +24,14 @@ import static org.georchestra.security.HeaderNames.BASIC_AUTH_HEADER;
 import static org.georchestra.security.HeaderNames.CONTENT_LENGTH;
 import static org.georchestra.security.HeaderNames.COOKIE_ID;
 import static org.georchestra.security.HeaderNames.HOST;
-import static org.georchestra.security.HeaderNames.LOCATION;
+import static org.georchestra.security.HeaderNames.PROTECTED_HEADER_PREFIX;
 import static org.georchestra.security.HeaderNames.REFERER_HEADER_NAME;
 import static org.georchestra.security.HeaderNames.SEC_PROXY;
 import static org.georchestra.security.HeaderNames.SEC_ROLES;
 import static org.georchestra.security.HeaderNames.SEC_USERNAME;
 import static org.georchestra.security.HeaderNames.TRANSFER_ENCODING;
-import static org.georchestra.security.HeaderNames.PROTECTED_HEADER_PREFIX;
 
+import java.net.HttpCookie;
 import java.util.ArrayList;
 import java.util.Collection;
 import java.util.Collections;
@@ -44,15 +44,16 @@ import javax.servlet.http.HttpServletRequest;
 import javax.servlet.http.HttpServletResponse;
 import javax.servlet.http.HttpSession;
 
-import org.apache.commons.logging.Log;
-import org.apache.commons.logging.LogFactory;
 import org.apache.http.Header;
 import org.apache.http.HttpHeaders;
 import org.apache.http.HttpResponse;
 import org.apache.http.client.methods.HttpRequestBase;
 import org.apache.http.message.BasicHeader;
-import org.springframework.beans.factory.annotation.Autowired;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
 import org.springframework.util.StringUtils;
+
+import com.google.common.annotations.VisibleForTesting;
 
 /**
  * A strategy for copying headers from the request to the proxied request and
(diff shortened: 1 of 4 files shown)


pmauduit:
PR rebased

pmauduit:
Deployed on geo2france-dev, no issues with:

* anonymous mode / mapfishapp
* connect onto CAS
* cookie management from the CMS / GDPR acceptance
* geoserver
* geonetwork
* geoserver rest

security-proxy/src/main/java/org/georchestra/security/HeadersManagementStrategy.java
```diff
@@ -303,59 +313,114 @@ public synchronized void copyResponseHeaders(HttpServletRequest originalRequest,
...
+     */
+    @VisibleForTesting
+    public void handleResponseCookies(String originalRequestURI, HttpServletResponse finalResponse, Header[] headers,
             HttpSession session) {
-        String originalPath = originalRequestURI.split("/")[0];
+        String overridenPath = null;
+        try {
+            overridenPath = originalRequestURI.split("/")[1];
+        } catch (Exception e) {
+        }
```
RemiDesgrange: Can't get this part, why are we silently continue ? maybe at least log ? it means that `overridenPath` will be null. Could lead to NPE.

security-proxy/src/main/java/org/georchestra/security/HeadersManagementStrategy.java
```diff
@@ -303,59 +313,114 @@ public synchronized void copyResponseHeaders(HttpServletRequest originalRequest,
...
-                    if (cookies.length() > 0)
-                        cookies.append("; ");
-                    cookies.append(cookie);
                 }
+                // a path is already set on the cookie, but we make sure to limit its scope
+                // to the current proxified webapp
+                else {
+                    logger.debug("Current cookie {} has a path set to {}, forcing it to {}", currentCookie.getName(),
+                            currentCookie.getPath(), overridenPath);
+                    actualHeaderValue = actualHeaderValue.replaceAll("Path=[^;$]+", "Path=/" + overridenPath);
```
RemiDesgrange: `overridenPath` could be `null` here, what we want is maybe an empty string in this case ?

security-proxy/src/main/java/org/georchestra/security/HeadersManagementStrategy.java
pmauduit: This should never happen, because if we got into this part of the code, it is due to a call to `<scheme>://<my georchestra>/<webapp` declared in the `targets-mapping.properties`. The default request mapping is handled before in the code:

https://github.com/georchestra/georchestra/blob/19.04/security-proxy/src/main/java/org/georchestra/security/Proxy.java#L352-L361

I think it should be even safe to throw an exception here actually

security-proxy/src/main/java/org/georchestra/security/HeadersManagementStrategy.java
pmauduit: if we throw an exception before, we will be ensured that overridenPath won't be null

PR #2776, opened 2019-10-16
Update tomcat.md

bendeg:
Add tomcat instances configuration : 
1) copy of /usr/share/tomcat8 for each new tomcat instances
2) modification of files /etc/init.d/tomcat-xxx

### DIFF
--- a/docs/setup/tomcat.md
+++ b/docs/setup/tomcat.md
@@ -93,6 +93,7 @@ sudo tomcat8-instance-create -p 8180 -c 8105 /var/lib/tomcat-proxycas
 
 Then:
 ```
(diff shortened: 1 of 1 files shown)


RemiDesgrange:
BTW @landryb @fvanderbiest any idea why all command are prefixed by `sudo` ?

landryb:
Dunno, usually i'd go for commands prefixed with a # sign for _command is meant to be run as root_ and prefixed by $ for _command is meant to be run as regular user_, but that might be an old-school unix convention unheard by ppl in 2019.

fvanderbiest:
Thanks @bendeg for the fix.
Thanks also @landryb and others for the suggestion. Please create another issue if you think this requires action.

Issue #2782, opened 2019-10-16
[master] Geoserver - Bump to version 2.15.3


pmauduit:
For the Deutsch-Telekom project, we had to upgrade to a new version of GeoServer. A branch in the GS fork has been prepared for this (see https://github.com/georchestra/geoserver/tree/2.15.3-georchestra), but there are still some work to wire it to the georchestra repository.

@groldan do you know if the directory here https://github.com/georchestra/georchestra/tree/master/geoserver/lib is still needed ? Since we have been with GeoTools v21.3 in this version of GeoServer.

I will prepare a branch in this repository to be able to PR.

fvanderbiest:
superseded by 2.16.1

PR #2783, opened 2019-10-16
[WIP] GS - v2.15.3

pmauduit:
See https://github.com/georchestra/georchestra/issues/2782

Tests: not checked the testsuite yet, nor tested at runtime outside of the DT context which is kind of particular.

### DIFF
--- a/geoserver/pom.xml
+++ b/geoserver/pom.xml
@@ -12,15 +12,15 @@
   <properties>
     <server>generic</server>
     <skipTests>true</skipTests>
-    <gs.version>2.15.0</gs.version>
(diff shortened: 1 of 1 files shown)


fvanderbiest:
Rennes Metropole is taking care of the required additional testing.
 We'll have 2.15.3 for the 19.12 release.

pmauduit:
Testsuite seems ok (only utests on customized artifacts). Surprinsingly, tests are skipped by default in the pom:
https://github.com/georchestra/georchestra/blob/19.04/geoserver/pom.xml#L13

fvanderbiest:
Superseded by #2850

PR #2788, opened 2019-10-18
Update tomcat.md

bendeg:
line 281 :

sudo git clone -b gn3.4.1 .......

instead of

sudo git clone -b gn3.0.x


fvanderbiest:
Good catch !
Thanks

PR #2800, opened 2019-10-24
Add GitHub actions

fvanderbiest:
Branches master, 19.04, 18.06 should build images stored in docker hub
Other branches should build images stored in github registry

Current status:
- [x] geonetwork
- [x] geoserver
- [x] analytics
- [x] atlas
- [x] cas
- [x] console
- [x] extractorapp
- [x] geowebcache standalone
- [x] header
- [x] mapfishapp
- [x] security proxy


pmauduit:
> it will require development of specific actions.

There are already some (https://github.com/marketplace/actions/publish-docker), and maven can also already publish docker image by itself (using the docker profile then `mvn docker:build docker:push`, as long as the docker-hub server configuration is set in the m2 settings.xml file.

fvanderbiest:
@RemiDesgrange @pmauduit Can you update the checkboxes in the issue description, to know where we stand ? Thanks !

pmauduit:
> Can you update the checkboxes in the issue description, to know where we stand ? Thanks !

does checking a box mean that the job is done on each and every branches ? because that is not the case for mfapp. The only one currently being built by gh actions is the SP on master.

pmauduit:
an action has been defined for geonetwork in its own repository for the `georchestra-gn3.8.2-master` branch.

pmauduit:
I pushed a commit to create some gh actions, taking inspiration from the current ones already on master (sp & console), see
https://github.com/georchestra/georchestra/commit/7b6f2cb5821a6df99b72ea85ae79c68ce4e14066

Issue #2795, opened 2019-10-22
GDPR - add an option to disable account deletion


fvanderbiest:
Some professionnal plateforms on intranet are not subject to GDPR.
Having users deleting their account (accidentally or not) is not expected in this context.

We should implement in `console.properties` a boolean value like `gdpr.enable` (defaulting to `true`). 

When set to `false`, it:
 * hides the "delete account" button
 * prevents the service from being called programatically

I guess it does not harm to keep the "download my data" button, even if `gdpr.enable = false`

fvanderbiest:
PRs are not merged nor reviewed, and, to my knowledge, do not obey the issue description, since the boolean flag disables both account deletion and data download when only account deletion is required.

Maybe the option is badly named. It should probably be renamed into `gdpr.account.deletion`.

RemiDesgrange:
Ahhhh I thought you wanted to disable the whole feature :-) I'll update the PR accordingly

fvanderbiest:
> Maybe the option is badly named. 

Yes, the option should be renamed into gdpr.allowAccountDeletion (defaulting to true) to make it clear what it really means.
Sorry for the noise.

fvanderbiest:
> the option should be renamed into gdpr.allowAccountDeletion

not taken into account ?

pmauduit:
> not taken into account ?

missed it.

PR #2806, opened 2019-10-29
Console - toggle GDPR

RemiDesgrange:
for example for internal use of georchestra or for country where GDPR
    doesn't apply.

See https://github.com/georchestra/datadir/pull/173 for datadir related PR

Issue : https://github.com/georchestra/georchestra/issues/2795

### DIFF
--- a/console/src/main/java/org/georchestra/console/ws/backoffice/users/UsersExport.java
+++ b/console/src/main/java/org/georchestra/console/ws/backoffice/users/UsersExport.java
@@ -19,18 +19,7 @@
 
 package org.georchestra.console.ws.backoffice.users;
 
-import java.io.File;
-import java.io.IOException;
-import java.util.Arrays;
-import java.util.List;
-import java.util.Set;
-import java.util.stream.Collectors;
-import java.util.stream.StreamSupport;
-
-import javax.servlet.http.HttpServletResponse;
-
-import com.google.common.annotations.VisibleForTesting;
-import com.sun.org.apache.xpath.internal.operations.Bool;
+import lombok.NonNull;
 import org.apache.commons.io.IOUtils;
 import org.georchestra.console.dao.AdvancedDelegationDao;
 import org.georchestra.console.ds.DataServiceException;
@@ -47,7 +36,14 @@ import org.springframework.security.core.context.SecurityContextHolder;
 import org.springframework.stereotype.Controller;
 import org.springframework.web.bind.annotation.*;
 
-import lombok.NonNull;
+import javax.servlet.http.HttpServletResponse;
+import java.io.File;
+import java.io.IOException;
+import java.util.Arrays;
+import java.util.List;
+import java.util.Set;
+import java.util.stream.Collectors;
+import java.util.stream.StreamSupport;
 
 @Controller
 public class UsersExport {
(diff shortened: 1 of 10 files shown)


console/src/main/java/org/georchestra/console/ws/backoffice/users/UsersExport.java
```diff
@@ -53,6 +56,10 @@
 
     private UserInfoExporter accountInfoExporter;
 
+    @VisibleForTesting
+    @Value("${gdpr.enable}")
+    String gdprEnable;
```
pmauduit: the VisibleForTesting annotation might suggest more than "package-private" modifier (i.e. public or so). I'd have put the member private, and defined getters/setters on it, but no idea how spring can react to this, to "autowire" it using the value annotation.

Also isn't spring capable of automatically casting true / false to a boolean ?

console/src/main/java/org/georchestra/console/ws/backoffice/users/UsersExport.java
```diff
@@ -68,6 +75,15 @@
     @RequestMapping(method = RequestMethod.GET, value = "/account/gdpr/download", produces = "application/zip")
     public void downloadUserData(HttpServletResponse response)
             throws NameNotFoundException, DataServiceException, IOException {
+        /*
+         * Disabling this endpoint is the gdpr.enable property is set to false. we are
```
pmauduit: "*if the gdpr.enable"

console/src/main/java/org/georchestra/console/ws/backoffice/users/UsersExport.java
```diff
@@ -68,6 +75,15 @@
...
+         * not using spring boot, so we cannot disable this endpoint with
+         * the @ConditionalOnProperty annotation.
```
pmauduit: not sure it is necessary to mention the fact that the annotation is unavailable

console/src/main/java/org/georchestra/console/ws/backoffice/users/UsersExport.java
pmauduit: relying on the rest of your PR, spring does not care about private member variables when it deals with the @Value annotation, as it probably does some Java reflection to access the field it needs to set.

console/src/main/java/org/georchestra/console/ws/backoffice/users/UsersExport.java
RemiDesgrange: Yeah, dunno, I was just bored that I found this perfect annotation in the doc and then my IDE said "Nope, you can't use that" So I mentioned it, but you're probably right.

fvanderbiest:
Customer is not interested anymore in this feature. Closing.

fvanderbiest:
Well, it is considered helpful in some cases ...

@RemiDesgrange can you rebase ? Thanks

cmangeat (approved):

pmauduit:
Runtime tested:

* [ ] UsersController: /account/gdpr/delete
* [ ] EditUserDetailsFormController: /account/userdetails
* [ ] UserExport: /account/gdpr/download

If gdpr.enable = false, then the previous endpoints are not accessible, if gdpr.enable = true, they are.

pmauduit:
rebased / force-pushed

PR #2814, opened 2019-11-05
Getting rid of charset guessing 2812

pmauduit:


pmauduit:
Deployed on geo2france-dev, no issue encountered so far with this particular version of the SP.

fvanderbiest:
> Deployed on geo2france-dev, no issue encountered so far with this particular version of the SP.

Also tested several WMS servers on my side with no issue.
LGTM

PR #2816, opened 2019-11-06
console: null-proof GDPR record parsing and encoding

groldan:
Make sure no null value coming from the database can
break the GDPR data export.

Use commons-csv to write csv file to ensure the contents
of the files are correctly formatted.


console/src/main/java/org/georchestra/console/ws/backoffice/users/GDPRAccountWorker.java
```diff
@@ -238,37 +246,51 @@ private static Path createZipFile(@NonNull Path bundleFolder) {
     }
 
     private static interface ContentProducer<T> extends Consumer<T>, AutoCloseable {
+        static <V, R> R ifNonNull(@Nullable V value, Function<V, R> mapper) {
```
pmauduit: Not the easiest to read / catch it IMHO, but fine

PR #2820, opened 2019-11-08
Update openldap.md

bendeg:
I would like to add a little explanation for "memberOf" in bootstrap.ldif step because, for a non-expert person, it doesn't tell very clearly how to find out the name we have to use.

### DIFF
--- a/docs/setup/openldap.md
+++ b/docs/setup/openldap.md
@@ -31,6 +31,9 @@ sudo ldapadd -Y EXTERNAL -H ldapi:/// -f bootstrap.ldif
 
 If successful, the above command should display: ```adding new entry "olcDatabase=mdb,cn=config"```.
 
+Now, be careful : after a fresh slapd install there's already a olcDatabase={1}mdb.ldif
+file in /etc/ldap/slapd.d/cn=config/ and this command has created a new one (most probably olcDatabase={2}mdb.ldif after a fresh slapd install) : you must update memberOf.ldif file with that name in "The "memberOf" overlay" step.
+
 
 ## Root DN
 


fvanderbiest:
Thanks !

bendeg:
I closed this PR before it was merged : re-opening it, sorry...

fvanderbiest:
@pmauduit can you check this is correct ? Thanks.

landryb:
This applies mostly only to the default debian setup (of creating a database upon install), but since it's more or less what we target..

bendeg:
I don't mind adding a comment that specifically targets debian users  if needed. Just ask and I'll modify accordingly

docs/setup/openldap.md
```diff
@@ -31,6 +31,9 @@ sudo ldapadd -Y EXTERNAL -H ldapi:/// -f bootstrap.ldif
 
 If successful, the above command should display: ```adding new entry "olcDatabase=mdb,cn=config"```.
 
+Now, be careful : after a fresh slapd install there's already a olcDatabase={1}mdb.ldif
```
pmauduit: I would say instead:

"After a fresh slapd install, you might already have a database set up for you, depending on how the package manager of the distribution is configured, or if the package planned to do so.

To check the database number in your LDAP setup, you can go through the `/etc/ldap/slapd.d/cn=config/` subdirectory and check the files `olcDatabase={x}mdb.ldif`". The different files should contain a `olcSuffix` attribute. The expected geOrchestra database is the one which contains a `olcSuffix: dc=georchestra,dc=org`.

the memberOf.ldif file should match the expected database number in OpenLDAP, and you might need to adapt this LDIF file.
"

bendeg: +1
Indeed, that's THE answer : checking the file that contains right olcSuffix.

I may seem obvious for those who knows ldap, it is not for others, like me...

Issue #2786, opened 2019-10-18
analytics - CSV export on master is broken


fvanderbiest:
/analytics/ws/export/ogclayers?month=10&year=2019

Caused by: java.lang.NullPointerException
	at org.georchestra.analytics.AbstractApplication.respondCSV(AbstractApplication.java:163)
	at org.georchestra.analytics.AbstractApplication.exportCSV(AbstractApplication.java:230)
	at org.georchestra.analytics.OGCStats.exportLayers(OGCStats.java:86)

PR #2823, opened 2019-11-14
as count is jsonified as an int (OGCStatsModel, l57), don't try to re…

cmangeat:
…ad it as a json string

https://github.com/georchestra/georchestra/issues/2786


fvanderbiest:
To backport to 19.04 IIRC

PR #2824, opened 2019-11-14
analytics - filter on role where role and group were ignored, filter on group xor…

cmangeat:
… role where group was taken into account


analytics/src/main/java/org/georchestra/analytics/StatisticsController.java
```diff
@@ -792,8 +794,12 @@ private String convertUTCDateToLocal(String rawDate, GRANULARITY granularity) th
     private String getGroup(JSONObject payload) throws JSONException {
         if (payload.has("group"))
```
fvanderbiest: It seems strange to keep this test IMO

PR #2847, opened 2019-12-02
19.12 migration notes for LDAP

fvanderbiest:
Untested for now

### DIFF
--- a/migrations/19.12/README.md
+++ b/migrations/19.12/README.md
@@ -2,8 +2,30 @@
 
 ## LDAP upgrade
 
-With [#2149](https://github.com/georchestra/georchestra/issues/2149), a new attribute (privacyPolicyAgreementDate) is added to the geOrchestra users in LDAP, to keep track of the date when the user agreed to the terms and conditions (or privacy policy) of the platform.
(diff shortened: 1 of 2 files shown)


landryb:
instead of dumping/dropping/reimporting users, wouldnt it be simpler to use the 'modify' ldap operation to add the new classes ? sed on an ldif feels awkward.. while the proper atomic way for a single user entry would be

```
dn: uid=testuser,ou=users,dc=georchestra,dc=org
changetype: modify
add: objectClass
objectClass: georchestraUser
```

so something along the lines of (of course completely untested):
```
ldapsearch -x -h localhost -b ou=users,dc=georchestra,dc=org dn |grep "^dn: uid=" | while read f ; do
    echo "$f\nchangetype:modify\nadd:objectClass\nobjectClass:georchestraUser\n-" >> modify.ldif
done

ldapsearch -x -h localhost -b ou=users,dc=georchestra,dc=org '(objectClass=organization)' dn |grep "^dn : o=" | while read f ; do
    echo "$f\nchangetype:modify\nadd:objectClass\nobjectClass:georchestraOrg\n-" >> modify.ldif
done

ldapmodify -D "cn=admin,dc=georchestra,dc=org" -W -f modify.ldif

```

fvanderbiest:
Looks indeed like a better way !
Thanks Landry

Issue #2848, opened 2019-12-02
[master] geowebcache - the build does not include the dependencies anymore


pmauduit:
**Version and function concerned by the bug**

Master (18.06 is not affected)

**Describe the bug**

A `mvn clean package` will generate a webapp which does not contain some libs (spring*).

**To Reproduce**

`mvn clean package` in the `geowebcache-webapp` directory, check in the geowebache.war the `WEB-INF/lib`, if there are only 2 files:

```
    testing: WEB-INF/lib/apache-log4j-extras-1.1.jar   OK
    testing: WEB-INF/lib/log4j-1.2.17.jar   OK
```

Then the webapp is likely to be not functioning. checking statically the 18.06 docker image, the webapp seems correct.

**Expected behavior**

The generated webapp contains every needed dependencies (all spring-* jars and so on)

PR #2644, opened 2019-06-25
CAS - unzip webapp in docker image

fvanderbiest:
```
jetty@f61670ded212:/var/lib/jetty/webapps/cas$ ls -al
total 32
drwxr-xr-x 6 jetty jetty 4096 Jun 25 13:52 .
drwxr-xr-x 1 jetty jetty 4096 Jun 25 13:52 ..
drwxr-xr-x 6 jetty jetty 4096 Jun 25 13:52 WEB-INF
drwxr-xr-x 2 jetty jetty 4096 Jun 25 13:52 css
-rw-r--r-- 1 jetty jetty  170 Jun 25 13:52 favicon.ico
drwxr-xr-x 2 jetty jetty 4096 Jun 25 13:52 images
-rw-r--r-- 1 jetty jetty 1069 Jun 25 13:52 index.jsp
drwxr-xr-x 2 jetty jetty 4096 Jun 25 13:52 js
```
Also reduces image size:
```
$ docker images | grep georchestra/cas
georchestra/cas   19.06-SNAPSHOT            286MB
georchestra/cas   19.04                     479MB
georchestra/cas   18.06                     478MB
```

PR #2849, opened 2019-12-02
docker - unzip webapps in the container

pmauduit:
This makes the customization (via volumes or so) easier.

inspired from what as already been done on CAS & SP, like the following PR:
https://github.com/georchestra/georchestra/pull/2644/files

Note: I encountered an issue while working on this, which might be worth waiting for a fix:
https://github.com/georchestra/georchestra/issues/2848

Tests:
Each module tested with `mvn clean package docker:build`, then manually checking that:

1. the webapp is unzipped in `/var/lib/jetty/webapps`
2. it is (almost) correctly launched (not taking care of db not ready or missing / unresolved variables, just that the webapp context remains correct).

### DIFF
--- a/analytics/pom.xml
+++ b/analytics/pom.xml
@@ -469,9 +469,8 @@
               <dockerDirectory>${project.basedir}/src/docker</dockerDirectory>
               <resources>
                 <resource>
-                  <targetPath>/var/lib/jetty/webapps</targetPath>
-                  <directory>${project.build.directory}</directory>
-                  <include>${project.artifactId}.war</include>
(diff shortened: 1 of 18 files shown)


pmauduit:
Shall we consider merging, knowing the fact that it won't address #2848 ?

pmauduit:
> Shall we consider merging, knowing the fact that it won't address #2848 ?

As discussed earlier, we can consider merging (once the checks are OK).

pmauduit:
Pending travis build is actually green, merging.

Issue #2826, opened 2019-11-21
Make deb-build-geoserver doesnt produce the same geoserver war as deb-build-georchestra


landryb:
Somewhere in the build system maze, probably in the Makefile,
 the two `deb-build-geoserver` and `deb-build-georchestra` dont provide the same geoserver war.

```
-rw-r--r-- 1 landry users 113877933 Nov 21 18:55 /tmp/geoserver-generic.war2
-rw-r--r-- 1 landry users  92207949 Nov 21 18:57 geoserver/webapp/target/geoserver-generic.war
```

the first war is generated by `make deb-build-geoserver` and the second by `make deb-build-georchestra` which seems to rebuild geoserver with default options, even if its already built.

dioffing their contents yields this:
```
[21/11 19:18] landry@build.thruxton:/data/src/georchestra/georchestra $diff -u /tmp/good.list /tmp/bad.list |grep ^-.*jar
-    54423  2018-03-19 10:06   WEB-INF/lib/commons-exec-1.3.jar
-   173113  2019-11-21 18:41   WEB-INF/lib/gs-monitor-core-2.15.0.jar
-  2035804  2019-11-21 18:29   WEB-INF/lib/jcodec-0.2.3.jar
-   156817  2019-11-21 18:28   WEB-INF/lib/gt-css-21.0.jar
-  2715721  2019-11-21 18:28   WEB-INF/lib/poi-4.0.0.jar
-   208484  2019-11-21 18:42   WEB-INF/lib/gs-importer-web-2.15.0.jar
-   415578  2018-03-19 10:06   WEB-INF/lib/commons-vfs2-2.0.jar
-    28756  2019-11-21 18:23   WEB-INF/lib/gt-jdbc-h2-21.0.jar
-   195418  2019-11-21 18:42   WEB-INF/lib/gs-csw-api-2.15.0.jar
-    21204  2019-11-21 18:23   WEB-INF/lib/geodb-0.9.jar
-    79823  2019-11-21 18:28   WEB-INF/lib/asm-util-6.2.1.jar
-     3888  2018-03-19 09:38   WEB-INF/lib/xlink-1.0-1.0.0-3.jar
-    25429  2018-03-19 10:06   WEB-INF/lib/regexp-1.3.jar
-    32871  2019-11-21 18:27   WEB-INF/lib/gt-csv-21.0.jar
-     9211  2019-11-21 18:41   WEB-INF/lib/gs-css-2.15.0.jar
[...]

PR #2850, opened 2019-12-04
Geoserver 2.16.1

pmauduit:
Upgrading to GeoServer 2.16.1. Some of the georchestra customizations have been kept:

* header integration
* patch to avoid a loop when slaves are notified that the GS datadir changed (which was modifying the datadir then creating an infinite loop)
* Changing default cached layer hashmap size from 100 elements to 400

The following have been dropped:

* possibility to override some jars in the lib subdirectory

Tests: only compilation with testsuite OK, runtime tested in a docker image (JRE11), UI launches fine, but I have not dug further for now. No GS modules tested so far either.


landryb:
fwiw, georchestra/mapstore2-georchestra also requires a recent geoserver for timeline and authkey support, cant remember the exact version but a recent 2.15 or even 2.16 would be nice !

fvanderbiest:
Excellent, thanks Pierre !
Can you also check (at least manually) that extensions we usually build are working fine ?
```
        "control-flow",
        "css",
        "gdal",
        "geofence",
        "inspire",
        "jp2k",
        "mbtiles",
        "monitor",
        "pyramid",
        "wps",
        "wps-download",
        "libjpeg-turbo",
        "vectortiles"
```

landryb:
and eventually look at #2826 ? ;)

pmauduit:
> Can you also check (at least manually) that extensions we usually build are working fine ?

These extensions are just maven profiles which add some extra dependencies available on the default maven repositories to the webapp. So, it is mainly a matter of checking if they are still available on the repo upstream and in the expected version (mvnrepository.com is my friend)

fvanderbiest:
Tested on a dev environment with the above mentioned plugins. 
Looks good to me but a second review would be nice.

groldan:
A few notes:

* At startup, the following error can be disregarded, guess it should be logged at the INFO level without stack trace from geoserver:
```
geoserver_1                 | Exception in thread "Thread-43" java.lang.NoSuchMethodError: 'com.sun.jna.Library com.sun.jna.Native.load(java.lang.String, java.lang.Class)'
geoserver_1                 | 	at oshi.jna.platform.linux.Libc.<clinit>(Libc.java:38)
geoserver_1                 | 	at oshi.software.os.linux.LinuxOperatingSystem.getProcessId(LinuxOperatingSystem.java:344)
geoserver_1                 | 	at org.geoserver.system.status.OSHISystemInfoCollector.lambda$new$0(OSHISystemInfoCollector.java:70)
geoserver_1                 | 	at java.base/java.lang.Thread.run(Unknown Source)
```

* Logs are not logged to the console, we should enable "log to stdout" by default. And probably _only_ log to the console, being in a dockerized environment.

landryb:
>     * Logs are not logged to the console, we should enable "log to stdout" by default. And probably _only_ log to the console, being in a dockerized environment.

please, no. Not everyone is running in a dockerized environment, and traditional-style has always been to log to a file.

fvanderbiest:
> We'd need to investigate further before upgrading to java 11.

Thanks Gabriel. 
We'll migrate to java 11 in a second time.

Issue #2482, opened 2019-02-08
All pom.xml - replace deprecated ${basedir} by ${project.basedir}


severo:
See https://maven.apache.org/ref/3.2.3/maven-model-builder/

pmauduit:
PR merged, closing the issue

PR #2854, opened 2019-12-11
s/basedir/project.basedir/ (#2482)

pmauduit:
`basedir` variable is deprecated, switching to `project.basedir` instead, see #2482 

Tests: compile-time on analytics OK resources are still copied.


fvanderbiest:
Seems that GeoNetwork is using `basedir` a lot too ...

PR #2861, opened 2019-12-24
console: avoid "connection_read" ldap server error logs with connection pooling

### LINKED ISSUE #2487: ldap - connection_read(19): no connection!
While playing with the console on master, the ldap container logs the following:
```
ldap_1                      | 5c5d72f2 connection_read(19): no connection!
ldap_1                      | 5c5d72f8 connection_read(19): no connection!
ldap_1                      | 5c5d7340 connection_read(19): no connection!
ldap_1                      | 5c5d7340 connection_read(19): no connection!
ldap_1                      | 5c5d7340 connection_read(19): no connection!
ldap_1                      | 5c5d7340 connection_read(19): no connection!
ldap_1                      | 5c5d7346 connection_read(19): no connection!
ldap_1                      | 5c5d7352 connection_read(19): no connection!
ldap_1                      | 5c5d7354 connection_read(19): no connection!
```

According to https://bugs.launchpad.net/ubuntu/+source/openldap/+bug/594840 this is because "It's normal for slapd to emit that message if a client disconnects without unbinding."

Should the console code be improved to unbind before disconnecting ?

groldan:
Fixes #2487

Using an LDAP connection pool as described in https://docs.spring.io/spring-ldap/docs/2.3.1.RELEASE/reference/#pooling

Solves the problem of having connections closed on the client without notifying the server and hence the immense ammount of `connection_read(19): no connection!` log entries in the ldap server.

The connection pool `destroy` method makes sure all connections are properly closed at server shutdown.

Should also improve performance since it avoids creating and destroying a connection for each ldap interaction.

### DIFF
--- a/console/src/main/webapp/WEB-INF/spring/webmvc-config.xml
+++ b/console/src/main/webapp/WEB-INF/spring/webmvc-config.xml
@@ -87,7 +87,7 @@
     <property name="dirContextValidator" ref="ldapContextValidator"/>
     <property name="testOnBorrow" value="${ldap.pool.testOnBorrow:true}"/>
     <property name="maxActive" value="${ldap.pool.maxActive:8}"/>
-    <property name="minIdle" value="${ldap.pool.minIdle:0}"/>
+    <property name="minIdle" value="${ldap.pool.minIdle:1}"/>
     <property name="maxIdle" value="${ldap.pool.maxIdle:8}"/>
     <property name="maxTotal" value="${ldap.pool.maxTotal:-1}"/>
     <property name="maxWait" value="${ldap.pool.maxWait:-1}"/>
--- a/console/src/main/webapp/WEB-INF/spring/webmvc-config.xml
+++ b/console/src/main/webapp/WEB-INF/spring/webmvc-config.xml
@@ -79,8 +79,18 @@
     <property name="password" value="${ldapAdminPassword}"/>
   </bean>
 
+  <bean id="ldapContextValidator" class="org.springframework.ldap.pool.validation.DefaultDirContextValidator">
+  </bean>
+  
   <bean id="contextSource" class="org.springframework.ldap.pool.factory.PoolingContextSource" destroy-method="destroy">
     <property name="contextSource" ref="singleContextSource"/>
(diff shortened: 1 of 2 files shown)


landryb:
untested yet but definitely nicer to the server (even if i dont see the messages on my setup which doesnt use docker..)

console/src/main/webapp/WEB-INF/spring/webmvc-config.xml
```diff
@@ -72,13 +72,27 @@
...
 
+  <bean id="ldapContextValidator" class="org.springframework.ldap.pool.validation.DefaultDirContextValidator">
+  </bean>
+  
+  <bean id="contextSource" class="org.springframework.ldap.pool.factory.PoolingContextSource" destroy-method="destroy">
+    <property name="contextSource" ref="singleContextSource"/>
+    <property name="dirContextValidator" ref="ldapContextValidator"/>
+    <property name="testOnBorrow" value="${ldap.pool.testOnBorrow:true}"/>
+    <property name="maxActive" value="${ldap.pool.maxActive:8}"/>
+    <property name="minIdle" value="${ldap.pool.minIdle:0}"/>
```
fvanderbiest: Shouldn't we always have at least one idle connection by default ?

groldan: I don't know, should we? to what end?

fvanderbiest: I'm not sure to fully understand how this works, but I'm tempted to think that having always at least one connection available in the pool will improve fetch time.

landryb: and reduce slapd log spam..

groldan: We can have an idle connection always ready, yet note we didn't before (e.g. no connection pool at all). I'll change it so the default is one by your request though.

groldan: done. No `connection_read(19): no connection!` in the ldap logs.

Issue #2693, opened 2019-08-23
[console] Organizations with accented char in the short name can cause charset issues


pmauduit:
* console can be unwilling to perform modifications
* Apache Directory Studio will fail in decoding the string (ArrayOutOfBoundException)

ldapvi is then the last resort:

![ca-trois-frontieres](https://user-images.githubusercontent.com/594335/63602283-5bc03e00-c5c7-11e9-9f5b-8cfed0597b68.png)

landryb:
Welcome to encoding hell :)

fvanderbiest:
I'm not sure what we should do here: prevent accents in Org short names, or clean them ?

jeanpommier:
What is exactly the purpose of Org short name ?

landryb:
a machine-readable short key, used in all the internal places/fields where you dont need a "pretty name" with spaces and accents and whatnot.

jeanpommier:
Thanks Landry. Well, if everybody agrees on this definition, I don't see any reason why we should accept accents there. It would be just looking for complications.

Additionally, if they are machine-readable for internal places only, do we need the capacity to change them afterwards ? I'm thinking about #2855

landryb:
here we settled for uppercase letters, numbers and underscore only. I'd like to see something like this (maybe not that strong) enforced for short names directly in the field validation, at least forbidding non-alphanumeric/spaces/accentued chars..

As for the capacity to change them, sadly i think it's needed b/c the user requesting the account fills it, and often puts non-meaningful values that we 'fix'/harmonize (CC_  & COM_ prefixes, etc...) wrt the existing orgs when we validate the account.

fvanderbiest:
> I'd like to see something like this (maybe not that strong) enforced for short names directly in the field validation, at least forbidding non-alphanumeric/spaces/accentued chars..

Agreed !

These forms should not allow non-alphanumeric/accentued chars for the Organisation short name:
 * `/console/account/new`
 * `/console/manager/#!/orgs/all?new=org`

PR #2865, opened 2020-01-02
[console] Org shortname validation

tonio:
see https://github.com/georchestra/georchestra/issues/2693


fvanderbiest:
This one does not cover `/console/account/new` IIUC.

tonio:
Good catch. I'll look at it.

Issue #2666, opened 2019-07-30
console - error on org creation without businessCategory


Gaetanbrl:
On master, i can create org without specify a business category. This action generate a weird behavior.

* I get a 500 error (PUT request) :
![image](https://user-images.githubusercontent.com/16317988/62133632-639a0480-b2df-11e9-9444-907edb37a311.png)
![image](https://user-images.githubusercontent.com/16317988/62133666-76143e00-b2df-11e9-9cff-1e0c6fddc6f8.png)

* This new org is create / accessible in the ldap with ldapsearch command, display in all orgs page but i can't read informations or do anything on this org from the console.
![image](https://user-images.githubusercontent.com/16317988/62134583-0a32d500-b2e1-11e9-80e9-84aa65df4784.png)

* this is slapd.log information : 
`RESULT tag=105 err=21 text=businessCategory: value #0 invalid per syntax`

* If i create a new org with business category, all seems to be better without errors.

About the business category, should not it be required with name and short name to create org ?
![image](https://user-images.githubusercontent.com/16317988/62134948-a957cc80-b2e1-11e9-81ce-dd8ebab89afc.png)

landryb:
businessCategory/orgType should be mandatory by default per https://github.com/georchestra/datadir/blob/19.04/console/console.properties#L110 - if it isnt, maybe something in the frontend doesnt properly handle  `requiredFields` now

Gaetanbrl:
Maybe i'm wrong... but i had  uncomment  _requireFields_ line in the console.properties, [and the expected type value](https://github.com/georchestra/georchestra/blob/19.04/console/src/main/webapp/manager/app/templates/orgForm.tpl.html#L27) was not return by the requireFields service. 

Currently, the requireFields service return any type value :
`https://georchestra.example.org/console/public/orgs/requiredFields`
```
[
    "name",
    "shortName"
]
```

If we manually set the next line in [OrgsController.java](https://github.com/georchestra/georchestra/blob/19.04/console/src/main/java/org/georchestra/console/ws/backoffice/orgs/OrgsController.java#L327), the type selection is required.
``` add fields.put("type"); ```

But this don't take account of requiredFields set in console.properties.

fvanderbiest:
I can confirm that /console/manager/#!/orgs/all?new=org does not mention that `businessCategory` is a redired field and it refuses to create the org as long this field is not filled.

fvanderbiest:
[...]

PR #2868, opened 2020-01-06
console - fix service publishing mandatory fields for orgs

RemiDesgrange:
Now web service that return the mandatory fields for orgs
(/public/requiredFields) depends on the Validation object. Which can be
overrided by properties.


Gaetanbrl:
> Cant test it at runtime right now but that seems to make sense - @Gaetanbrl can you check it fixes #2666 for you ?

I could check it now.

Gaetanbrl:
LGMT with this tests:

* orgType not mandatory -> OK for this PR.
I could create org without `orgType` but this action create an  error as explain in #2666.

* orgType mandatory -> OK.
Could not validate form without orgType.

Values from `console.properties` :
```
# Fields that MUST be filled in forms
# Possible values for org creation form: "orgAddress" and "orgType"
# default: firstName,surname,org,orgType
requiredFields=firstName,surname,org, orgType
```

Gaetanbrl:
@RemiDesgrange could you just change the name if this PR ? Thanks.

PR #2869, opened 2020-01-06
console - fix area persisting logic

### LINKED ISSUE #2471: [console] setting empty area for an org isnt persisted in the ldap
ok that's a corner case - setting an empty area for a given org does a PUT with a json containing `cities:[]` - that returns a 200 code, but the ldap isnt updated - i see the operation in the log, but the DIT isnt updated for some reason:
```
Feb  7 14:21:30 demo slapd[15919]: conn=31892 op=1 MOD dn="cn=CRAIG,ou=orgs,dc=craig,dc=fr"
Feb  7 14:21:30 demo slapd[15919]: conn=31892 op=1 MOD attr=description
Feb  7 14:21:30 demo slapd[15919]: conn=31892 op=1 RESULT tag=103 err=0 text=

$ $ldapsearch -x -b ou=orgs,dc=craig,dc=fr ou=CRAIG description
dn: cn=CRAIG,ou=orgs,dc=craig,dc=fr
description: 03155
```

I think it's because of this test : https://github.com/georchestra/georchestra/blob/master/console/src/main/java/org/georchestra/console/ds/OrgsDao.java#L296

i agree it's a fringe case, but the UI shouldnt let the admin believe that the change was persisted. Imo, the test should be removed, thus setting an empty `description` attribute ? @cmangeat ?

RemiDesgrange:
fix #2471 

Also:
* fix a little bug in checkAuthentication
* change the url of the default geojson area, since geopicardie one is
gone
* replace some good old java with his new style one.

### DIFF
--- a/console/src/main/java/org/georchestra/console/ds/OrgsDao.java
+++ b/console/src/main/java/org/georchestra/console/ds/OrgsDao.java
@@ -43,10 +43,10 @@ import javax.naming.directory.Attribute;
 import javax.naming.directory.Attributes;
 import java.io.IOException;
 import java.util.ArrayList;
-import java.util.Arrays;
 import java.util.Base64;
 import java.util.Collections;
 import java.util.List;
+import java.util.Objects;
 import java.util.regex.Pattern;
 import java.util.stream.Collectors;
 import java.util.stream.Stream;
@@ -141,7 +141,7 @@ public class OrgsDao {
                     } catch (DataServiceException e) {
                         return null;
                     }
-                }).filter(account -> null != account).map(account -> accountDao.buildFullUserDn(account))
+                }).filter(Objects::nonNull).map(account -> accountDao.buildFullUserDn(account))
                         .collect(Collectors.toList()).toArray(new String[] {}));
             }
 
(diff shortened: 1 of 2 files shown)


landryb:
>     * change the url of the default geojson area, since geopicardie one is
>       gone

fwiw this probably also needs changes in the datadir console config.. and i'd prefer if the commit was really atomic and only cared about fixing the issue, without all the syntaxic tweaks here and there  - but seems you java devs cant refrain from it ;)

RemiDesgrange:
About the default url for area you're right. But about the syntactic improvement, it's just improvement and we, as a community, need to strive doing our best to improve code quality bit by bits. 

I can remove the default area url if you want.

landryb:
> But about the syntactic improvement, it's just improvement and we, as a community, need to strive doing our best to improve code quality bit by bits.

dont get me wrong, i like it, it's just for future readability that having them as separate commits (in this PR or another, whichever) would be better, because your future self might not understand why those syntactic changes were in a commit fixing an actual issue :) and i think @fvanderbiest would agree with me here..

as for the default url, i think there was another one in a spring xml config like often but seems i'm wrong, so that's the only place its default value is defined. Granted, an ids admin *will* probably override it, but for the 'just works with the defaults' effect we should keep it.

fvanderbiest:
> dont get me wrong, i like it, it's just for future readability that having them as separate commits (in this PR or another, whichever) would be better, because your future self might not understand why those syntactic changes were in a commit fixing an actual issue :) and i think @fvanderbiest would agree with me here..

100% agree. A separate commit in this PR would have been the best option.
Next time, @RemiDesgrange will take care of it.

Issue #2431, opened 2019-01-28
[geonetwork] upgrade to 3.6


landryb:
Placeholder after #2198, as upgrading to a .0 is probably not a good idea - cf https://geonetwork-opensource.org/manuals/3.6.x/en/overview/change-log/version-3.6.0.html & http://osgeo-org.1560.x6.nabble.com/GeoNetwork-3-6-0-released-td5392645.html

landryb:
3.8.0 now .. https://geonetwork-opensource.org/manuals/3.8.x/en/overview/change-log/version-3.8.0.html

fvanderbiest:
3.8.2 nowadays

pmauduit:
My working branch for this update:
https://github.com/georchestra/geonetwork/tree/georchestra-gn3.8.2-master

Currently testing at runtime that nothing has been left out, and that the expected georchestra customizations are still working.

pmauduit:
Currently wondering if we could adress this issue at the same time:
https://github.com/georchestra/geonetwork/issues/90

pmauduit:
Commits that have been cherry-picked from georchestra-gn3.4-master to a 3.8.2 tag:

```

b359ca89d855a5c31b3c7e92bccbd4eceff0e40f JA get variables from datadir
7369547771cb006962b690e3a9166e1d6514841d PM remove jetty directory listing
ac68c71d9efab915180e7351eea2e2bcc565a160 JA fix variable name
3e9e1b393840adf590d49ccdac4a1517aafb02cb AA avoid trailing slash
7acb33385abf80ee0c7f7963754b58bb93844326 LB uploaded files private by default
e1e70339bac0063fe1cbe164c7fa2c27202eab6e FVB fix grid overflow
018cbc0b391a33977dabe217cedeb9e39e330f65 FVB GN messing with cors headers
baa7bd2fada72428b76b86abba9503512c20f5c4 CMT online src should be displayed as layer protocol -- SKIPPED ?
9f6a3d9945c5dde354b365234c8f86630dff2839 OGT improve map thumbnail                            -- NO IDEA what to be picked ?!
89d214de5cbe03201bf4495a2818cf56ecf35a2e LB jcs diskpath to a dire where Java can write
69656fb1bc54d373e06ea49e72b23ccb110aa3d7 OGT hides view mode in full view                     -- ??? using 3.8.2 version
2a46f5ba5cbd7535cf3b8e6fc0356f4ca5fd50f5 JP docker add --chown
[...]

PR #2871, opened 2020-01-07
geonetwork - update submodule

pmauduit:
Bump to version 3.8.2 (see #2431)

Tests: as far as I remember, utests & itests ok, some runtime tests (locally and onto dev.geograndest.fr/geonetwork)

Note: there are still at least one known issue about this version: the default configuration of the UI is broken and requires some fiddling around to be fixed (you will see 2 "full view" tab entry in the MD view), ping @cmangeat


landryb:
> Note: there are still at least one known issue about this version: the default configuration of the UI is broken and requires some fiddling around to be fixed (you will see 2 "full view" tab entry in the MD view), ping @cmangeat

that's what i'd call a nice blocker :) is there a tracking bug for it ? isnt it just a matter of yet again updating the default schema_plugins in the geonetwork default datadir ?

fxprunayre:
> the default configuration of the UI is broken

You should probably create a proper default UI config for geOrchestra in https://github.com/geonetwork/core-geonetwork/blob/master/web-ui/src/main/resources/catalog/js/CatController.js#L50-L285 as you don't want to use the default GN one (eg. use external viewer) and with that, at least all users can easily reset to the default from the admin:

image

and there is still room for improvements for updating this config during version updates combining existing with updates ...

Issue #2449, opened 2019-02-04
[console] tab titles should be routable urls without the middleware


landryb:
Added in 42b26032, fixed in 89c6a48 & d0c6c769, right now the 'users' tab isnt a routable url (the regexp doesnt match `browse/users` which should redirect to `browse/all/users`)and relying on the middleware/webserver to handle this is awkward - the webapp (or the sec-proxy?) should handle those rewrites/redirections so that things work ootb, without apache or nginx horrors...

right now (with the provided apache config), middle-clicking on the org console tab to open it in a new browser tab redirects to `/console/manager/#!/home#%2Forgs%2Fall` which just shows the console homepage, and not the org list. Same for the other tabs....

fvanderbiest:
Agreed.

fvanderbiest:
What do we miss here @tonio ?

severo:
I couldn't find how to configure the `angular-new-router` (see #2470).

Two options, I think:

- fix the URLs in a frontend proxy, as currently done
- update the router package to [ui-router](https://ui-router.github.io/ng1/) (to be checked: will it allow to solve our current issue?)
- update the app from [AngularJS](https://angularjs.org) to [Angular](https://angular.io) or another framework (React, Vue.js...)

severo:
Please comment on that issue before I keep working on it.

fvanderbiest:
JS app generates links like /console/manager/org/c2c/infos which are not server-side routable (hence cannot be opened in a new browser window)
```
$ curl -I http://localhost:8080/console/manager/org/c2c/infos
HTTP/1.1 404 Not Found
```

/console/manager/#!/org/c2c/infos is routable
```
$ curl -I 'http://localhost:8080/console/manager/#!/org/c2c/infos'
HTTP/1.1 200 OK
```

Modifying the frontend proxy to handle these rewrite is not the good way to go, as Landry mentions.
I think we need a custom `urlrewritefilter` config for the console, as for mapfishapp.

severo:
I understand that. What I propose is to replace

> JS app generates links like `/console/manager/org/c2c/infos` which are not server-side routable

by

> JS app generates links like `/console/manager/@!/org/c2c/infos` which are server-side routable

fvanderbiest:
okay !

fvanderbiest:
> if you think I should only fix the Apache configuration

I don't like this solution, but the doc is now fixed with c6926128129fd6a7252ed0ff8c134cb947e1403f
[...]

PR #2872, opened 2020-01-07
console - (re ?)introducing urlrewrite (#2449)

pmauduit:
See issue #2449 for the motivation.

Caveats: using path-info to get the request-uri without the context name
did not seem to work at runtime, so I used the request URI but with a
wildcard pattern on the webapp name (see in urlrewrite.xml). As a
consequence, "/console/manager/public/libraries.css" will match as well
as "/aaaa/bbbb/manager/public/libraries.css". But at least, it shall
allow people deploying in an other context name to still trigger the
rule accordingly.

tests:

* utests OK
* IT OK
* runtime tests OK


fvanderbiest:
Excellent, thanks !

PR #2875, opened 2020-01-10
Upgrade and restore geowebcache standalone functionality

### LINKED ISSUE #2848: [master] geowebcache - the build does not include the dependencies anymore

groldan:
Upgrade `geowebcache-webapp` to 1.16.1, being the same version used by the embedded gwc in geosewrver.

Fixes #2848

Depends on:
- georchestra/datadir#179
- georchestra/docker#49


geowebcache-webapp/README.md
```diff
@@ -0,0 +1,41 @@
...
+
+This is a geOrchestra specific re-packaging of GeoWebcache with customized extensions to
+
+- Handle authentication through geOrchestra's authentication HTTP request headers
+- Use a custom URL "mangler" for GWC to return URL's based on geOrchestra's public URL
+
+## Build
+
+```
+mvn clean install -f geowebcache-webapp
```
pmauduit: `-f` option expects an alternate pom.xml file, did you mean `-f geowebcache-webapp/pom.xml` or `-pl geowebcache-webapp` ? relying on the makefile modification above, I'd expect the second one.

geowebcache-webapp/README.md
```diff
@@ -0,0 +1,41 @@
...
+
+```
+make docker-build-geowebcache
+```
+
+## Integration approach
+
+Vanilla GWC `gwc-web` maven published artifact is actually a .war file packaged as .jar,
+containing all the web application libraries and resources, which turns it into a useless
+dependency as a jar.
```
pmauduit: And also could lead to classloader pollution if integrated in another war

geowebcache-webapp/README.md
```diff
@@ -0,0 +1,41 @@
...
+Geowebcache loads its main configuration from `WEB-INF/lib/geowebcache-servlet.xml`, which in turn
+imports a bunch of other `geowebcache-*.xml` spring beans configuration files at the same location.
+
+`geowebcache-servlet.xml` is configured in web.xml to be loaded by both `DispatcherServlet`
+and `ContextLoaderListener`, which results in double initialization of singleton beans.
+
+For this reason, we instead make `ContextLoaderListener` load from `applicationContext.xml`, which in
+turn includes `geowebcache-georchestra.xml`, which allows to add new bean definitions as well as to
+override any previously defined bean by name; and let `DispatcherServlet` load from an empty
+`geowebcache-servlet.xml`.
```
pmauduit: nice catch

geowebcache-webapp/README.md
pmauduit: Considering the pom modification below (moving the parent from org.georchestra.root to geowebcache), I think the -f would be more likely to work, as -pl is likely to work on children maven project ?

geowebcache-webapp/src/docker/Dockerfile
```diff
@@ -10,7 +10,7 @@ COPY --chown=jetty:jetty . /
 USER root
 
 RUN mkdir /mnt/geowebcache_tiles /mnt/geowebcache_datadir &&                              \
-    unzip -d /var/lib/jetty/webapps/geowebcache /var/lib/jetty/webapps/geowebcache.war && \
+    unzip -d /var/lib/jetty/webapps/geowebcache /var/lib/jetty/webapps/geowebcache*.war && \
     rm -f /var/lib/jetty/webapps/geowebcache.war &&                                       \
```
pmauduit: should it be something else than geowebcache.war in the docker image ? if adding a wildcard, why not on the `rm -f` statement ?

geowebcache-webapp/src/main/java/org/georchestra/geowebcache/security/PreAuthFilter.java
```diff
@@ -47,45 +49,60 @@
     public static final String SEC_USERNAME = "sec-username";
     public static final String SEC_ROLES = "sec-roles";
 
+    public PreAuthFilter() {
```
pmauduit: Not related to this PR, but I am pretty sure that something in spring-security already exists which does basically the same as this class

geowebcache-webapp/src/test/java/org/georchestra/geowebcache/GeorchestraURLManglerTest.java
```diff
@@ -0,0 +1,42 @@
+package org.georchestra.geowebcache;
+
+import static org.junit.Assert.assertTrue;
+
+import org.junit.Test;
+
+public class GeorchestraURLManglerTest {
+
+    @Test
+    public void testGeorchestraURLMangler() {
```
pmauduit: I am the original author of these tests but I am wondering by reading of them once again if they don't expose some flaws: ";" and multiple slashes in the URL might now be rejected by spring security filters.

geowebcache-webapp/src/docker/Dockerfile
groldan: good catch, I'm not sure how we handle or will handle the `-generic` et al suffix, hence the wildcard.

geowebcache-webapp/src/main/java/org/georchestra/geowebcache/security/PreAuthFilter.java
groldan: no freakin' idea TBH

geowebcache-webapp/src/test/java/org/georchestra/geowebcache/GeorchestraURLManglerTest.java
groldan: woot... dunno, open a separate issue?

geowebcache-webapp/src/docker/Dockerfile
groldan: Actually, removed the *, the less magic the better.

groldan:
@fvanderbiest @pmauduit rebased, fixed conflicts, fixed build, added to travis profile. It would be ready to go.

geowebcache-webapp/src/docker/Dockerfile
pmauduit: normally when the docker profile is activated, we should get rid of the "-generic" suffix (I think we should get rid of it in any cases, but I think it is more of historical, when we were still using configuration profile, and the generated webapps were suffixed with the ${server} variable)

Issue #2781, opened 2019-10-16
console - role names should be regexp-validated


fvanderbiest:
UPPERCASED, no special chars, no space ...
![Capture du 2019-10-16 14-23-09](https://user-images.githubusercontent.com/265319/66918906-9de28a80-f020-11e9-9835-c93d796c6149.png)

landryb:
yes please !!!!

fvanderbiest:
No milestone for now ...

fvanderbiest:
Something like https://github.com/georchestra/georchestra/commit/1046f9dce6ef80fb76b6e7fb99bf5973e0f608d2

PR #2897, opened 2020-01-23
[console] validate role shortname

tonio:
see #2781

### DIFF
--- a/console/src/main/webapp/manager/app/app.es6
+++ b/console/src/main/webapp/manager/app/app.es6
@@ -145,6 +145,13 @@ angular.module('manager', [
     moment.locale($translate.use())
     return date => moment(date).format('LL')
   }])
+  .directive('shortname', () => ({
+    require: 'ngModel',
+    restrict: 'A',
+    link: (scope, elm, attrs, ctrl) => {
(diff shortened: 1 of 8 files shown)


fvanderbiest:
Did I mention the org ? It was about the role IIRC.

fvanderbiest:
Also, the regexp should only allow uppercased chars (and underscores !) for the role name.

console/src/main/webapp/manager/app/app.es6
```diff
@@ -145,6 +145,18 @@ angular.module('manager', [
     moment.locale($translate.use())
     return date => moment(date).format('LL')
   }])
+  .directive('shortname', () => ({
+    require: 'ngModel',
+    restrict: 'A',
+    link: (scope, elm, attrs, ctrl) => {
+      const alphanum = v => v && v.match(/^\w*$/)
```
fvanderbiest: also need to allow the underscore char (ex: role `EL_TRUITE`)

tonio: https://developer.mozilla.org/fr/docs/Web/JavaScript/Guide/Expressions_r%C3%A9guli%C3%A8res#special-word :sunglasses:

Issue #2895, opened 2020-01-22
header - link to the GN admin console does not work 


pmauduit:
the lang parameter in the header is encoded on 2 characters, geonetwork in v3.8.x expects now the language to be 3 chars long. Hence the following link ends in 404:

https://platform/geonetwork/srv/en/admin.console

it should now be:

https://platform/geonetwork/srv/eng/admin.console

fvanderbiest:
Nice catch, thanks.

fvanderbiest:
Suggestion : in our fork, maintain the redirection `en -> eng`, `fr ->fra`, and so on ?
https://github.com/georchestra/geonetwork/blob/georchestra-gn3.8.2-master/web/src/main/webResources/WEB-INF/urlrewrite.xml

PR #2907, opened 2020-01-28
gn - update submodule (#2895)

pmauduit:
See #2895


pmauduit:
Sorry, I forgot that Dependabot should have proposed such a PR

PR #2909, opened 2020-01-29
update documentation

pmauduit:
documentation update:

* debian 10 (buster)
* tomcat9
* openjdk8 from adoptopenjdk (not available in buster)
* fiddling with systemd instead of regular initscripts
* documenting some tomcat specific behaviour (in regards to geoserver / wicket)
* adding a short explanation on how to get the generic wars from packages.g.o
* removing contents which were not relevant anymore

Tests: the guide has been followed from scratch into a vagrant debian buster VM. Of course, every features have not been tested thoroughly, but I finished with a running georchestra (login onto CAS / SP OK, geoserver along with integrated geofence OK, GN OK - MD editing tested ...).


docs/setup/apache.md
```diff
@@ -115,6 +115,10 @@ AddType application/vnd.ogc.context+xml .wmc
 
 ErrorDocument 502 /errors/50x.html
 ErrorDocument 503 /errors/50x.html
+
+# redirects onto the viewer by default
+RewriteRule ^/$ https://%{SERVER_NAME}/mapfishapp/ [R=301,L]
```
fvanderbiest: This is handled by the SP IIRC
https://github.com/georchestra/datadir/blob/6eb3dd3f3c09098804a1728de3aa493480a2ce4a/security-proxy/security-proxy.properties#L131-L134

docs/setup/tomcat.md
```diff
@@ -73,59 +78,96 @@ echo "" | openssl s_client -connect LDAPHOST:LDAPPORT -showcerts 2>/dev/null | o
...
+### Finally,
+
 verify the list of keys in keystore:
 ```
 keytool -keystore /etc/tomcat9/keystore -list
 ```
 
+## geOrchestra datadir
+
+Since the 15.12 version, geOrchestra allows its main configuration settings to be done into a specific directory, the "georchestra Datadir".
```
fvanderbiest: More than "allows", it is now required.

docs/setup/tomcat.md
```diff
@@ -73,59 +78,96 @@ echo "" | openssl s_client -connect LDAPHOST:LDAPPORT -showcerts 2>/dev/null | o
...
+
+By convention, we will consider that the path `/etc/georchestra` will be dedicated to this datadir. We can bootstrap a default datadir by
```
fvanderbiest: /can/have to/

docs/setup/apache.md
pmauduit: by default, I had a directory listing from apache2. If nothing handles the '/', apache2 will handle it.

But for sure, if you hit the SP directly, it will suggest a redirection to the default target.

docs/setup/tomcat.md
pmauduit: you're right, at the begining of the geor datadir, it was not mandatory though.

docs/setup/tomcat.md
pmauduit: strictly speaking, the datadir could be elsewhere (that is what we do for our customers), I'd prefer to stay less restrictive than "have to", maybe a future ("we will bootstrap") would be more appropriate.

docs/setup/tomcat.md
```diff
@@ -73,59 +78,96 @@ echo "" | openssl s_client -connect LDAPHOST:LDAPPORT -showcerts 2>/dev/null | o
...
-```
-... and replace it with:
-```
-# Provides:          tomcat-proxycas
-```
+Finally, edit the `/lib/systemd/system/tomcat9-proxycas` script and adapt to the created instance:
 
-In that same file, a few lines below :
-replace
 ```
```
fvanderbiest: diff syntax ?

docs/setup/tomcat.md
```diff
@@ -73,59 +78,96 @@ echo "" | openssl s_client -connect LDAPHOST:LDAPPORT -showcerts 2>/dev/null | o
...
 
+### Customize the Java runtime and options
 
-### Customize Java options
+In ```/etc/default/tomcat9-proxycas```, we will adapt the JAVA_HOME & JAVA_OPTS environment variables to suit our needs:
 
-In ```/etc/default/tomcat-proxycas```, we need to remove the ```-Xmx128m``` option: 
 ```
+JAVA_HOME=/usr/lib/jvm/adoptopenjdk-8-hotspot-amd64
 JAVA_OPTS="-Djava.awt.headless=true -XX:+UseConcMarkSweepGC"
```
fvanderbiest: GC availability/relevance to be double checked

docs/setup/tomcat.md
```diff
@@ -251,52 +334,24 @@ JAVA_OPTS="$JAVA_OPTS \
               -Djava.util.prefs.userRoot=/tmp \
               -Djava.util.prefs.systemRoot=/tmp"
 ```
-This allocates 2Gb of your server RAM to all geOrchestra webapps (except proxy, cas and geoserver).
+This allocates 2GB of your server RAM to all geOrchestra webapps (except proxy, cas and geoserver, which are located in other tomcat instances).
```
fvanderbiest: Probably insufficient nowadays ...

docs/setup/tomcat.md
```diff
@@ -305,7 +360,7 @@ JAVA_OPTS="$JAVA_OPTS \
               -Dgeonetwork.jeeves.configuration.overrides.file=/etc/georchestra/geonetwork/config/config-overrides-georchestra.xml"
 ```
 
-Note: You can also override every geonetwork sub-data-directories by modifying
+Note: You can also overide every geonetwork sub-data-directories by modifying
```
fvanderbiest: override

docs/setup/tomcat.md
```diff
@@ -454,19 +552,58 @@ JAVA_OPTS="$JAVA_OPTS \
...
+            -Ds3.properties.location=/etc/georchestra/geoserver/s3-geotiff/s3.properties \"
+```
+
+### Note for geofence users
+
+If geofence is activated in our geoserver, one will require some specific environment variables set. in the `/etc/default/tomcat9-geoserver0` file, consider adding:
+
+```
+JAVA_OPTS="$JAVA_OPTS \
+	      -DGEOSERVER_XSTREAM_WHITELIST=org.geoserver.geoserver.authentication.auth.GeoFenceAuthenticationProviderConfig \
```
fvanderbiest: I do not see this option in https://github.com/georchestra/georchestra/blob/master/geoserver/webapp/src/docker/Dockerfile

docs/setup/tomcat.md
```diff
@@ -454,19 +552,58 @@ JAVA_OPTS="$JAVA_OPTS \
...
+        -Dgeofence-ovr=file:/etc/georchestra/geoserver/geofence/geofence-datasource-ovr.properties"
+```
+
+### Note for geowebcache users
+
+GeoWebCache can be used either as a standalone webapp or integrated to GeoServer. If we want to go for the first option, we will also require the following JAVA_OPTS options:
+
+```
+	JAVA_OPTS="$JAVA_OPTS -DGEOWEBCACHE_CACHE_DIR=/opt/ "
```
fvanderbiest: /opt/ ?

docs/setup/tomcat.md
```diff
@@ -475,37 +612,39 @@ In ```/var/lib/tomcat-geoserver0/conf/server.xml```:
...
-<entry key="mapfishapp"    value="proxyDefaultTarget/mapfishapp/" />
-<entry key="static"        value="proxyDefaultTarget/header/" />""".replaceAll("\n|\t","").replaceAll("proxyDefaultTarget",proxyDefaultTarget)
-```
+Of course you will have to adapt the parameter to suit your setup. Setups based on the Jetty Servlet Container do not seem to be affected.
+
+
+## Be careful
+
+Remember that the geOrchestra binaries must be built according to the tomcat configuration described above.
+By default, forking the template configuration should guarantee this.
```
fvanderbiest: ???

pmauduit: That was already present in the documentation, even if it appears in the diff, I have not touched this part. We can probably get rid of the paragraph, though.

docs/setup/tomcat.md
fvanderbiest: I would propose 6 by default.

PR #2919, opened 2020-02-04
Docker images built on release published

fvanderbiest:
This PR:
 * builds LDAP and DB docker images on release published
 * prevents checking out useless submodules

Probably interesting to generalize to all other modules if it works as expected

### DIFF
--- a/.github/workflows/atlas.yml
+++ b/.github/workflows/atlas.yml
@@ -26,12 +26,9 @@ jobs:
     - name: "installing geOrchestra root pom"
       run: ./mvnw install --non-recursive
 
-    - name: installing georchestra-commons into .m2
-      run: ./mvnw install -P-all,atlas -pl commons
-
     - name: install & check formating
       working-directory: atlas/
-      run: ../mvnw --no-transfer-progress -B -Dfmt.action=validate install -Dadditionalparam=-Xdoclint:none -DskipTests
+      run: ../mvnw --no-transfer-progress -B -P-all,atlas -Dfmt.action=validate install -Dadditionalparam=-Xdoclint:none -DskipTests
 
     - name: run tests
       working-directory: atlas/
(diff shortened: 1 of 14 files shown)


.github/workflows/geoserver.yml
```diff
@@ -10,12 +10,15 @@ jobs:
     steps:
     - name: "Checking out"
       uses: actions/checkout@v1
+
     - name: "Checking out submodules"
       uses: textbook/git-checkout-submodule-action@2.0.0
```
RemiDesgrange: maybe we could clone only the submodule we need ? 

maybe `git submodule update geoserver/geoserver-submodule` ?

.github/workflows/atlas.yml
```diff
@@ -25,24 +25,30 @@ jobs:
...
+
     - name: installing georchestra-commons into .m2
-      run: ./mvnw install -pl commons
+      run: ./mvnw install -P-all,atlas -pl commons
+
     - name: install & check formating
       working-directory: atlas/
       run: ../mvnw --no-transfer-progress -B -Dfmt.action=validate install -Dadditionalparam=-Xdoclint:none -DskipTests
-    - name: run test
+
```
pmauduit: Don't know why, but my comment has been lost :-(

I think you can refactor this into only one action, as using `-P-all,atlas` will take advantage of maven dependency resolution process, that should figure out by itself that it has to compile commons first:

```
[get rid of root pom & georchestra-commons]

    - name: install & check formating
      run: ./mvnw --no-transfer-progress -B -P-all,atlas -Dfmt.action=validate install -Dadditionalparam=-Xdoclint:none -DskipTests
    
   - name: run test
[...]
```

.github/workflows/geoserver.yml
pmauduit: yes but still you'll have to deactivate the "all" module, so that you will exclude the modules which are not cloned (in your case here geonetwork):

https://github.com/georchestra/georchestra/blob/master/pom.xml#L893-L916

Hence the `-P-all,module-you-are-interested-in`

PR #2922, opened 2020-02-04
ldap - do not require openssh.ldif

fvanderbiest:
Most users will not benefit from this change. 
It might be a burden for them if we make it mandatory.

self-note: to port to master


fvanderbiest:
To port to master

PR #2928, opened 2020-02-06
remove unrelated docker images recipes & submodules

### LINKED ISSUE #2255: Question: do we need all the git submodules?
The https://github.com/georchestra/georchestra/blob/master/.gitmodules file currently contains links to the following projects:

## Forks

- https://github.com/georchestra/geonetwork
- https://github.com/georchestra/geoserver
- https://github.com/georchestra/extjs
- https://github.com/georchestra/geoext
- https://github.com/georchestra/openlayers
- https://github.com/georchestra/styler
- https://github.com/georchestra/geotools

I think we should explain why we forked these projects and use them as submodules. See https://github.com/georchestra/geonetwork/issues/96.

## External projects

https://github.com/camptocamp/docker_smtp is maintained by Camptocamp. Maybe it should be forked by the https://github.com/georchestra/ organization in order to maintain control on it as a community.

For https://github.com/probonogeek/extjs, I don't really understand why we use it, if we do not contribute to the project. As stated in https://git-scm.com/book/en/v2/Git-Tools-Submodules#_working_on_a_submodule: 

> It’s quite likely that if you’re using submodules, you’re doing so because you really want to work on the code in the submodule at the same time as you’re working on the code in the main project (or across several submodules). Otherwise you would probably instead be using a simpler dependency management system (such as Maven or Rubygems).

fvanderbiest:
These Dockerfiles will be moved to their own repositories

Closes #2255


fvanderbiest:
pushed to master too

Issue #2926, opened 2020-02-05
geoserver - actions - cannot create resource output directory


fvanderbiest:
https://github.com/georchestra/georchestra/commit/e79986dd52353953f6677418679c8be497afe652/checks?check_suite_id=437192065

[ERROR] Failed to execute goal org.apache.maven.plugins:maven-resources-plugin:2.6:resources (default-resources) on project gs-main: Cannot create resource output directory: /home/runner/work/georchestra/georchestra/geoserver/geoserver-submodule/src/main/target/classes -> [Help 1]

pmauduit:
Full stacktrace:
```
[INFO] ------------------------------------------------------------------------
[INFO] Reactor Summary:
[INFO] 
[INFO] Main Module 2.16.1-georchestra ..................... FAILURE [01:43 min]
[INFO] gs-restconfig 2.16.1-georchestra ................... SKIPPED
[INFO] GeoWebCache (GWC) Module 2.16.1-georchestra ........ SKIPPED
[INFO] Core UI Module (geOrchestra customized) 2.16.1-georchestra SKIPPED
[INFO] GeoServer 2.x root module 20.1-SNAPSHOT ............ SKIPPED
[INFO] geOrchestra GeoServer Web Application 20.1-SNAPSHOT  SKIPPED
[INFO] ------------------------------------------------------------------------
[INFO] BUILD FAILURE
[INFO] ------------------------------------------------------------------------
[INFO] Total time:  01:44 min
[INFO] Finished at: 2020-02-06T10:39:34Z
[INFO] ------------------------------------------------------------------------
[ERROR] Failed to execute goal org.apache.maven.plugins:maven-resources-plugin:2.7:resources (default-resources) on project gs-main: Cannot create resource output directory: /home/runner/work/georchestra/georchestra/geoserver/geoserver-submodule/src/main/target/classes -> [Help 1]
org.apache.maven.lifecycle.LifecycleExecutionException: Failed to execute goal org.apache.maven.plugins:maven-resources-plugin:2.7:resources (default-resources) on project gs-main: Cannot create resource output directory: /home/runner/work/georchestra/georchestra/geoserver/geoserver-submodule/src/main/target/classes
    at org.apache.maven.lifecycle.internal.MojoExecutor.execute (MojoExecutor.java:215)
    at org.apache.maven.lifecycle.internal.MojoExecutor.execute (MojoExecutor.java:156)
[...]

PR #2934, opened 2020-02-06
GS - gh actions - using git instead of the submodule action

pmauduit:
See #2926


pmauduit:
Well, the 2 lines in the README were just meant to trigger an gh action ...

Issue #2948, opened 2020-02-10
SP - improving log traces & HTTP handleRequest logic


pmauduit:
When the SP is unable to reach a target (defined in the targets-mapping.properties from the Datadir), we got the following trace in the logs:

```
Exception occured when trying to connect to the remote host
```

Adding some infos about which host  / url is causing isues is not a big deal / trivial:
https://github.com/georchestra/georchestra/blob/master/security-proxy/src/main/java/org/georchestra/security/Proxy.java#L712

We could add the `sUrl` parameter in the log message.

There is another issue with the log trace from the following method:

https://github.com/georchestra/georchestra/blob/master/security-proxy/src/main/java/org/georchestra/security/Proxy.java#L396-L403

It could be either `request.getServerName()` causing the Exception, as well as `InetAddress.getByName(url.getHost())`.

While digging the issues above, I also found out that in case of a DNS resolution error, the Async HTTP client was reporting the error quite fast (some ms), but we still get stuck for 5 minutes at the following statement waiting for an HTTP Request that won't never come:
https://github.com/georchestra/georchestra/blob/master/security-proxy/src/main/java/org/georchestra/security/Proxy.java#L813

fvanderbiest:
Status ?

pmauduit:
I can't remember but I think we can close it

PR #2949, opened 2020-02-10
Sp improvements (#2948)

pmauduit:
As described in #2948, this PR aims to solve the following "issues"

* When a URL cannot be retrieved, we have a meaningless "an error occured while trying to reach the remote host." with no info of which server/url were causing an issue
* When calling `isSameServer()`, the error logged in the try/catch might be wrong and report the wrong server as failing
* In case of unresolveable host in the target mappings, the future won't be detected as finished, and will block during 5 minutes, where it could return almost immediately.

Asking @cmangeat for a review, because of the brainstorm we had about these Future / asynchronous debates and which object should take care of reporting the error.

### DIFF
--- a/security-proxy/src/main/java/org/georchestra/security/Proxy.java
+++ b/security-proxy/src/main/java/org/georchestra/security/Proxy.java
@@ -397,7 +397,7 @@ public class Proxy {
         try {
             return InetAddress.getByName(request.getServerName()).equals(InetAddress.getByName(url.getHost()));
         } catch (UnknownHostException e) {
-            logger.error("Unknown host: " + request.getServerName());
+            logger.error(e.getMessage(), e);
             return false;
         }
     }
@@ -709,7 +709,8 @@ public class Proxy {
             doHandleRequest(finalResponse, proxiedResponse);
         } catch (IOException | ExecutionException | InterruptedException | TimeoutException e) {
             // connection problem with the host
-            logger.error("Exception occured when trying to connect to the remote host: ", e);
+            String errMsg = String.format("Exception occured when trying to connect to the remote url '%s'", sURL);
+            logger.error(errMsg, e);
             try {
                 finalResponse.sendError(HttpServletResponse.SC_SERVICE_UNAVAILABLE);
             } catch (IOException e2) {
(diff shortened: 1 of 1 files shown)


pmauduit:
And yes, it would be great to have some integration tests about it (like a header, and a non-existing target)

security-proxy/src/main/java/org/georchestra/security/Proxy.java
```diff
@@ -709,8 +709,15 @@ private void handleRequest(HttpServletRequest request, HttpServletResponse final
             doHandleRequest(finalResponse, proxiedResponse);
         } catch (IOException | ExecutionException | InterruptedException | TimeoutException e) {
             // connection problem with the host
-            logger.error("Exception occured when trying to connect to the remote host: ", e);
+            String errMsg = String.format("Exception occured when trying to connect to the remote url '%s'", sURL);
+            logger.error(errMsg, e);
             try {
+                if (e.getCause() instanceof java.net.UnknownHostException) {
```
cmangeat: one could catch UnknownHostException first ?

this would implies doing something in the https://stackoverflow.com/questions/3526167/exception-handling-try-catch-inside-catch/3526225 taste, i.e. declaring HandleTrySomethingFailure.

security-proxy/src/main/java/org/georchestra/security/Proxy.java
```diff
@@ -709,8 +709,15 @@ private void handleRequest(HttpServletRequest request, HttpServletResponse final
             doHandleRequest(finalResponse, proxiedResponse);
         } catch (IOException | ExecutionException | InterruptedException | TimeoutException e) {
             // connection problem with the host
-            logger.error("Exception occured when trying to connect to the remote host: ", e);
+            String errMsg = String.format("Exception occured when trying to connect to the remote url '%s'", sURL);
+            logger.error(errMsg, e);
```
cmangeat: I don't know why there is a debug(format, arg1, arg2, ...) and a debug(msg, throwable), and no debug(format, arg1, arg2, ..., throwable), but there is maybe a good reason.

pmauduit: don't know either

security-proxy/src/main/java/org/georchestra/security/Proxy.java
pmauduit: > one could also catch UnknownHostException first ?

I was thinking of something like that after having opened the PR, but we want to test the cause of the exception (e.getCause()), not the exception itself.

fvanderbiest:
Added the `backport` label, to keep in mind it should probably be backported to `20.0.x` branch after a while (probably between 20.0.0 and 20.0.1).

PR #2956, opened 2020-02-13
Add pull requests support to gh actions

RemiDesgrange:
Currently gh actions are triggers by push on branches only. With this PR
we also trigger actions on pull request (fork of georchestra).


fvanderbiest:
Nice.
To push to master + do also for GeoNetwork.
Thanks !

Issue #2965, opened 2020-02-20
[Header] - Special characters display incorrectly with OpenJDK 11


jusabatier:
Since Java9, properties files are read as UTF8, before they were read as ISO-8859-1.

So in the 20.x version which target Java11, all accents display incorrectly : d�connexion

The best solution seems to adapt this line : https://github.com/georchestra/georchestra/blob/20.0.x/header/src/main/java/org/georchestra/_header/Utf8ResourceBundle.java#L78

Another temporary one is to add : 
-Djava.util.PropertyResourceBundle.encoding=ISO-8859-1
in JAVA_OPTS

Source : https://community.microstrategy.com/s/article/Special-characters-display-incorrectly-with-OpenJDK-11?language=en_US

landryb:
what if ppl still use java8? we consider it's unsupported ?

jusabatier:
Tested with : 
https://github.com/georchestra/georchestra/pull/2966/commits/6266b728c3758c620ac9a81a43fe74d137fe35e9

And it works well with JAVA 8 and JAVA 11.

fvanderbiest:
Do you currently run geOrchestra OpenJDK 11 in production ? I'd be interested by your feedback

jusabatier:
Not in production yet, some modules of Georchestra doesn't support JAVA11.

Actually appear to work : 

- Security Proxy
- Header
- GeoServer
- Mapfishapp

Actually not work : 

- GeoNetwork, have a pending pull request to make it work : https://github.com/geonetwork/core-geonetwork/pull/4488

Actually not tested yet : 

- GeoWebCache
- Extractor
- Cadastrapp
- Atlas
- Analytics
- Console

fvanderbiest:
> * GeoServer

Actually, Gabriel had issues running it with Java 11 - recommended to stick to Java 8 until further notice

jusabatier:
Any link to the issue ?
I use it and actually not have any problems.

fvanderbiest:
Issues with shapefile stores IIRC. ping @groldan

PR #2966, opened 2020-02-21
[Header] - Correction for special characters display with OpenJDK 11

jusabatier:
Correction for https://github.com/georchestra/georchestra/issues/2965


pmauduit:
you also got some code from apache commons for testing such things:
https://commons.apache.org/proper/commons-lang/apidocs/org/apache/commons/lang3/SystemUtils.html

Unfortunately, it is available from commons-lang3 which is not available in the header's classpath, so I'm fine with parsing the version by hand ...

fvanderbiest:
pushed to master too

Issue #2983, opened 2020-03-10
console - profile endpoint is doing an extra (unecessary ?) LDAP request


pmauduit:
Looking at the endpoint `private/users/profile` in the console, this one is supposed to return something of the current form:

```json
{
  "uid": "pmauduit",
  "org": "c2c",
  "roles": [
    "ADMINISTRATOR",
    "SUPERUSER",
    "EXTRACTORAPP",
    "GN_ADMIN",
    "USER",
    "CADASTRE_CNIL1",
    "CADASTRE_CNIL2"
  ]
}
```

If we look at the code here: https://github.com/georchestra/georchestra/blob/master/console/src/main/java/org/georchestra/console/ws/backoffice/users/UsersController.java#L239-L275

We can see that (line 262) an extra request is made to look up the current connected user. I cannot see the point of querying the LDAP, as everything should be in the headers provided by the SP (org -> 'sec-org', uid -> 'sec-username', roles -> sec-roles.split(";")).

In addition, in case of mixing several users / roles datasources, CAS / SP can grant access to the console to an admin user who actually does not exist in the LDAP (coming from another datasource). In that case, the webservice will return a `404 - not found` response, breaking some UI aspects of the webapp.

PR #2984, opened 2020-03-10
console - using the http headers instead of LDAP (#2983)

pmauduit:
Instead of using an extra LDAP connection to retrieve the orgs and roles
of the current logged user, guess the infos from the HTTP headers
received from the security-proxy.

See https://github.com/georchestra/georchestra/issues/2983.

Tests:

* utests ok
* IT ok

Note:

Since the connection to an OpenLDAP is not necessary anymore, we don't
rely on it in the tests. That is the reason why I removed the IT which
was testing the profile endpoint, and migrated to some new unit tests
instead.


pmauduit:
Hmmm, actually we have to remove the "ROLE_" prefix from the headers ...

fvanderbiest:
> Hmmm, actually we have to remove the "ROLE_" prefix from the headers ...

Good catch ;-)

pmauduit:
runtime tested

Issue #2985, opened 2020-03-11
actions - remove the need for a M2_SETTINGS_FILE secret


fvanderbiest:
https://github.com/georchestra/georchestra/pull/2966/checks failed because forked repositories do not provide the `M2_SETTINGS_FILE` secret.
This file is currently only used to provide a maven proxy. It does not hold any secret.
As such, it should be versioned in the source code, or provided by any other means.

pmauduit:
It might also be interesting to make sure some of the actions are skipped if not run from the "official" repo (i.e. in a fork), like trying to connect onto docker-hub or all the docker push actions:
https://github.community/t5/GitHub-Actions/don-t-run-actions-on-pull-request-from-fork/td-p/45499

about the m2_settings_file variable, I think we could version it without the docker-hub credentials, as we always use "docker push" instead of "mvn docker:push" to push the images onto docker-hub, so the only interesting part remaining into this settings.xml file will be the url to the maven proxy (artifactory on packages.g.o) which is public.

PR #2987, opened 2020-03-11
Github actions - Activating extra profiles

pmauduit:
This PR actually aims to improve some of the github actions workflows:

1. removing the `M2_SETTINGS_FILE` secret: there is actually nothing secret about it (see #2985 )
2. limit docker stuff to the main geOrchestra repository: There is no need to consume CPU resources if we don't do anything with the docker image after the build. Following something similar to the previous issue: we likely to have no secrets configured on the fork to push the generated docker images.
3. Enable `sentry-spring` and `jsonevent-layout` m2 profiles when generating the official docker images (the actual purpose of this PR). The overhead on the generated artifact is less than 1MB, so even if it is not used by the final user, it seems to provide an interesting added value to the webapps, allowing to plug them to external tools (sentry, logstash/kibana and so on).

### DIFF
--- a/.github/resources/m2-settings.xml
+++ b/.github/resources/m2-settings.xml
@@ -1,0 +1,10 @@
+<?xml version="1.0"?>
+<settings xmlns="http://maven.apache.org/SETTINGS/1.0.0" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" xsi:schemaLocation="http://maven.apache.org/SETTINGS/1.0.0   http://maven.apache.org/xsd/settings-1.0.0.xsd">
+  <mirrors>
+    <mirror>
+      <id>artifactory-georchestra</id>
+      <mirrorOf>*</mirrorOf>
+      <url>https://packages.georchestra.org/artifactory/maven</url>
+    </mirror>
+  </mirrors>
+</settings>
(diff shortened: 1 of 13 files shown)


RemiDesgrange:
:+1: for me

fvanderbiest:
Probably a lighter syntax can be achieved with https://github.com/marketplace/actions/cancel-this-build

pmauduit:
> Probably a lighter syntax can be achieved with https://github.com/marketplace/actions/cancel-this-build

I think we still want the test / verify goals actions from maven being run in the forks. My strategy had been to basically "skip the step whenever it needs a secret". I'm curious to know what does the suggested action does exactly (which state will get the job in case of cancellation ?)

pmauduit:
> I'm curious to know what does the suggested action does exactly (which state will get the job in case of cancellation ?)

It's using the github actions API to cancel the job, so the job will get a `cancelled` status, which is not exactly what we are willing to do.

fvanderbiest:
Please also consider porting to https://github.com/georchestra/geonetwork/blob/georchestra-gn3.8.2-master/.github/workflows/geonetwork-docker.yml

fvanderbiest:
Should be backported to 20.0.x branch too

pmauduit:
> Please also consider porting to https://github.com/georchestra/geonetwork/blob/georchestra-gn3.8.2-master/.github/workflows/geonetwork-docker.yml

I cannot see the point of even activating the github-actions for forks on the georchestra/geonetwork repository: the tests are skipped.

pmauduit:
Also, there are no `sentry-spring` maven profile in the GeoNetwork fork

Issue #2990, opened 2020-03-12
actions - if statements with strings should be single-quoted


pmauduit:
The PR #2987 & #2989 introduced a bug in the way the workflow files are parsed: The doc explicitely states that single quotes must be used, we are currently using double quotes:

https://help.github.com/en/actions/reference/context-and-expression-syntax-for-github-actions#literals

"You must use single quotes. Escape literal single-quotes with a single quote."

PR #2991, opened 2020-03-13
gh actions - if statements string " -> ' (#2990)

pmauduit:
See #2990


fvanderbiest:
to be backported too ?

PR #2999, opened 2020-03-23
Remove init-method and use PostConstruct annotation

### LINKED ISSUE #2504: Replace Spring init-method configuration parameter by @PostConstruct annotation
It seems to be recommended to use the [@PostConstruct](http://docs.spring.io/spring/docs/current/spring-framework-reference/htmlsingle/#beans-postconstruct-and-predestroy-annotations) annotation in the classes, instead of the Spring XML configuration `init-method` parameter.

See reasons here: https://stackoverflow.com/a/1089691/7351594.

RemiDesgrange:
Close #2504. 

I also could not resist refactor this little lambda :-)

### DIFF
--- a/analytics/src/main/webapp/WEB-INF/ws-servlet.xml
+++ b/analytics/src/main/webapp/WEB-INF/ws-servlet.xml
@@ -9,7 +9,7 @@
 
     <context:annotation-config/>
 
-    <bean id="waitForDb" class="org.georchestra.commons.WaitForDb" init-method="test">
+    <bean id="waitForDb" class="org.georchestra.commons.WaitForDb">
       <property name="url" value="jdbc:postgresql://${pgsqlHost}:${pgsqlPort}/${pgsqlDatabase}"/>
       <property name="username" value="${pgsqlUser}"/>
       <property name="password" value="${pgsqlPassword}"/>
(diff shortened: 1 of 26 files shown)


fvanderbiest:
mfapp and sp do not build ?

RemiDesgrange:
See #3000

RemiDesgrange:
@fvanderbiest I'm going to wait for CI to work again, then I'll merge this.

fvanderbiest:
mapfishapp builds not, "ImportError: No module named virtualenv"
We might miss something like
```
    - name: setup Python 2.x (for jsbuild)
      uses: actions/setup-python@v1
      with:
        python-version: '2.x'

    - name: "installing jsbuild and virtualenv via pip"
      run: |
        pip install jsbuild
        pip install virtualenv
```
in https://github.com/georchestra/georchestra/blob/master/.github/workflows/mapfishapp.yml

landryb:
> mapfishapp builds not, "ImportError: No module named virtualenv"

i had do do something similar in cadastrapp as it also uses jsbuild (from last millenium), maybe you want to reuse it.

fvanderbiest:
> i had do do something similar in cadastrapp as it also uses jsbuild (from last millenium), maybe you want to reuse it.

Comes from there ;-)

RemiDesgrange:
The python problem is, indeed a problem but does not fail the build (dunno why), The real problem si the lack of a dependency which I need to figure out.

RemiDesgrange:
I forgot to rebase on master. Now OK

fvanderbiest:
Thanks !

PR #3005, opened 2020-03-25
Console - Display Org name instead of id

### LINKED ISSUE #2935: console - userdetails page displays org short name rather than full name
On /console/account/userdetails:
Capture du 2020-02-06 14-55-58

RemiDesgrange:
Fix #2935 - In the detail page (`/console/account/userdetails`) we
are now displaying the org name instead of the id (ex: Camptocamp
instead of c2c).

### DIFF
--- a/console/src/main/java/org/georchestra/console/ws/edituserdetails/EditUserDetailsFormController.java
+++ b/console/src/main/java/org/georchestra/console/ws/edituserdetails/EditUserDetailsFormController.java
@@ -133,7 +133,7 @@ public class EditUserDetailsFormController {
      *
      * @param account input data
      */
-    private EditUserDetailsFormBean createForm(final Account account) {
+    EditUserDetailsFormBean createForm(final Account account) {
 
         EditUserDetailsFormBean formBean = new EditUserDetailsFormBean();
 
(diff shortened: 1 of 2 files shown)


console/src/main/java/org/georchestra/console/ws/edituserdetails/EditUserDetailsFormController.java
```diff
@@ -133,7 +133,7 @@ private boolean isReferentOrSuperUser(Account userAccount) throws DataServiceExc
      *
      * @param account input data
      */
-    private EditUserDetailsFormBean createForm(final Account account) {
+    EditUserDetailsFormBean createForm(final Account account) {
```
pmauduit: ok for package-private access (so that the method is testable), you could also make it public and use the `@VisibleForTesting` annotation (which is only meant for doc purposes anyway, and only if guava is available in the classpath, I am not sure if this is the case for the console)

RemiDesgrange: I tested both. Hard to make a decision but prefer this one in the end :man_shrugging:

fvanderbiest:
To be ported to master too...

PR #3006, opened 2020-03-25
Add caching to build

### LINKED ISSUE #2979: [gh-actions] use cache action
geoserver did this in geoserver/geoserver#4008 (cf https://help.github.com/en/actions/configuring-and-managing-workflows/caching-dependencies-to-speed-up-workflows) and that might be useful to waste less resources:
```
    - name: Maven repository caching
      uses: actions/cache@v1
      with:
        path: ~/.m2/repository
        key: gs-${{ runner.os }}-maven-${{ hashFiles('**/pom.xml') }}
        restore-keys: |
          ${{ runner.os }}-maven-
    - name: Build with Maven
      run: mvn -B clean install -T2 -Prelease --file src/pom.xml
    - name: Remove SNAPSHOT jars from repository
      run: |
        find .m2/repository -name "*SNAPSHOT*" -type d | xargs rm -rf {}
```

RemiDesgrange:
closes #2979 - now testing with analytics only. Will add more web app and squash once
successful.

### DIFF
--- a/.github/workflows/analytics.yml
+++ b/.github/workflows/analytics.yml
@@ -29,6 +29,14 @@ jobs:
           mkdir -p $HOME/.m2
           cp .github/resources/m2-settings.xml $HOME/.m2/settings.xml
 
+    - name: "Maven repository caching"
+      uses: actions/cache@v1
+      with:
+        path: ~/.m2/repository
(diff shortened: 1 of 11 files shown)


fvanderbiest:
Thanks. Earth says thank you.
Same thing should happen for geonetwork too.
What about cherry-picking this to 20.0 branch too ?

PR #3011, opened 2020-03-27
Mapfishapp - Fix tests dependency fail introduceda while ago

RemiDesgrange:
it seems to be introduced by
https://github.com/georchestra/georchestra/commit/0da5782f3a891e183f0da1fe5c0a52d98d4de0e0#diff-acb8fe7e00154b6e6b35bb938c4c4bba
but ant was never a direct dep of the project, it was transitive, so harder to find.

### DIFF
--- a/mapfishapp/pom.xml
+++ b/mapfishapp/pom.xml
@@ -159,6 +159,12 @@
       <artifactId>mockito-all</artifactId>
       <scope>test</scope>
     </dependency>
+    <dependency>
+      <groupId>org.apache.ant</groupId>
+      <artifactId>ant</artifactId>
+      <version>1.10.7</version>
+      <scope>test</scope>
+    </dependency>
   </dependencies>
   <build>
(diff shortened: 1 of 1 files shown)


pmauduit:
> it seems to be introduced by
> [0da5782#diff-acb8fe7e00154b6e6b35bb938c4c4bba](https://github.com/georchestra/georchestra/commit/0da5782f3a891e183f0da1fe5c0a52d98d4de0e0#diff-acb8fe7e00154b6e6b35bb938c4c4bba)
> but ant was never a direct dep of the project, it was transitive, so harder to find.

Actually this is the recent update of urlrewritefilter from org.tuckey which does not depend anymore to ant:

https://github.com/georchestra/georchestra/pull/2975

see:
https://mvnrepository.com/artifact/org.tuckey/urlrewritefilter/3.0.4
vs
https://mvnrepository.com/artifact/org.tuckey/urlrewritefilter/4.0.4

where ant becomes optional.

PR #3028, opened 2020-04-06
Serve json area file locally.

### LINKED ISSUE #2599: console - communes_simplified.json should be served locally
https://github.com/georchestra/datadir/blob/b11671d25e3b4f214bad6b82380f342136f87687/console/console.properties#L132

RemiDesgrange:
Fix #2599

This commit creates a new controller `AreaController` which creates
the `console/area.json` endpoint. This controller will, based on the
`AreaUrl` property:

* redirect you to the proper json file if `AreaUrl` is an HTTP endpoint
* send you the JSON file if `AreaUrl` is a file located on the filesystem. It looks it up in the datadir automatically.

> Why doing an `area.json` instead of just sending the file ?

Doing this way allows for discovery in a cheap manner. Like convention over configuration.


console/src/main/java/org/georchestra/console/ws/AreaController.java
```diff
@@ -0,0 +1,82 @@
...
+    private String datadir;
+
+    /**
+     * serve a json file put in the datadir or redirect to a url if AreaUrl in the
+     * config is set to a url.
+     * 
+     * @return json or rediurect to the resource if area is an URL
+     * @throws IOException
+     */
+    @GetMapping(value = "/area.json", produces = MediaType.APPLICATION_JSON_UTF8_VALUE)
```
fvanderbiest, line 40: `/public/area.json` ?
So it falls into https://github.com/georchestra/datadir/blob/0010f8d609788267451fa12dd0583caf1f5598a0/security-proxy/security-mappings.xml#L11

console/src/main/java/org/georchestra/console/ws/AreaController.java
```diff
@@ -0,0 +1,82 @@
...
+import java.nio.file.Paths;
+import java.util.Arrays;
+
+/**
+ * A simple controller to serve an area.json file from your datadir.
+ */
+@Controller
+public class AreaController {
+
+    @Value("${AreasUrl:https://www.geo2france.fr/public/communes_zones_competence.geojson}")
```
fvanderbiest, line 27: I would point at a local resource like `/etc/georchestra/console/areas.geojson` instead

console/src/main/java/org/georchestra/console/ws/AreaController.java
RemiDesgrange: I thought we were now striving to not depend on the SP for securing the endpoint and let the app do it ? I didn't secure this endpoint (no security check), but I can change the ressource to `/console/public/area.geosjon`.

fvanderbiest: > I thought we were now striving to not depend on the SP for securing the endpoint and let the app do it ?

You're right, but it's not done yet. In addition, it helps keeping a kind of logic with existing services.

RemiDesgrange:
I am going to do a datadir PR too to add [this file](https://github.com/gregoiredavid/france-geojson/blob/master/regions-version-simplifiee.geojson) Do you agree ?

landryb:
> I am going to do a datadir PR too to add [this file](https://github.com/gregoiredavid/france-geojson/blob/master/regions-version-simplifiee.geojson) Do you agree ?

not sure this file matches what the console expects at runtime, especially with nesting/grouping between communes and departements.. cant take a random geojson found on the interwebs :)

landryb:
other than that this repo looks like a good source for the file, it just needs massaging with the expected attributes :)

fvanderbiest:
Asking for a second review from @pmauduit 
Thanks !

Regarding the data, I went through the process of simplifying https://www.data.gouv.fr/fr/datasets/contours-communes-france-administrative-format-admin-express-avec-arrondissements/ and will come up to something very soon

fvanderbiest:
I think we can use https://gist.github.com/fvanderbiest/92a69dbe7025f01f357ec70ae8b2d46e which has been obtained from https://www.data.gouv.fr/fr/datasets/contours-communes-france-administrative-format-admin-express-avec-arrondissements/ on 2020-07-01 and gone through the following:
```
ogr2ogr -f geojson -s_srs EPSG:3857 -t_srs EPSG:4326 -select INSEE_COM,NOM_COM_M,INSEE_DEP communes_4326.json COMMUNE.shp
```
Followed by [mapshaper](https://mapshaper.org) with `visvalingam / weighted area` algo, preventing shape removal, simplify parameter set to `0.05%`
And to finish with, [geojson-precision](https://github.com/jczaplew/geojson-precision) to further reduce the file size:
```
geojson-precision -p 4 communes_4326_simplified.json communes_4326_simplified_precision.json
```

The resulting file can be used with the following console options (will require a minor datadir update):
```
AreasKey=INSEE_COM
AreasValue=NOM_COM_M
AreasGroup=INSEE_DEP
```

landryb:
what size is the resulting file ?

landryb:
replying to myself.. 9Mb for complete france, which is huge. Keep in mind that's a file that's going to get sent to every client going on  the new account page, or its own account page, or an admin going to the org area tab in the console. gzip in the webserver/reverse proxy might help; but still...

landryb:
after filtering on AURA, that results in a 1.2Mb file, which is much better than the file i was using so far (2.5Mb)

RemiDesgrange:
9.2Mo

> what size is the resulting file ?

9.2Mo which is probably too large.

fvanderbiest:
It's indeed too large for a regular usage, but it's just meant to serve as an example / default setting.
I'm not fond of having a single region for equity reasons ;-)

landryb:
> It's indeed too large for a regular usage, but it's just meant to serve as an example / default setting.
> I'm not fond of having a single region for equity reasons ;-)

well, the default config has been shipping with picardie for the past years, so i dont think it really matters :) Imo, the default config should be ***light***, so maybe use corsica only, or keep france and go over the top with simplification ?

console/src/main/java/org/georchestra/console/ws/AreaController.java
```diff
@@ -0,0 +1,85 @@
...
+     * 
+     * @return json or rediurect to the resource if area is an URL
+     * @throws IOException
+     */
+    @GetMapping(value = "/public/area.geojson", produces = MediaType.APPLICATION_JSON_UTF8_VALUE)
+    @ResponseBody
+    public String serveArea(HttpServletResponse response) throws IOException {
+        // if arealUrl in config is an http endpoint, then send it directly.
+        if (isURL(areasUrl)) {
+            response.sendRedirect(areasUrl);
```
pmauduit, line 44: the file is meant to be loaded by the client via XHR ? Will it work if it points to whatever other server ?

fvanderbiest:
> or keep france and go over the top with simplification

Already simplified to the maximum ... cannot squeeze more bits !

console/src/main/java/org/georchestra/console/ws/AreaController.java
RemiDesgrange: If the url is located outside the georchestra domain, then a preflight request will be launch yes, but we cannot control that, we could download the geojson and send it, but we could miss an edge case, redirecting seems more clean to me. I think people have the choice here : host the file in the datadir or elsewhere.

console/src/main/java/org/georchestra/console/ws/AreaController.java
```diff
@@ -0,0 +1,85 @@
...
+ * A simple controller to serve an area.json file from your datadir.
+ */
+@Controller
+public class AreaController {
+
+    @Value("${AreasUrl:area.geojson}")
+    private String areasUrl;
+
+    @Value("${georchestra.datadir:/etc/georchestra}")
+    private String datadir;
```
pmauduit, line 30: If the datadir is not defined, then I'd use a default area.json nested in the webapp and serve it instead. But since we now consider that a georchestra datadir is mandatory, I wonder if it's not over-designing.

fvanderbiest: > I wonder if it's not over-designing.

Indeed.
Such a file is pure personalization : a useful default cannot be provided

RemiDesgrange:
Tests now use real geojson. 

See https://github.com/georchestra/datadir/pull/201 for the datadir PR. I choose Corsica for the area (since France entire country is a bit too heavy only for default purpose).

console/README.md
```diff
@@ -103,12 +103,19 @@ orgTypeValues=Association,Company,NGO,Individual,Other
...
-
-It defaults to:
+Features are loaded from a custom endpoint, `/public/area.geojson` which will:
+ * serve a GeoJSON file from the datadir
+    * From the filesystem (beware of rights on file)
+    * from the datadir roots folder
+    * from the console directory in datadir. 
+ * redirect to a URL if `AreaUrl` starts with `http`.
+ 
+The geojson FeatureCollection need to be in EPSG:4326 projection. The native SRS of the layer MUST be in EPSG:4326
```
fvanderbiest, line 113: has to be

pmauduit (approved):

RemiDesgrange:
So sorry I commited the `package-lock.json`

PR #3030, opened 2020-04-08
all webapps - remove sentry-spring / reintroduce sentry-log4j

pmauduit:
Quoting the sentry doc:

>A Sentry logging integration is more general and will capture errors
(and possibly warnings, depending on your configuration) that occur
inside or outside of a Spring controller. In most scenarios, using one
of the logging integrations instead of sentry-spring is preferred.

https://docs.sentry.io/clients/java/integrations/#spring

Which is actually what we did in the first place.

### DIFF
--- a/.github/workflows/analytics.yml
+++ b/.github/workflows/analytics.yml
@@ -52,7 +52,7 @@ jobs:
     - name: "Building docker image"
       if: github.repository == 'georchestra/georchestra'
       working-directory: analytics/
-      run: ../mvnw --no-transfer-progress clean package docker:build -Pdocker,log4j-logstash,sentry-spring -DdockerImageName=georchestra/analytics:${{ steps.version.outputs.VERSION }} -DskipTests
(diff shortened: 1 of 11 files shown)


RemiDesgrange:
What will we send to sentry then ? Is usage is for crash report/stack trace, I don't know if log4j integration will send it properly :thinking: I need to test this to understand.

pmauduit:
> What will we send to sentry then ? Is usage is for crash report/stack trace, I don't know if log4j integration will send it properly  I need to test this to understand.

Following the sentry doc I pointed and from my understanding, it is more relevant to use the logger to send things to sentry, as we should be able to get more "things" than from the Spring's HandlerExceptionResolver. I don't know the details either, but my feeling would be to follow Sentry's official doc.

From the log4j configuration, I think that introducing a sentry logger which would be configured on ERROR level should be ok. If we got too much things from there, the issue might not be the configuration, but the way we are logging things.

fvanderbiest:
to backport to 20.0.x

PR #3031, opened 2020-04-13
mapfishapp - prevent logical filters with a single term

fvanderbiest:
This PR fixes an issue in mapfishapp's CSW querier which could be ignored as long as geonetwork was not too strict about the incoming query filters.
It seems it's not the case anymore as of GeoNetwork 2.8+


fvanderbiest:
Pushed to master with 83a43a54cabe5e311e9daf8d02e18009cf07d439

PR #3050, opened 2020-04-29
[mapfishapp] - Fix mapfishapp wfs connection in sld classifier

RemiDesgrange:
It seems that recent (october 2019..) work to update geotools, breaks the SLD classifier when using WFS source.

### DIFF
--- a/mapfishapp/src/main/java/org/georchestra/mapfishapp/ws/classif/SLDClassifier.java
+++ b/mapfishapp/src/main/java/org/georchestra/mapfishapp/ws/classif/SLDClassifier.java
@@ -102,7 +102,7 @@ public class SLDClassifier {
         // "prefix:layername" pattern and we do not want to add an extra logic
         // to get the prefix URL in the code, we replace the character by hand.
 
-        _wfsngTypeName = command.getFeatureTypeName().replaceFirst(":", "_");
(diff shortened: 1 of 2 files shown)


RemiDesgrange:
It did not break unit test because unit test for this class was marked `ignored`. I'm marking them as not ignored, testing and submitting with proper unit test

RemiDesgrange:
Thanks to @groldan tests are now fixes.

Whole context: Following the update of `wfs-ng` test of the class `SLDClassifier` were not passing anymore. Those tests were marked as `Ignored` (not suppressed). At that time data wfs data passed to geotools needed to be slightly changed. With the recent update to geotool 21, those mysterious tricks are not needed anymore, and we can make the pass again :tada: 

Maybe need a backport to 20.0.x branch @fvanderbiest ?

fvanderbiest:
Thanks Remi. Can you port this to master now ?

fvanderbiest:
Oops, we now depend on (1.7Mb) https://geobretagne.fr/geoserver/wfs?REQUEST=GetCapabilities&VERSION=1.1.0&SERVICE=WFS being available for tests to pass.
And this does seem to be a good idea, if we consider https://github.com/georchestra/georchestra/runs/666094411

PR #3055, opened 2020-05-12
[mapfishapp] Fix tests

RemiDesgrange:
Following #3050. Whole GetCap from geobretagne WFS was 1.7Mo large. Tidying it a bit with
a smaller store.


fvanderbiest:
Thanks !
This looks like an improvement. Hopefully not too fragile.
In the future, can we imagine having static resources instead ?

RemiDesgrange:
We could mock the HTTP request, but it would have required a lot more work, So decided to stay with this solution even if mocking is the best option in the end.

fvanderbiest:
> Going to make the same PR on master

Or you can also cherry-pick. I'm fine with it, as long as it's done.

PR #3060, opened 2020-05-20
Console - fix bug where role cannot have hyphens.

RemiDesgrange:
Hyphens are ok on the backend side but not on the frontend side. Change
the validation logic in the js code allow a role to have hyphens. Ex:
GN_MY-GROUP-OF-USER is now valid.

### DIFF
--- a/console/src/main/webapp/manager/app/app.es6
+++ b/console/src/main/webapp/manager/app/app.es6
@@ -153,7 +153,7 @@ angular.module('manager', [
     restrict: 'A',
     link: (scope, elm, attrs, ctrl) => {
       const alphanum = v => v && v.match(/^[A-Z0-9-_]+$/)
-      ctrl.$validators.shortname = v => alphanum(v) && v.toUpperCase() === v
+      ctrl.$validators.shortname = alphanum
     }
   }))
 
--- a/console/src/main/webapp/manager/app/app.es6
+++ b/console/src/main/webapp/manager/app/app.es6
@@ -152,7 +152,7 @@ angular.module('manager', [
     require: 'ngModel',
     restrict: 'A',
     link: (scope, elm, attrs, ctrl) => {
-      const alphanum = v => v && v.match(/^\w*$/)
+      const alphanum = v => v && v.match(/^[a-zA-Z-_]+$/)
       let validator = alphanum
       if (scope.$eval(attrs.uppercased)) {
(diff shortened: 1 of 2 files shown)


fvanderbiest:
No server-side validation ?

RemiDesgrange:
The server side validation was already right see https://github.com/georchestra/georchestra/blob/9180d4bd627f3a1e0e1726aa2c456162a4358366/console/src/main/java/org/georchestra/console/ws/backoffice/roles/RolesController.java#L550

RemiDesgrange:
Ok I'm sorry. I don't know what I was thinking about. The regexp is false. This should be `/^[A-Z0-9-_]+$/` to not allow lowercase

console/src/main/webapp/manager/app/app.es6
```diff
@@ -152,7 +152,7 @@ angular.module('manager', [
     require: 'ngModel',
     restrict: 'A',
     link: (scope, elm, attrs, ctrl) => {
-      const alphanum = v => v && v.match(/^[a-zA-Z-_]+$/)
+      const alphanum = v => v && v.match(/^[A-Z0-9-_]+$/)
       let validator = alphanum
       if (scope.$eval(attrs.uppercased)) {
         validator = v => alphanum(v) && v.toUpperCase() === v
```
RemiDesgrange, line 158: @tonio does it still needed ?

tonio: Well the directive can be configured with `uppercased` attribute to check if case should be forced or not.
You may have to check where is this directive used…

tonio:
Doesn't server side misses the `^…$`  part ? (or there's some java magic at work)

RemiDesgrange:
I tested the java regexp with several edge cases (lowercase, number, emoji) and it fails (or success our test) in each of these cases.

fvanderbiest:
To be ported to master once merged.

fvanderbiest:
Seem like the port to master is incomplete @RemiDesgrange : https://github.com/georchestra/georchestra/commit/b0d1ff00711021249400230799a474f152553068 only made it to master

RemiDesgrange:
:facepalm: sorry...

RemiDesgrange:
Fixed in 8f3f9670b0cc922ea64454a6eca79504119ade6c

PR #3064, opened 2020-05-26
mapfishapp - relaxing case on owstype parameter

fvanderbiest:
GeoNetwork 3.8 configured (on `/geonetwork/srv/fre/admin.console#/settings/ui`) with an external viewer (mapfishapp) as such: `https://www.geograndest.fr/mapfishapp/?owsurl=${service.url}&layername=${service.name}&owstype=${service.type}Layer` fails to load layers due to GeoNetwork not adhering strictly to [mapfishapp's protocol](https://github.com/georchestra/georchestra/blob/master/mapfishapp/README.md#parameters).
Namely, the case differs for the `owstype` parameter.

This PR aims to relax mapfishapp's constraint on the `owstype` parameter to fix this compatibility issue.


fvanderbiest:
Ported to master with 59d0d57b5

PR #3065, opened 2020-05-28
all - maven repo cleanup

pmauduit:
* boundlessgeo repositories do not respond anymore
* osgeo now set up a nexus

tests: only a mvn clean install -DskipTests, the purpose here is just to make sure we can fetch the dependencies. It seems to pass with no modification on GN, but this submodule will probably require a cleanup as well.

PR #3070, opened 2020-05-29
GeoServer 2.17.0

pmauduit:
What's new in 2.17 ? https://speakerdeck.com/simboss/whats-new-in-geoserver-2-dot-17

compilation OK, tests OK, need to runtime test to make sure everything is OK.

### DIFF
--- a/security-proxy/src/main/java/org/georchestra/security/UrlFormEncodedPostFilter.java
+++ b/security-proxy/src/main/java/org/georchestra/security/UrlFormEncodedPostFilter.java
@@ -61,15 +61,9 @@ public class UrlFormEncodedPostFilter extends OncePerRequestFilter {
          */
         public WrappedRequest(HttpServletRequest request) {
             super(request);
-            String charset = request.getCharacterEncoding();
+
             try {
-                Charset.forName(charset);
-            } catch (Throwable t) {
-                charset = "UTF-8";
-            }
-            try {
-                String payload = IOUtils.toString(request.getInputStream(), charset);
-                this.content = payload.getBytes();
+                this.content = IOUtils.toByteArray(request.getInputStream());
             } catch (IOException e) {
                 Logger.error("Unable to extract body payload", e);
             }
--- a/security-proxy/src/main/java/org/georchestra/security/UrlFormEncodedPostFilter.java
+++ b/security-proxy/src/main/java/org/georchestra/security/UrlFormEncodedPostFilter.java
@@ -1,0 +1,90 @@
+package org.georchestra.security;
+
+import org.apache.commons.io.IOUtils;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+import org.springframework.http.HttpMethod;
+import org.springframework.web.filter.OncePerRequestFilter;
+
+import javax.servlet.FilterChain;
+import javax.servlet.ServletException;
+import javax.servlet.ServletInputStream;
+import javax.servlet.http.HttpServletRequest;
+import javax.servlet.http.HttpServletRequestWrapper;
+import javax.servlet.http.HttpServletResponse;
+import java.io.BufferedReader;
+import java.io.ByteArrayInputStream;
+import java.io.IOException;
+import java.io.InputStreamReader;
+import java.nio.charset.Charset;
+
+public class UrlFormEncodedPostFilter extends OncePerRequestFilter {
+    public void destroy() {
+    }
+
+    @Override
+    protected void doFilterInternal(HttpServletRequest request, HttpServletResponse response, FilterChain filterChain)
+            throws ServletException, IOException {
+
+        if (request.getMethod().equalsIgnoreCase(HttpMethod.POST.name()) && isFormContentType(request)) {
+            // Morph the request to a WrappedRequest, as this will
+            // allow us to read the body content in the Proxy.java class once again.
+            request = new WrappedRequest(request);
+        }
+        filterChain.doFilter(request, response);
+    }
+
+    /**
+     * Returns if the request is a POST x-www-form-urlencoded or not.
+     *
+     * @return true if this is the case, else false.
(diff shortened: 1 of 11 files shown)


pmauduit:
Note: it will require https://github.com/georchestra/georchestra/pull/3065 because even if the maven repo for georchestra has been reconfigured to resolve to the new osgeo repo, somehow maven still fails because of boundlessgeo repository not responding anymore.

pmauduit:
There is (yet another) issue with the way the SP is making requests on behalf on the informations carried by what the servlet container provides us via the HttpServletRequest object:

https://github.com/georchestra/georchestra/blob/master/security-proxy/src/main/java/org/georchestra/security/Proxy.java#L967-L977

the getParameters*() - at least in case of Jetty, but I am pretty sure Tomcat reacts the same way - returns every parameters, either they come from the request body (POST) or via the query string (GET). In the general case, that does not cause any issues, but it seems that with geoserver 2.17, a parameter "filter=false" is passed via GET request, and we also have a "filter=pattern" POSTed when trying to filter on the layers' view.

* I tried to send the body as it is coming (i.e. merging the if / else in the previous code snippet), but no luck: an empty POST is sent. I suspect the InputStream to be already parsed before we can get a hand on it. Maybe we could try to rewind it ?
* I also tried the @RequestBody from Spring, hoping that it will provide the payload as it is received by the servlet container, but it also contains the unexpected "filter=false" parameter (chrome dev tools proves that this is not posted to the SP, so it should not appear here, it would be worth looking at how spring generates the variable).

pmauduit:
Tested so far:

* GN: administration / adding MD schemas & samples, creating a new MD, modifying several fields, sending a overview picture, upload of a file OK
* Mapfishapp: loading a geojson file OK

pmauduit:
There might be a more "spring" approach actually: https://www.baeldung.com/spring-reading-httpservletrequest-multiple-times#:~:text=HttpServletRequest%20is%20an%20interface%20which,can%20be%20read%20only%20once.

pmauduit:
> There might be a more "spring" approach actually: https://www.baeldung.com/spring-reading-httpservletrequest-multiple-times#:~:text=HttpServletRequest%20is%20an%20interface%20which,can%20be%20read%20only%20once.


well, it seems to do basically the same actually. The only difference I notice is that it makes use of spring filters instead of plain servlet filters.

pmauduit:
I changed my mind: if we can use spring, then I'd like to go for it. I'd like to rework this PR to try the ContentCachingRequestWrapper class.

pmauduit:
After several tries, it seems that something is consuming the inputstream before I can get a hand on it. I'll stick to my version if I can't get it work.

security-proxy/src/main/java/org/georchestra/security/UrlFormEncodedPostFilter.java
```diff
@@ -0,0 +1,90 @@
...
+        public WrappedRequest(HttpServletRequest request) {
+            super(request);
+            String charset = request.getCharacterEncoding();
+            try {
+                Charset.forName(charset);
+            } catch (Throwable t) {
+                charset = "UTF-8";
+            }
+            try {
+                String payload = IOUtils.toString(request.getInputStream(), charset);
```
pmauduit, line 71: decoding the byte array might not be necessary, we could read it as it is.

fvanderbiest:
@pmauduit would you like to do some more final checks before this is merged ?
It goes to master (future 20.1.0), so I think it's safe.

fvanderbiest:
This PR may now require an updated web.xml as per https://github.com/georchestra/georchestra/pull/3094/files#diff-fadc98b3758d33e5ac734727b7b9ab08

fvanderbiest:
Shoud we also change the base image to `jetty:9-jre11` for the docker image ?
https://docs.geoserver.org/stable/en/user/production/java.html#use-supported-jre

pmauduit:
> Shoud we also change the base image to `jetty:9-jre11` for the docker image ?

This will require an upgrade of one of the dependencies from the root pom, see: https://github.com/georchestra/georchestra/issues/3130#issuecomment-679909644 (commons-lang3 from version 3.7 to 3.8.1).

pmauduit:
Still missing:
* [] bump commons-lang3 version from 3.7 to 3.8.1 in the geOrchestra root pom
* [] "forward port" of georchestra/geoserver#20

pmauduit:
> This PR may now require an updated web.xml as per https://github.com/georchestra/georchestra/pull/3094/files#diff-fadc98b3758d33e5ac734727b7b9ab08

that might require to add jetty-servlets to the dependencies, as it won't probably be available inside a tomcat servlet container. And in that case, we have to make sure the version we will bundle remains compatible with the jetty version used as base docker images ...

I would prefer leave it as it is, and people willing to activate CORS can edit their web.xml, add the needed dependencies depending on the servlet container used, ... as it is already documented here: https://docs.geoserver.org/latest/en/user/production/container.html

Sorry, there is no way of tackling this issue easily, because it depends on implementation choices.

pmauduit:
About this CORS integration, I did a rapid test building a generic webapp with the extra dependency to jetty-servlets, it seems to work on both servlet containers (jetty & tomcat) at runtime, but I'm not fan of mixing the servlet container version, with a fixed version of jetty-servlets in the webapp. If we really want CORS enabled, maybe we shall take the risk.

fvanderbiest:
> If we really want to enable CORS, maybe we shall take the risk.

I think it is worth it, yes.

fvanderbiest:
Rebase 2.17.2 ?

pmauduit:
Currently giving a shot on a 2.17.2 instead, see https://github.com/georchestra/georchestra/tree/gs-2.17.2-update

pmauduit:
Closing, in order to create a new one based on GS 2.17.2

PR #3073, opened 2020-06-05
mapfishapp - SLD builder - fix knownSymbol

fvanderbiest:
Proposal for #3072 kudos to @groldan for the hint

PR #3074, opened 2020-06-05
Classification now works on the frontend side and backend side

RemiDesgrange:
This PR will be ported to 20.0.x branch.


fvanderbiest:
See #3073

PR #3079, opened 2020-06-10
mapfishapp - fix GFI on QGIS layers

fvanderbiest:
Issue arose while querying a QGIS server (namely https://qgisserver.hautsdefrance.fr/cgi-bin/qgis_mapserv.fcgi?MAP=%2Fvar%2Fwww%2Fdata%2Fqgis%2Fapplications%2Fdev_eco.qgs&SERVICE=WMS&REQUEST=GetCapabilities).

Features got from parser had no layer identifier other than the fid.
As a result, mapfishapp's default behavior is now to use the fid to group features by layer name.


fvanderbiest:
Ported to `20.0.x` with f14edbfdc93f492eb2f4501bd82e324548870de8

PR #3082, opened 2020-06-16
[mapfishapp]fixing ci

RemiDesgrange:
It seems that adding `System.setProperty("org.geotools.referencing.forceXY", "true")`
before tests does nothing in maven (work in ide for example since it
launch the jvm directly) so in order for the test to pass, I'm adding
`forceXY` as args to maven.

### DIFF
--- a/.github/workflows/mapfishapp.yml
+++ b/.github/workflows/mapfishapp.yml
@@ -45,7 +45,7 @@ jobs:
 
     - name: "Running tests"
       working-directory: mapfishapp/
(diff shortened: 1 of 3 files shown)


fvanderbiest:
Please backport also to 20.x

RemiDesgrange:
done in 20.0.x branch too : https://github.com/georchestra/georchestra/commit/7cdfca8a77cbde4aa386351e4999e1d6cbe479cf

PR #3088, opened 2020-06-26
mfapp - improve logging

pmauduit:
* Avoid calls to printStackTrace() use logger with error level instead
* Better error feedback in case connecting to WFS ends up with an error
  because of bad credentials (401)

Tests:

* testsuite OK

Note: We should consider using slf4j instead of apache commons, but since the mfapp webapp is almost considered EOL ...


mapfishapp/src/main/java/org/georchestra/mapfishapp/ws/A_DocService.java
```diff
@@ -580,9 +580,9 @@ private String loadContent(final String fileName) {
                 content = new String(bytes);
 
             } catch (FileNotFoundException fnfExc) {
-                fnfExc.printStackTrace();
+                LOG.error("file not found", fnfExc);
```
fvanderbiest, line 583: File with a capital F for the sake of coherence with the other log messages ;-)

fvanderbiest:
LGTM, thanks !
Would deserve a backport to 20.0.x ...

fvanderbiest:
89f947249 on `20.0.x`

Issue #3084, opened 2020-06-23
geoserver - enable CORS by default


fvanderbiest:
See https://docs.geoserver.org/latest/en/user/production/container.html#enable-cors
Nearly every SDI needs it ... makes sense to have it ON by default I think.

landryb:
in my experience, toggling/uncommenting those bits in web.xml were not enough to enable CORS for geoserver, i ended up doing it in the nginx reverse-proxy:
```
add_header "Access-Control-Allow-Origin" "*";
add_header "Access-Control-Allow-Headers" "Origin, X-Requested-With, Content-Type, Accept";
```

didnt really understood what was wrong, so i'm interested in a solution.

RemiDesgrange:
And when I see the [implementation](https://github.com/eclipse/jetty.project/blob/jetty-10.0.x/jetty-servlets/src/main/java/org/eclipse/jetty/servlets/CrossOriginFilter.java), I cannot see anything about handling OPTIONS request...

RemiDesgrange:
After some investigation:

* To support CORS we need: 
  * preflight support (HTTP OPTIONS request)
  * `Access-Control-*` headers
* It can happen:
  * before georchestra in your apache/nginx/haproxy/traefik/whatever. But some of these software cannot respond to preflight request, because it's not their job.
  * in the security-proxy, by creating a new filter. This could be tricky because you'll want it to be highly configurable (like I accept CORS on geoserver but not on the console)
  * Code it in  every webapp.

PR #3093, opened 2020-07-01
Proxy all HTTP verb

RemiDesgrange:
This PR add explicitly all HTTP verb and allow the proxy to forward
all types of HTTP verb.

Also handle the case of HTTP PATCH which was not handled by proxy.

This PR was primarily made to support forwarding of CORS pre-flight request. and to address part of #3084

### DIFF
--- a/security-proxy/src/main/java/org/georchestra/security/Proxy.java
+++ b/security-proxy/src/main/java/org/georchestra/security/Proxy.java
@@ -20,9 +20,14 @@
 package org.georchestra.security;
 
 import static org.springframework.web.bind.annotation.RequestMethod.GET;
+import static org.springframework.web.bind.annotation.RequestMethod.HEAD;
+import static org.springframework.web.bind.annotation.RequestMethod.OPTIONS;
 import static org.springframework.web.bind.annotation.RequestMethod.POST;
+import static org.springframework.web.bind.annotation.RequestMethod.PUT;
+import static org.springframework.web.bind.annotation.RequestMethod.PATCH;
+import static org.springframework.web.bind.annotation.RequestMethod.DELETE;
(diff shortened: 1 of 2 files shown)


fvanderbiest:
Looks good, thanks ! 
I think it will be safe to backport to `20.0.x`

security-proxy/src/main/java/org/georchestra/security/Proxy.java
```diff
@@ -364,7 +370,8 @@ public void handleUrlParamRequest(HttpServletRequest request, HttpServletRespons
      * sometimes used by the underlying webapps (e.g. mapfishapp and the mfprint
      * configuration). hence we need to allow it in the following "params" array.*
      */
-    @RequestMapping(value = "/**", params = { "!login" })
+    @RequestMapping(value = "/**", params = { "!login" }, method = { GET, POST, HEAD, OPTIONS, PUT, PATCH, DELETE,
+            TRACE })
```
pmauduit, line 374: I still cannot get why it is necessary to pass explicitly the methods that should match, it should already be the default ones:
https://docs.spring.io/spring/docs/4.3.22.RELEASE/javadoc-api/org/springframework/web/bind/annotation/RequestMapping.html#method--

But I do trust you anyway, if it's necessary then we should go for it

RemiDesgrange: I don't understand it neither... not enough spring guru to find the actual interface implementation

RemiDesgrange:
Backported to 20.0.x branch: https://github.com/georchestra/georchestra/commit/fbe51b207c65a2f4af6f15e0eee4b45d25cef32f

PR #3094, opened 2020-07-02
Enable CORS request in geoserver

RemiDesgrange:
It includes: preflight request (HTTP options method request) and
`Allow-Origin-*` headers (present in all requests). `web.xml` has been
extracted from the previous geoserver image and parts corresponding to
CORS have been uncommented.

### DIFF
--- a/geoserver/webapp/src/docker/Dockerfile
+++ b/geoserver/webapp/src/docker/Dockerfile
@@ -26,6 +26,17 @@ RUN wget https://downloads.sourceforge.net/project/libjpeg-turbo/2.0.4/libjpeg-t
 RUN mkdir /mnt/geoserver_datadir /mnt/geoserver_geodata /mnt/geoserver_tiles /mnt/geoserver_native_libs && \
     chown jetty:jetty /etc/georchestra /mnt/geoserver_datadir /mnt/geoserver_geodata /mnt/geoserver_tiles /mnt/geoserver_native_libs
 
+# add a tweaked configuration. First use case is support of OPTIONS.
+ADD web.xml /var/lib/jetty/webapps/geoserver/WEB-INF/web.xml
+
+# This is needed to activate the CrossOriginFilter in web.xml
+RUN ln -s /usr/local/jetty/lib/jetty-servlets-*.jar        \
+        /usr/local/jetty/lib/jetty-continuation-*.jar   \
+        /usr/local/jetty/lib/jetty-http-*.jar           \
+        /usr/local/jetty/lib/jetty-util-*.jar           \
+        /usr/local/jetty/lib/jetty-io-*.jar             \
+        /var/lib/jetty/webapps/geoserver/WEB-INF/lib/
+
 USER jetty
 
 VOLUME [ "/mnt/geoserver_datadir", "/mnt/geoserver_geodata", "/mnt/geoserver_tiles", "/mnt/geoserver_native_libs", "/tmp", "/run/jetty" ]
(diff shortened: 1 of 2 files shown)


fvanderbiest:
This LGTM, thanks !

We have to keep in mind that `web.xml` may change from one geoserver version to another (eg switching to 2.17 currently with #3070 ) and this will be our job to maintain it.

fvanderbiest:
> for the other usages (debian packages, wars and so on), it will require some extra configuration.

This is documented [here](https://docs.geoserver.org/latest/en/user/production/container.html#enable-cors)

geoserver/webapp/src/docker/Dockerfile
```diff
@@ -26,6 +26,17 @@ RUN wget https://downloads.sourceforge.net/project/libjpeg-turbo/2.0.4/libjpeg-t
 RUN mkdir /mnt/geoserver_datadir /mnt/geoserver_geodata /mnt/geoserver_tiles /mnt/geoserver_native_libs && \
     chown jetty:jetty /etc/georchestra /mnt/geoserver_datadir /mnt/geoserver_geodata /mnt/geoserver_tiles /mnt/geoserver_native_libs
 
+# add a tweaked configuration. First use case is support of OPTIONS.
+ADD web.xml /var/lib/jetty/webapps/geoserver/WEB-INF/web.xml
+
+# This is needed to activate the CrossOriginFilter in web.xml
+RUN ln -s /usr/local/jetty/lib/jetty-servlets-*.jar        \
```
fvanderbiest, line 33: Rather than symlinking, did you check what @MaxiReglisse proposed [here](https://gis.stackexchange.com/posts/277798/revisions) ?

landryb:
> > for the other usages (debian packages, wars and so on), it will require some extra configuration.
> 
> This is documented [here](https://docs.geoserver.org/latest/en/user/production/container.html#enable-cors)

as i said in https://github.com/georchestra/georchestra/issues/3084#issuecomment-648043681, just following the geoserver doc wasnt enough to get CORS working (but maybe i tested wrong).

If you want to properly fix #3084, please also test/adapt the default apache configuration that is documented in the repo... or explicitely state that non-docker configs are now on their own, and 2nd-class citizens :)

fvanderbiest:
> or explicitely state that non-docker configs are now on their own, and 2nd-class citizens :)

They're definitely not !

RemiDesgrange:
Ported on master:   https://github.com/georchestra/georchestra/commit/b63804a86d3efc6014c20e96da21ba379c8f0395

pmauduit:
> (but maybe i tested wrong)

My understanding is that it requires jetty-servlets (and its transitive dependencies) in your webapp classpath:
https://mvnrepository.com/artifact/org.eclipse.jetty/jetty-servlets/9.4.30.v20200611

note that jetty-servlets (https://www.eclipse.org/jetty/javadoc/current/org/eclipse/jetty/servlets/package-summary.html) is "just" a useful set of objects (servlets, filters, ...) implementations of `javax.servlets.*` interfaces, and there should not have any adherence to using jetty or tomcat or whatever else as servlet container. You could have these dependencies into your webapp, and nest your webapp into a tomcat or a jetty without having any compatibility issues.

And if I'm not wrong, we might even consider adding jetty-servlets as a dependency on the gs webapp's pom.xml.

landryb:
> And if I'm not wrong, we might even consider adding jetty-servlets as a dependency on the gs webapp's pom.xml.

is that's actually a hidden requirement (not stated in the geoserver doc) to enable CORS (and adding it in the context for everyone doesnt cause fallback) then yes it's probably the better way to fix that, for all configurations..

fvanderbiest:
> You could have these dependencies into your webapp, and nest your webapp into a tomcat or a jetty without having any compatibility issues.

Let's do it for the next release, we're running out of time here.

PR #3085, opened 2020-06-23
extractor addon - help text

fvanderbiest:
Help text displayed before the extractor fields ...
Capture du 2020-06-23 15-58-21

PR #3100, opened 2020-07-03
mapfishapp - allow addons to override i18n strings

fvanderbiest:
Since #3085 one needs a mechanism to override i18n strings provided by the addon.
This is now possible, see eg https://github.com/georchestra/datadir/pull/188


fvanderbiest:
pushed to master with 30aeeaa11

PR #3103, opened 2020-07-07
docs - adding some doc to customize GN / LDAP synchronization

pmauduit:


docs/geonetwork-synchronization.md
```diff
@@ -0,0 +1,61 @@
...
+map the label of the group onto the `description` attribute from the ldap:
+
+```diff
+-      <entry key="ldapGroupSearchPattern" value="(.*)"/>
+-      <entry key="ldapGroupLabelAttribute" value="o"/>
++     <entry key="ldapGroupSearchPattern" value="EL_(.*)"/>
++     <entry key="ldapGroupLabelAttribute" value="description"/>
+```
+
+Finally, we need to edit the
```
fvanderbiest, line 39: I would put this § first, it helps understanding what we're doing here.

docs/geonetwork-synchronization.md
```diff
@@ -0,0 +1,61 @@
...
+-ldap.groups.search.base=ou=orgs
++ldap.groups.search.base=ou=roles
+```
+
+# Automatic creation of groups
+
+GeoNetwork in geOrchestra is also configured to create automatically
+non-existing groups when the user connects onto the catalogue. With the
+previous configuration, mapping the geonetwork groups onto the `ou=roles` LDAP
+branch will create groups like `ADMINISTRATOR`, `REVIEWER` and so on at first
```
fvanderbiest, line 53: Still dont get it - why would they be synced since `ldapGroupSearchPattern=EL_(.*)` ?

pmauduit: because of this method:
https://github.com/georchestra/geonetwork/blob/ee5644710df42acc2900fb01c8b28f993b33afec/core/src/main/java/org/fao/geonet/kernel/security/ldap/LDAPUtils.java#L129-L175 Which does not take care of the pattern.

Note that the previous method is not called during the synchronization process, but at session creation (e.g. when the connected user first loads a geonetwork page). The `user.getPrivileges()` call line 134 returns a map indexed as  `group name => user profile for this group`, where we find back the elements from the LDAP branch. The main question is why this multimap is populated this way ?

PR #3119, opened 2020-08-12
Support ldap authentication via SASL

### LINKED ISSUE #2705: Console - Handle password regeneration for SASL users
Currently, password regeneration is made directly on the user in the OpenLDAP.

The problem is that when we use SASL for delegating authentification to an external Active Directory (or other solutions), the OpenLDAP user's password store informations for SASL account linking (ex : {SASL}julien.sabatier).

If an user imported from the AD regenerate his password, it will be done only in the LDAP, and not repercuted to the AD.

Is it possible to allow to configure an optional URL for AD users where they can reset their password ?

Maybe we can use an LDAP class (ADUser) to identify users from AD, and then if the user has this class, redirect to this page when he want regenerate his password, or if no URL found, not allow him to regenerate.

RemiDesgrange:
Fix #2705 and #1533 

When a user has a password starting with `{SASL}` georchestra openldap should ask bind to another ldap. This PR also implement check in the console to warn or disable some endpoint:

* `changePassword` :  forbidden to users
* `passwordRecovery`: display a warning message to the administrator. This message say that resetting the password will lose the ability to control password in an external ldap (or ActiveDirectory, but it's the same).
* Implement a checkbox in the console to activate logging via SASL with this user.

What's not in this PR:

* SASL config for your openldap. docs [here](https://github.com/georchestra/georchestra/blob/master/docs/tutorials/sasl.md)


console/src/main/java/org/georchestra/console/ds/AccountDaoImpl.java
```diff
@@ -601,6 +608,13 @@ public Object mapFromContext(Object ctx) {
             }
 
             account.setPending(context.getDn().startsWith(pendingUserSearchBaseDN));
+            // deals with sasl user (external authentication)
+            byte[] rawPassword = (byte[]) context.getObjectAttribute("userPassword");
+            String password = new String(rawPassword);
+            int typeIndexLast = password.lastIndexOf("}");
+            if (PasswordType.valueOf(password.substring(1, typeIndexLast)) == PasswordType.SASL) {
```
cmangeat, line 619: PasswordType.valueOf("mo") throwing IllegalArgumentException

pmauduit:
encountered while testing, it seems that the security-mappings.xml from the datadir is not loaded anymore when initiating the bean (@PostConstruct annotation does not work):
https://github.com/georchestra/georchestra/commit/bcf9ed22133b6452fb4ddc470193f701592ed705

We had to manually re-set a init-method in the xml file so that we could have the /console/manager/public/*.js files as anonymous, or else the default rules were set.

pmauduit:
the passwordRecovery endpoint does not work anymore, returning a 404 status (this might not be related to the current PR)

pmauduit:
still TODO:

* [ ] passwordRecovery ends up with 404
* [ ] PostConstruct does not work
* [ ] -1 on scanInterval, maven jetty plugin ?
* [ ] live test on a docker compo (for now the runtime tests have been made on a ansible-provisioned vagrant)
* [ ] Jacoco or any other tools to measure coverage, and see if some modified parts could be tested
* [ ] adding utests ? see previous point
* [ ] live-test on a SASL-enabled env

pmauduit (approved):

cmangeat (approved):
(we should maybe wait first feedback from integration before to merge ?)

console/src/main/webapp/manager/app/assets/lang/en.json
```diff
@@ -284,5 +284,7 @@
   "date.3month"           : "Last 3 months",
   "date.year"             : "Last year",
   "date.custom"           : "Custom range",
-  "date.to"               : "to"
+  "date.to"               : "to",
+  "sasl.externallogin"    : "External login",
```
fvanderbiest, line 288: Where is `sasl.externallogin` being used ?

console/src/main/webapp/manager/app/templates/userForm.tpl.html
```diff
@@ -126,18 +126,28 @@
...
           <i class="glyphicon glyphicon-new-window"></i>
         </a>
       </div>
-
+      <div ng-if="platformInfos.saslEnabled">
+        <div class="form-group form-group-sm" ng-if="model.$promise">
+          <label class="col-sm-4" for="saslUser">
+            <span translate>sasl.remote.user</span>
+            {{ platformInfos.saslServer }}
+          </label>
```
fvanderbiest, line 144: CC @tonio I'd like to have a tooltip on this field, to explain a bit more. What's your opinion on https://www.w3schools.com/howto/howto_css_tooltip.asp ?

tonio: There's already bootstrap, so I'd advise to use their tooltip system: https://getbootstrap.com/docs/3.4/javascript/#tooltips

Quickly grepin' the code, looks like it may already be used [here](https://github.com/georchestra/georchestra/blob/eab11000ea82717a5ce0ef8a5ea8221726e51fcf/console/src/main/webapp/manager/app/components/logger/logger.tpl.html#L92)

fvanderbiest: Cool. Looks easy to implement !
@pmauduit would you like to give it a try ?

console/src/main/webapp/manager/app/assets/lang/en.json
pmauduit: anywhere

console/src/main/webapp/manager/app/templates/userForm.tpl.html
pmauduit: > @pmauduit would you like to give it a try ?

Screenshot from 2020-09-17 13-02-33

fvanderbiest:
LGTM, thanks !

PR #3128, opened 2020-08-24
LDAP - Add mapstore admin role 3112

pmauduit:
Will probably require a backport onto `20.0.x`

PR #3131, opened 2020-08-25
MAPSTORE_ADMIN role - Adding a migration note

pmauduit:
Backport of #3128 on 20.0.x


pmauduit:
The former PR has already been reviewed, I'm considering merging it as it is, actually.

PR #3144, opened 2020-09-14
GeoServer 2.17.2 update

pmauduit:
This is the following of the previous one which targeted GS 2.17.0, see: https://github.com/georchestra/georchestra/pull/3070

* utests OK
*  runtime tested on a Jetty & a tomcat

Note for tomcat users:

Using the `useRelativeRedirects=false` parameter on the context is still mandatory (or else, you will encounter redirection loops). See:
https://github.com/georchestra/georchestra/blob/master/docs/setup/tomcat.md#create-the-instance-2

About the  CSRF filter triggered and returning "400 bad request" errors:
https://github.com/georchestra/georchestra/blob/master/docs/setup/tomcat.md#in-case-of-troubles-with-the-geoserver-ui

It is possible to avoid this parameter by configuring the Connector as follows (proxyName, proxyPort and scheme attribute):

```xml
    <Connector port="8080" protocol="HTTP/1.1"
               connectionTimeout="20000"
               proxyName="georchestra.mydomain.org"
               proxyPort="443"
               scheme="https"
               redirectPort="8443" />
```

### DIFF
--- a/.github/workflows/geoserver.yml
+++ b/.github/workflows/geoserver.yml
@@ -55,13 +55,13 @@ jobs:
     - name: "Building docker image with native security"
       if: github.repository == 'georchestra/georchestra'
       working-directory: geoserver/webapp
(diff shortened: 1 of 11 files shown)


pmauduit:
* runtime tested in a docker environment (jetty + tomcat)
* runtime tested in a vagrant provisioned via Ansible

Screenshot from 2020-09-14 19-39-38

It still need to update the documentation, as the proxyName / proxyPort / scheme attributes are required once again on the Connector for tomcat users.

pmauduit:
proxyName / proxyPort / scheme not needed in tomcat if web.xml is modified for the CSRF whitelist parameter, the current state of the doc is correct.

fvanderbiest:
IIUC we're good to go ? :+1:

PR #3157, opened 2020-10-01
Fix Access to area.geojson.

RemiDesgrange:
Access to area.geojson in  console/account/new and userdetails where
broken (in some cases). Fix the OrgsController

### DIFF
--- a/console/src/main/java/org/georchestra/console/ws/backoffice/orgs/OrgsController.java
+++ b/console/src/main/java/org/georchestra/console/ws/backoffice/orgs/OrgsController.java
@@ -109,17 +109,6 @@ public class OrgsController {
     @Value("${AreaMapZoom:6}")
     private String areaMapZoom;
 
-    /*
-     * URL of a static file or a service with a GeoJSON FeatureCollection object
-     * string in EPSG:4326.
-     *
-     * Example of a "dynamic" areasUrl:
-     * https://my.server.org/geoserver/ows?SERVICE=WFS&REQUEST=GetFeature&typeName=
-     * gadm:gadm_for_countries&outputFormat=json&cql_filter=ISO='FRA' or ISO='BEL'
-     */
-    @Value("${AreasUrl:https://www.geo2france.fr/public/communes_zones_competence.geojson}")
-    private String areasUrl;
-
     /* The following properties are used to configure the map widget behavior */
 
     /* Key stored in the org LDAP record to uniquely identify a feature. */
@@ -386,7 +375,7 @@ public class OrgsController {
             LOG.info("Could not parse value", e);
         }
         JSONObject areas = new JSONObject();
-        areas.put("url", areasUrl);
+        areas.put("url", "/console/public/area.geojson");
         areas.put("key", areasKey);
         areas.put("value", areasValue);
         areas.put("group", areasGroup);
--- a/console/src/main/java/org/georchestra/console/ws/backoffice/orgs/OrgsController.java
+++ b/console/src/main/java/org/georchestra/console/ws/backoffice/orgs/OrgsController.java
@@ -358,7 +358,7 @@ public class OrgsController {
      * "value": "commune_name", "group": "department_name"} }
      */
 
-        @RequestMapping(value = PUBLIC_REQUEST_MAPPING + "/areaConfig.json", method = RequestMethod.GET)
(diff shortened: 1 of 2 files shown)


console/src/main/java/org/georchestra/console/ws/backoffice/orgs/OrgsController.java
```diff
@@ -386,7 +375,7 @@ public void getAreaConfig(HttpServletResponse response) throws IOException, JSON
             LOG.info("Could not parse value", e);
         }
         JSONObject areas = new JSONObject();
-        areas.put("url", areasUrl);
+        areas.put("url", "/console/public/area.geojson");
```
fvanderbiest, line 378: Doesn't this look like technical debt ?

RemiDesgrange: We have replaced the discovery of "area" by an HTTP endpoint which will send you a file or a redirect to another HTTP endpoint. The configuration is now part of this "Area HTTP endpoint". Before doing this I looked at the first part of the path  and if it can be customized. It seems not, since `/console` is hardcoded everywhere in the manager js code (and in some Java class file).

fvanderbiest: Sorry I should have been more explicit.
"url: /console/public/area.geojson" does not need to be part of the areaConfig.json response, since it is now static.
You could have modified the JS code to directly hit /console/public/area.geojson instead...

console/src/main/webapp/manager/app/components/area/area.es6
```diff
@@ -121,7 +121,7 @@ class AreaController {
 
       this.updateSelection(selected)
       this.loading = false
-    })
+    }).catch(ex => console.log('Cannot display area. GeoJSON file is malformed or Area config is wrongly configured in the data', ex))
```
RemiDesgrange, line 124: Asking thought of @tonio on this.

tonio: I think a broken GeoJSON file will break at line 97.
What about testing this above, & add a specific test for config with an explicit error message ?

console/src/main/webapp/manager/app/components/area/area.es6
```diff
@@ -97,10 +97,17 @@ class AreaController {
       vector.getSource().addFeatures(format.readFeatures(json, conf))
       vector.getSource().forEachFeature(f => {
         const group = f.get(config.areas.group)
+        if (group === undefined) {
+          throw new Error(`Cannot get AreaGroup ${config.areas.group} in provided geojson. Check datadir config.`)
```
tonio, line 101:
```suggestion
          throw new Error(`Cannot get AreaGroup «${config.areas.group}» in provided geojson. Check datadir config.`)
```

console/src/main/webapp/manager/app/components/area/area.es6
```diff
@@ -97,10 +97,17 @@ class AreaController {
...
+        }
         if (this.groups.indexOf(group) < 0) {
           this.groups.push(group)
         }
-        f.set('_label', f.get(config.areas.value).toString() || '')
+        const displayName = f.get(config.areas.value)
+        if (displayName === undefined) {
+          throw new Error(`Cannot get AreaValue ${config.areas.value} in provided geojson. Check datadir config.`)
```
tonio, line 108:
```suggestion
          throw new Error(`Cannot get AreaValue «${config.areas.value}» in provided geojson. Check datadir config.`)
```

console/src/main/webapp/manager/app/components/area/area.es6
```diff
@@ -83,7 +83,7 @@ class AreaController {
     const format = new ol.format.GeoJSON()
 
     this.loading = true
-    this.$injector.get('$http').get(config.areas.url).then(
+    this.$injector.get('$http').get('/console/public/area.geojson').then(
```
tonio, line 86: you may create a service so that is easier to track

tonio (approved):
Small comments, otherwise (client side) lgtm. :sunglasses:

console/src/main/webapp/manager/app/components/area/area.es6
RemiDesgrange: Sorry but « and » are for French. If English I'll use double quote :-)

RemiDesgrange:
Ok tests are failing no idea why, will investogate this.

Issue #3173, opened 2020-10-09
console - jetty - unable to resolve mail templates anymore 


pmauduit:
Comparing 2 different setups in a version close to master, we got an NPE when trying to resolve templates, indeed:

https://github.com/georchestra/georchestra/blob/master/console/src/main/java/org/georchestra/console/mailservice/Email.java#L81

`servletContext.getRealPath(fileTemplate)` returns null in jetty v9.4.32, but a non-null string in v9.4.31, it looks like the method has been improved to return null in case the file does not exist in the webapp"s classpath.

The templates in the webapp have been removed since 21 nov 2018 with d4d39712756140060909beb440c84d13eabd6fe0.

pmauduit:
Jetty returns null here for the previous call around here:

https://github.com/eclipse/jetty.project/blob/jetty-9.4.32.v20200930/jetty-server/src/main/java/org/eclipse/jetty/server/handler/ContextHandler.java#L2292

In 9.4.32, the resource object is of type `JarResource` which returns null for getFile calls:
https://github.com/eclipse/jetty.project/blob/jetty-9.4.32.v20200930/jetty-util/src/main/java/org/eclipse/jetty/util/resource/JarResource.java#L100-L105

in 9.4.31, the resource object is of type `PathResource`, which is ok for resolving the path even if it does not exist.

PR #3174, opened 2020-10-09
console - fixing mail templates logic (#3173)

pmauduit:
Note: I reintroduced the templates inside the webapp. I know that the `georchestra.datadir` is now mandatory, but I dislike the fact that the webapps cannot run by themselves, and require external resources that they could provide.

* tests: added some basic integration tests for the EmailFactory class,
  that still could be improved.
* UT ok
* IT ok


pmauduit:
I'm willing to push a custom tag on docker-hub to be able to runtime test it.

pmauduit:
runtime tested, works again on RM-test infrastructure.

RemiDesgrange (approved):
Only review the code. Clean :ok_hand:

Issue #3182, opened 2020-10-28
[JAVA11] Dependency problem in Georchestra commons


jusabatier:
Compiling Georchestra with Java11 generate a dependency error because of the @PostConstruct annotation.

It has already been mentioned by @pmauduit, but has not been fixed yet : https://github.com/georchestra/georchestra/issues/3130#issuecomment-679909644

RemiDesgrange:
Now fixed

PR #3183, opened 2020-10-28
Solve annotation dependency problem with Java11 for geOrchestra's commons

jusabatier:
Correct https://github.com/georchestra/georchestra/issues/3182


fvanderbiest:
Thanks Julien !
Pierre will have a look when he's back.

pmauduit:
Not related to the PR, but the GH actions are failing, we shall make sure that the "login onto docker.io" step is skipped once not on the georchestra organization. Actually this is weird, as it should already be the case:
https://github.com/georchestra/georchestra/blob/master/.github/workflows/analytics.yml#L58

Regarding the PR, adding the dependency is ok for me, but I'd have done it in a profile, as adding the dependency is not necessary in former Java versions. Also, the `@PostConstruct` annotation has also been introduced in some other modules (SP, console ...). As they depend on georchestra-commons, it is probably ok to keep it this way, but I'd prefer have the dependency explicitely required each time we need it.

fvanderbiest:
@jusabatier would you be able to rework the PR as advised by Pierre ? Thanks !

jusabatier:
> dependency is not necessary in former Java versions

Isn't version 20.1 of georchestra supposed to be deployed in Debian Buster (10) ?
In this case, JAVA 11 is the default version, so I don't understand why create a profile for support the default version...
It only can make the build of georchestra more complex to understand by adding an unnecessary profile to set.

jusabatier:
This issue still block the compilation with JAVA11.

So for now it's difficult to compile war of latest release (21.1.1)

fvanderbiest:
OK, what is the status of this PR ? 
Seems it cannot be merged as is, since there are conflicts.

jusabatier:
Conflits corrected

fvanderbiest:
@pmauduit can you review a second time or take over ? Thanks !

jusabatier:
Wait a little, I get some errors linked to this compiling GN.
I'm checking out what is blicking and will add a commit for this.

jusabatier:
Sorry, as GN is a separate repo, you can check and merge this PR

pmauduit:
superseded by #3581

Issue #3184, opened 2020-10-30
[Console] NullPointerException if no org


jusabatier:
I tested console from master branch (20.1), and when I log in with a user who isn't assigned to an organisation, I get a NullPointerException on the account detail page.

fvanderbiest:
Thanks for the report Julien !
Can you paste the stack trace ?

jusabatier:
> 2020-10-28 14:25:34 GlobalExceptionHandler [ERROR] java.lang.NullPointerException
> java.io.IOException: java.lang.NullPointerException
> 	at org.georchestra.console.ws.edituserdetails.EditUserDetailsFormController.setupForm(EditUserDetailsFormController.java:130)
> 	at org.georchestra.console.ws.edituserdetails.EditUserDetailsFormController$$FastClassBySpringCGLIB$$26f36d5a.invoke(<generated>)
> 	at org.springframework.cglib.proxy.MethodProxy.invoke(MethodProxy.java:204)
> 	at org.springframework.aop.framework.CglibAopProxy$CglibMethodInvocation.invokeJoinpoint(CglibAopProxy.java:736)
> 	at org.springframework.aop.framework.ReflectiveMethodInvocation.proceed(ReflectiveMethodInvocation.java:157)
> 	at org.springframework.security.access.intercept.aopalliance.MethodSecurityInterceptor.invoke(MethodSecurityInterceptor.java:69)
> 	at org.springframework.aop.framework.ReflectiveMethodInvocation.proceed(ReflectiveMethodInvocation.java:179)
> 	at org.springframework.aop.framework.CglibAopProxy$DynamicAdvisedInterceptor.intercept(CglibAopProxy.java:671)
> 	at org.georchestra.console.ws.edituserdetails.EditUserDetailsFormController$$EnhancerBySpringCGLIB$$751487fe.setupForm(<generated>)
> 	at sun.reflect.NativeMethodAccessorImpl.invoke0(Native Method)
> 	at sun.reflect.NativeMethodAccessorImpl.invoke(NativeMethodAccessorImpl.java:62)
> 	at sun.reflect.DelegatingMethodAccessorImpl.invoke(DelegatingMethodAccessorImpl.java:43)
> 	at java.lang.reflect.Method.invoke(Method.java:498)
> 	at org.springframework.web.method.support.InvocableHandlerMethod.doInvoke(InvocableHandlerMethod.java:205)
[...]

PR #3200, opened 2020-11-04
[Console] Check if user have an org. Orgs are not mandatory in georchestra

RemiDesgrange:
This is a quick fix in order to get rid of NPE, but a nice refactor could be to add _explicitly_ in the type the optional nature of the Org. like `Optional<String>` in front of all `getOrg()` method.

linked to #3184 cc @jusabatier


fvanderbiest:
@RemiDesgrange can you backport too ? Thanks

fvanderbiest:
backported to 20.0.x and 20.1.x

Issue #3202, opened 2020-11-09
[Mapfishapp] Streetview mixed content


jusabatier:
Some users have problems with streetview addon : 

Their navigator (chrome) block mixed content, so the image from streetview service isn't displayed.

In the addon code, the URL are hardcoded with http :  

https://github.com/georchestra/georchestra/blob/master/mapfishapp/src/main/webapp/app/addons/streetview/js/main.js#L292

For solve this problem, it should be https if the website is over https.

PR #3203, opened 2020-11-09
Handle https to avoid mixed content in streetview addon

jusabatier:
This is a patch to avoid mixed content errors in the streetview addon : 

https://github.com/georchestra/georchestra/issues/3202


fvanderbiest:
Thanks Julien.

Patch was backported to 20.0.x and 20.1.x, targeting the 20.0.7 and 20.1.1 releases.

BTW: github's milestone indicator only allows a single target, this will complicate things if we want to support two branches at the same time. Labels ?

PR #3204, opened 2020-11-09
Spring authentication based on pre-authenticated georchestra headers

groldan:
Leverage pre-authenticated georchestra's "security proxy" request headers to implement a Spring authentication configuration.

Instead of checking for the request's `sec-user` and `sec-roles` manually on each api entry point, implement an authentication manager, filter, and configure them to set a regular spring authentication token.
Api methods can be annotated like `@RolesAllowed("ROLE_ADMIN")`, etc.

### DIFF
--- a/datafeeder/src/main/java/org/georchestra/config/security/GeorchestraSecurityProxyAuthenticationFilter.java
+++ b/datafeeder/src/main/java/org/georchestra/config/security/GeorchestraSecurityProxyAuthenticationFilter.java
@@ -16,21 +16,46 @@ import lombok.extern.slf4j.Slf4j;
 public class GeorchestraSecurityProxyAuthenticationFilter extends AbstractPreAuthenticatedProcessingFilter {
 
     @Override
-    protected Object getPreAuthenticatedPrincipal(HttpServletRequest request) {
-        String principal = request.getHeader("sec-username");
-        return principal;
+    protected GeorchestraUserDetails getPreAuthenticatedPrincipal(HttpServletRequest request) {
+        for (String h : Collections.list(request.getHeaderNames())) {
+            System.err.printf("%s=%s%n", h, request.getHeader(h));
+        }
+        final boolean preAuthenticated = Boolean.parseBoolean(request.getHeader("sec-proxy"));
+        if (preAuthenticated) {
+            String username = request.getHeader("sec-username");
+            final boolean anonymous = username == null;
+            if (anonymous) {
+                username = "anonymousUser";
+            }
+            List<String> roles = extractRoles(request);
+            String email = request.getHeader("sec-email");
+            String firstName = request.getHeader("sec-firstname");
+            String lastName = request.getHeader("sec-lastname");
+            String organization = request.getHeader("sec-org");
+            String organizationName = request.getHeader("sec-orgname");
+            return new GeorchestraUserDetails(username, roles, email, firstName, lastName, organization,
+                    organizationName, anonymous);
+        }
+        return null;
     }
 
+    /**
+     * @return {@code true} if the request comes from georchestra's security proxy
+     */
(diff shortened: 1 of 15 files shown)


datafeeder/src/main/java/org/georchestra/datafeeder/api/DataFeederApiConfiguration.java
```diff
@@ -20,10 +20,15 @@
...
 import org.springframework.context.annotation.Configuration;
+import org.springframework.security.config.annotation.method.configuration.EnableGlobalMethodSecurity;
+import org.springframework.security.config.annotation.method.configuration.GlobalMethodSecurityConfiguration;
 import org.springframework.web.servlet.config.annotation.EnableWebMvc;
 
 @EnableWebMvc
 @ComponentScan
-public @Configuration class DataFeederApiConfiguration {
+@EnableGlobalMethodSecurity(//
+        jsr250Enabled = true // enable @RolesAlloed annotation in API methods
```
RemiDesgrange, line 30: Typo ? `RolesAllowed` ?

datafeeder/src/main/java/org/georchestra/datafeeder/autoconf/GeorchestraIntegrationAutoConfiguration.java
```diff
@@ -15,8 +16,8 @@
  */
 @Configuration
 @Profile("georchestra")
-@Import(GeorchestraDatadirConfiguration.class)
-public class GeorchestraIntegrationConfiguration {
+@Import({ GeorchestraDatadirConfiguration.class, GeorchestraSecurityProxyAuthenticationConfiguration.class })
```
pmauduit, line 19: for my own knowledge: what is the purpose of the `@Import` spring annotation ? I can read on the net that it helps grouping configuration classes, but I'd expect spring to scan the classpath for @configuration annotated classes and do the job of resolving the dependency graph by itself, without the need of using these imports.

datafeeder/src/main/java/org/georchestra/config/security/GeorchestraSecurityProxyAuthenticationFilter.java
```diff
@@ -0,0 +1,30 @@
...
+        return principal;
+    }
+
+    @Override
+    protected Object getPreAuthenticatedCredentials(HttpServletRequest request) {
+        String rolesHeader = request.getHeader("sec-roles");
+        if (StringUtils.isEmpty(rolesHeader)) {
+            return Collections.emptyList();
+        }
+        String[] roles = rolesHeader.split(",");
```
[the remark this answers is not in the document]
fvanderbiest: Good catch
https://github.com/georchestra/georchestra/blob/740a2ecbd428ac487065ce3bad7d6ff5a51bfa4f/security-proxy/src/main/java/org/georchestra/security/Proxy.java#L616

datafeeder/src/main/java/org/georchestra/datafeeder/autoconf/GeorchestraIntegrationAutoConfiguration.java
groldan: your expectation is based on "old fashion" spring "scan all classpath" approach though, which is actually an anti-pattern, besides time-consuming. I'd rather know what exactly is gonna be configured by grouping configuration classes hierarchically, and enabling them using profiles where appropriate. In this app we even use our own auto-configuration classes to automatically set up the required environment to work as a georchestra app or not, based on the active profile.

pmauduit: ok makes sense

PR #3206, opened 2020-11-12
[20.1.x] fix gdpr account deletion

### LINKED ISSUE #3181: console - the "delete my account" button does not work
Nothing happens when I click on the "delete my account" button in my user details page from the console:

Screenshot from 2020-10-15 16-49-26

No request is made from the browser, and I cannot see any error in the JS console either.

jeanpommier:
Fixes #3181


jeanpommier:
To port to master too
Commit bc981d4a might also interest 20.0 branch

jeanpommier:
commit ede92a5  was not fixing the issue, it was, actually, invalidating the condition.

What happened was that because of the ng-if instruction on the div, the div was created by angular, posterior to document load. Meaning that the jquery selector was trying to select a div that didn't exist yet. Hence no reaction at all. The logic behind clicking the delete button needed to be placed in the Angular Controller (or use ng-show instead of ng-if => the div would be part of the original DOM)

console/src/main/webapp/manager/app/app.es6
```diff
@@ -73,18 +73,43 @@ class AppController {
...
+    if (!$window.org) {
       $scope.org = new Org()
       return
     }
 
-    $scope.org = window.org
-    $scope.users = window.org.members
-    $scope.isReferentOrSuperUser = window.isReferentOrSuperUser
-    $scope.gdprAllowAccountDeletion = window.gdprAllowAccountDeletion
+    this.$injector = $injector
```
tonio, line 84: doesn't looks necessary to store it in object.

console/src/main/webapp/manager/app/app.es6
```diff
@@ -73,18 +73,43 @@ class AppController {
...
+          if (response.ok) {
+            $window.location.href = '/logout'
+          } else {
+            $window.alert(i18n.deleteFail)
+          }
+        })
+        .catch(function () {
+          $window.alert(i18n.deleteFail)
+        })
+    }
```
tonio, line 112: Now that you move this code into a place where it's transpiled, you could save some few chars using arrow functions.

tonio (approved):
Minor comments, the fix looks a lot more solid now. :+1: Thanks!

console/src/main/webapp/manager/app/app.es6
jeanpommier: right

console/src/main/webapp/manager/app/app.es6
jeanpommier: Sure. I considered it was time to stop fussing around, but there are still lots of improvements to do: probably use $http.post instead of fetch for instance. And indeed, arrow functions

jeanpommier:
I was also considering removing the corresponding i18n strings from WEB-INF/i18n/application_??.properties since they are not used any more in Java/jsp

console/src/main/webapp/manager/app/app.es6
jeanpommier: Anyway, it looks that there would be some optimization to do. The editUserDetailsForm.jsp is some strange mix between jsp form, jquery and angular... I'd rather use one system for all of the form...

jeanpommier: @tonio  Following this discussion, I've optimized a bit the code. OK for you ?

console/src/main/webapp/manager/app/app.es6
```diff
@@ -73,18 +73,32 @@ class AppController {
...
+    $injector.get('translate')('editUserDetailsForm.deleteConfirm', i18n)
+    $injector.get('translate')('editUserDetailsForm.deleteFail', i18n)
+
+    $scope.deleteUser = function () {
+      if (!$window.confirm(i18n.deleteConfirm)) return false
+      $http.post($scope.deleteURI)
+        .then(
+          function success (response) { $window.location.href = '/logout' },
+          function error () { $window.alert(i18n.deleteFail) }
+        )
```
tonio, line 100:
```suggestion
        .then(
          () => $window.location.href = '/logout',
          () => $window.alert(i18n.deleteFail)
        )
```

console/src/main/webapp/manager/app/app.es6
```diff
@@ -73,18 +73,32 @@ class AppController {
...
-    $scope.gdprAllowAccountDeletion = window.gdprAllowAccountDeletion
+    $scope.org = $window.org
+    $scope.users = $window.org.members
+    $scope.isReferentOrSuperUser = $window.isReferentOrSuperUser
+    $scope.gdprAllowAccountDeletion = $window.gdprAllowAccountDeletion
+    $scope.deleteURI = $injector.get('CONSOLE_BASE_PATH') + 'account/gdpr/delete'
+
+    const i18n = {}
+    $injector.get('translate')('editUserDetailsForm.deleteConfirm', i18n)
+    $injector.get('translate')('editUserDetailsForm.deleteFail', i18n)
```
tonio, line 92: Since you only use $injector for translate, you can inject `translate` instead of `$injector` and use it here.

tonio (approved):
Yep, nice, thanks! :+1:

console/src/main/webapp/manager/app/app.es6
jeanpommier: thanks for the info.  Although I'm also using it to get the CONSOLE_BASE_PATH variable. So let's keep it that way ?

tonio: yep

fvanderbiest:
> To port to master too
> Commit [bc981d4](https://github.com/georchestra/georchestra/commit/bc981d4ad3de667cb8060ad673758ddec8bc5a26) might also interest 20.0 branch

Quoting Jean ... cc @RemiDesgrange

fvanderbiest:
OK, I'm taking care of it.

PR #3224, opened 2020-12-11
[console][fix][master] edit organization link, from user form

jeanpommier:
(console/account/userdetails)

Link was followed in the browser's URL bar, but not actually loaded


RemiDesgrange (approved):
:shipit:

PR #3225, opened 2020-12-11
[console][fix][20.1.x] edit organization link, from user form

jeanpommier:
(console/account/userdetails)
Link was followed in the browser's URL bar, but not actually loaded


RemiDesgrange (approved):
:shipit:

PR #3236, opened 2020-12-21
Bump to geoserver v2.16.2

pmauduit:
This update is needed to have the web-resources plugin working (see https://osgeo-org.atlassian.net/browse/GEOS-9348).

Tests:
* compilation with tests OK
* runtime test OK (webapp is able to bootstrap, GS UI can load, I did not test further to be honest)

### DIFF
--- a/geoserver/pom.xml
+++ b/geoserver/pom.xml
@@ -11,12 +11,12 @@
   <name>GeoServer 2.x root module</name>
   <properties>
     <skipTests>true</skipTests>
-    <gs.version>2.16.1</gs.version>
-    <gt.version>22.1</gt.version>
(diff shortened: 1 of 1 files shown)


fvanderbiest (approved):
Thanks !
Makes me think we should trigger a 20.0.7 release ...

fvanderbiest:
Oh, why not going to 2.16.5 ? 
http://geoserver.org/2020/08/31/geoserver-2-16-5-released.html

pmauduit:
> Oh, why not going to 2.16.5 ?

it should not be a big deal

pmauduit:
moving to 2.16.5 instead, closing

PR #3246, opened 2021-01-13
Overriding default and setting up GWC with jdbc for docker.

RemiDesgrange:
For docker setup. JDBC is now configured for geoserver. 

The default value is still H2 when geoserver first boot, but you can change it in the interface.
image


geoserver/webapp/src/docker/jetty-env.xml
```diff
@@ -0,0 +1,19 @@
...
+<Configure class="org.eclipse.jetty.webapp.WebAppContext">
+   <New id="ds_gwc_geoserver" class="org.eclipse.jetty.plus.jndi.Resource">
+     <Arg>jdbc/gwc</Arg>
+     <Arg>
+       <New class="org.apache.commons.dbcp.BasicDataSource">
+         <Set name="driverClassName">org.postgresql.Driver</Set>
+         <Set name="username">georchestra</Set>
+         <Set name="password">georchestra</Set>
+         <Set name="url">jdbc:postgresql://database/georchestra</Set>
+         <Set name="validationQuery">SELECT 1</Set>
```
pmauduit, line 12: I don't know if this is relevant here, but we could also use env variables (too bad that we should also have the info in the default.properties from the georchestra datadir)

docs/setup/postgresql.md
```diff
@@ -54,6 +54,11 @@ If the **Atlas** is deployed:
 psql -d georchestra -f postgresql/060-atlas.sql
 ```
 
+## GeoWebCache schema
+
+If geowebcache is used it creates a geowebcache schema to store quota and relevant infos to
```
pmauduit, line 59: "is used, ..." (adding a coma)

docs/setup/postgresql.md
```diff
@@ -54,6 +54,11 @@ If the **Atlas** is deployed:
...
+geowebcache. This setup is highly encouraged in a production environment to replace local H2 database.
```
pmauduit, line 60: "... to replace the local H2 database"

geoserver/webapp/src/docker/Dockerfile
```diff
@@ -28,11 +28,17 @@ RUN mkdir /mnt/geoserver_datadir /mnt/geoserver_geodata /mnt/geoserver_tiles /mn
 
 # add a tweaked configuration. First use case is support of OPTIONS.
 ADD web.xml /var/lib/jetty/webapps/geoserver/WEB-INF/web.xml
+ADD jetty-env.xml /var/lib/jetty/webapps/geoserver/WEB-INF/jetty-env.xml
+# tweaked geowebcache config that use postgresql instead of H2. Copying it to /tmp since copying it
```
pmauduit, line 32: " ... which uses postgresql"

geoserver/webapp/src/docker/Dockerfile
```diff
@@ -28,11 +28,17 @@ RUN mkdir /mnt/geoserver_datadir /mnt/geoserver_geodata /mnt/geoserver_tiles /mn
 
 # add a tweaked configuration. First use case is support of OPTIONS.
 ADD web.xml /var/lib/jetty/webapps/geoserver/WEB-INF/web.xml
+ADD jetty-env.xml /var/lib/jetty/webapps/geoserver/WEB-INF/jetty-env.xml
+# tweaked geowebcache config that use postgresql instead of H2. Copying it to /tmp since copying it 
+# to /mnt/geoserver_tiles won't work (it's a volume, data will be override). File copied to
```
pmauduit, line 33: "data will be overridden). Instead, the file will be copied to the actual destination by the entrypoint script.

geoserver/webapp/src/docker/docker-entrypoint.d/00-geoserver-bootstrap
```diff
@@ -16,4 +16,9 @@ else
 
     echo 'Change log config according to docker setup'
     sed -i 's:stdOutLogging>false:stdOutLogging>true:g' /mnt/geoserver_datadir/logging.xml
+
+    # geowebcache config need to be placed in geoserver_tiles volume.
+    # copying it directly in the Dockerfile doesn't work. Copying it at runtime.
```
pmauduit, line 21: geowebcache config needs to be placed into the geoserver_tiles volume.
We cannot copy the file via the Dockerfile at build time, so we are doing it at runtime.

geoserver/webapp/src/docker/jetty-env.xml
RemiDesgrange: Your right. I'll update to env var + default values.

fvanderbiest:
@RemiDesgrange up for the release ? Thanks :-)

fvanderbiest:
> @RemiDesgrange up for the release ? Thanks :-)

Oh, forget it, I'm not releasing master ... yet ;-)

geoserver/webapp/src/docker/Dockerfile
```diff
@@ -28,11 +28,17 @@ RUN mkdir /mnt/geoserver_datadir /mnt/geoserver_geodata /mnt/geoserver_tiles /mn
...
+# /mnt/geoserver_tiles by a entrypoint script.
```
fvanderbiest, line 34: If the image provides `/mnt/geoserver_tiles/geowebcache-diskquota.xml` and `/mnt/geoserver_tiles/geowebcache-diskquota-jdbc.xml`, the first time an empty volume is mounted on it, it shall be initialized with the content provided by the image, which is exactly what we want to achieve here IIUC ?

fvanderbiest: No need to use /tmp then

docs/setup/postgresql.md
fvanderbiest: Missing psql statement with 110-geowebcache.sql  (for those who do the install manually)

geoserver/webapp/src/docker/geowebcache-diskquota-jdbc.xml
```diff
@@ -0,0 +1,14 @@
+<gwcJdbcConfiguration>
+  <dialect>PostgreSQL</dialect>
+  <JNDISource>jdbc/gwc</JNDISource>
+  <connectionPool>
+    <driver>org.postgresql.Driver</driver>
+    <url>jdbc:postgresql://database/georchestra</url>
+    <username>georchestra</username>
+    <password>georchestra</password>
+    <minConnections>1</minConnections>
```
fvanderbiest, line 9: How does GWC know it shall use the `geowebcache` schema to store its tables ?

fvanderbiest (changes requested):
Some gotchas ...

arnaud-morvan:
I moved the config in geoserver_minimal_datadir: https://github.com/georchestra/geoserver_minimal_datadir/pull/14

PR #3255, opened 2021-01-19
jetty - disable server tokens

fvanderbiest:
Do not disclose server version - a good practice


fvanderbiest:
Ported to 20.1.x & master

PR #3259, opened 2021-01-20
Datafeeder file upload and dataset analysis API and implementation

groldan:
Calls to `POST /import/upload` with content-type `multipart/form-data` receive a list of uploaded files with the `filename` parameter.
    
The files are stored to a persistent location under a job identifier (UUID).
    
The persistence root directory is configured though the `datafeeder.file-upload.persistent-location` configuration property,
 which differs from `datafeeder.file-upload.temporary-location` in that the former is guaranteed to be persistent until explicitly deleted, and the later is where the servlet container will save files _if_ their size exceed the configured size threshold, otherwise it keeps them in memory.
    
Uploading archived in .zip, .tar, .tar.gz, .tar.bz2 formats is supported through commons-vfs2 and automatically extracted when uploaded.
    
Uploaded datasets are recognized using simple filename heuristics by file name extension: .shp, .geojson, .gpkg.


fgravin (approved):
Sick work
@tonio has tested it and it seems to work well !

Thanks

PR #3262, opened 2021-01-25
Datafeeder: allow requesting the sample feature and bounds dynamically

groldan:
Remove `DatasetUploadStatus.sampleProperties` and
`DatasetUploadStatus.sampleGeometryWKT`, in favour end point
`/upload/{jobId}/{typeName}/sampleFeature`, which returns GeoJSON and
allows to set the dataset charset and CRS to override or reproject to.

Detect the shapefile's dbf charset from the uploaded .cpg side-car file, if any.

Add the following end points:

* `GET /upload/{jobId}/{typeName}/sampleFeature?[featureIndex=<index>&srs=<srs_code>&srs_reproject=<true|false>]`
    end point to return the sample feature in GeoJSON format, optionally
    specifying the feature index, the coord system identifier, and whether
    to reproject from the native CRS to the one provided.
    
    When `srs_reproject` is not `true`, the provided CRS overrides the
    dataset's native CRS without reprojection.

* `GET /upload/{jobId}/{typeName}/bounds?srs=<srs_code>&srs_reproject=<true|false>]`
    end point to return the dataset bounding box in either the native CRS or
    reprojected to another one.

This allows the UI to experiment/request the sample feature and bounds in different CRS's and encodings, before proceeding with the publication process, where the CRS and the encoding will be parametrized.

### DIFF
--- a/datafeeder/api.yaml
+++ b/datafeeder/api.yaml
@@ -345,13 +345,6 @@ components:
           type: integer
           format: int32
           description: Number of features in the dataset
-        sampleProperties:
-          type: array
-          items:
-            $ref: '#/components/schemas/Property'
-        sampleGeometryWKT:
-          type: string
-          description: Well Known Text representation of an exemplar geometry extracted from the dataset
         nativeBounds:
           $ref: '#/components/schemas/BoundingBox'
         encoding:
--- a/datafeeder/api.yaml
+++ b/datafeeder/api.yaml
@@ -106,6 +106,53 @@ paths:
         409:
           description: 'Conflict. The abort=true parameter was not provided and the analysis or publishing job is running'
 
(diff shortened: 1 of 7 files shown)


tonio:
Tested, works pretty good, thanks!
Only thing, what's the status of encoding detection ? I tried with a file with french `iso-8859-1`, I have to force the `encoding` param to make it correctly parse accentuated chars.

groldan:
> Tested, works pretty good, thanks!
> Only thing, what's the status of encoding detection ? I tried with a file with french `iso-8859-1`, I have to force the `encoding` param to make it correctly parse accentuated chars.

Encoding detection should work as long as a `.cpg` file was uploaded together with the `.shp` (see the "Other files" section in https://en.wikipedia.org/wiki/Shapefile)

Did you upload such file?

PR #3269, opened 2021-01-28
[LDAP] docker, allow passing admin password as a secret 

jeanpommier:
We shouldn't be passing, on docker, the passwords as plain environment variables. This PR allows : 

- the use of SLAPD_PASSWORD_FILE instead of SLAPD_PASSWORD
- the use of GS_MASTER_PASSWORD_FILE to be able to midify the password of geoserver_privileged_user on first run (modifying it on the datadir is not sufficient. We should provide a way to do it seamlessly here too)

This uses the same script used by  [postgresql](https://github.com/docker-library/postgres/blob/ba302205a1300a5ad262ee770f7ac8a1038e8fde/13/docker-entrypoint.sh#L5-L25) 

The use of secrets is demonstrated in the docker-compose.yml

### DIFF
--- a/ldap/docker-compose.yml
+++ b/ldap/docker-compose.yml
@@ -10,6 +10,7 @@ services:
     build: ./
     secrets:
       - slapd_password
+      - gs_master_passwd
     environment:
       - SLAPD_DOMAIN=georchestra.org
       - SLAPD_ORGANIZATION=georchestra
@@ -27,3 +28,5 @@ services:
 secrets:
   slapd_password:
     file: ./secrets/slapd_password.txt
+  gs_master_passwd:
+    file: ./secrets/gs_master_passwd.txt
--- a/ldap/docker-compose.yml
+++ b/ldap/docker-compose.yml
@@ -1,4 +1,4 @@
-version: '3'
+version: '3.1'
 
 volumes:
   ldap_data_test:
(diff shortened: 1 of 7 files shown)


fvanderbiest (approved):
LVGTM, thanks !

fvanderbiest:
Should be ported to https://github.com/georchestra/docker once approved by other devs

ldap/docker-compose.yml
```diff
@@ -8,14 +8,25 @@ services:
...
+      - SLAPD_PASSWORD=
+      - GS_MASTER_PASSWORD_FILE=/run/secrets/gs_master_passwd
       - SLAPD_LOG_LEVEL=32768
     volumes:
       - ldap_data_test:/var/lib/ldap
       - ldap_config_test:/etc/ldap
     ports:
       - '10389:389'
+
+secrets:
```
pmauduit, line 28: I was not aware that secrets were implemented in docker-compose. Will it work with classic docker / docker-compose locally ?

```
$ docker secret ls
Error response from daemon: This node is not a swarm manager. Use "docker swarm init" or "docker swarm join" to connect this node to swarm and try again.
```

pmauduit (approved):
Just a concern about backward compatibility, but otherwise looks good.

ldap/docker-compose.yml
jeanpommier: yep, it does. Docker-compose, locally, does not really store secrets, but uses the files stored in the `secrets` folder. This allows to emulate secrets behaviour.

fvanderbiest:
Thanks Jean for this contribution !

> Should be ported to https://github.com/georchestra/docker

Would you be able to take care of it too ?

jeanpommier:
Sure. Will do

PR #3271, opened 2021-01-29
[geoserver] docker, turn off logs for oshi module

jeanpommier:
The oshi.hardware.platform.linux is causing, on the docker instance, non-critical and non-useful errors, that tend to pollute the geoserver logs. By turning them off, we get more readable logs, and less logging traffic

### DIFF
--- a/geoserver/webapp/src/docker/docker-entrypoint.d/00-geoserver-bootstrap
+++ b/geoserver/webapp/src/docker/docker-entrypoint.d/00-geoserver-bootstrap
@@ -16,4 +16,7 @@ else
 
     echo 'Change log config according to docker setup'
     sed -i 's:stdOutLogging>false:stdOutLogging>true:g' /mnt/geoserver_datadir/logging.xml
+
+    echo 'deactivate oshi logs (hardware monitoring, fails on docker, the error logs pollute the geoserver logs)'
+    echo "log4j.category.oshi.hardware.platform.linux=OFF" >> /mnt/geoserver_datadir/logs/PRODUCTION_LOGGING.properties
 fi


fvanderbiest:
Sounds reasonable, but why not https://github.com/georchestra/geoserver_minimal_datadir/blob/master/logs/PRODUCTION_LOGGING.properties instead ?

jeanpommier:
Yep, might be better

jeanpommier:
Mmmh, actually no. The module works fine (AFAIK) when running on bare-metal or VMs. This is docker-specific

fvanderbiest:
100% right ! Good catch

fvanderbiest:
backported for 20.1.1

landryb:
> Mmmh, actually no. The module works fine (AFAIK) when running on bare-metal or VMs. This is docker-specific

iirc i also see those annoying log messages on lxc containers at each tomcat-geoserver startup.
```
2021-01-26 08:21:28,762 ERROR [linux.LinuxGlobalMemory] - Failed to get mem_unit from sysinfo. {}
java.lang.NoClassDefFoundError: com/sun/jna/platform/linux/LibC$Sysinfo
        at oshi.hardware.platform.linux.LinuxGlobalMemory.<init>(LinuxGlobalMemory.java:62)
        at oshi.hardware.platform.linux.LinuxHardwareAbstractionLayer.getMemory(LinuxHardwareAbstractionLayer.java:59)

```

so maybe fix that in the non-docker case too ?

jeanpommier:
@landryb not surprising, since it is very close to docker context.
The problem is to distingusih the cases where it doesn't work (=> remove logs message) and where it should (>keep them)
Or... a switch ? Using environment variable ?

Issue #3143, opened 2020-09-11
GeoServer - specify submodule's version in docker image


jeanpommier:
geOrchestra 20.x docker images  are designed to be overridden, when someone wants to add specific features, instaed of rebuilding it from the source code. 
For instance, when addind extensions to the geoserver instance, the best practice is supposedly to start FROM the georchestra/geoserver image and create an override Dockerfile above it.
For this, it would be nice to know straight away the version of the geoserver instance (needed to retrieve the extensions).
A simple way would probably be to set an environment variable with the geoserver version number, in the georchestra/geoserver image

fvanderbiest:
Agreed !

PR #3273, opened 2021-01-29
[Geoserver] Docker, add Geoserver version as env. var in docker image

jeanpommier:
solves https://github.com/georchestra/georchestra/issues/3143


RemiDesgrange (approved):
:+1:

PR #3292, opened 2021-02-10
[DataFeeder] Dataset upload to postgis

bhoefling-c2c:
This lets you import data into the PostGIS database.

PR #3299, opened 2021-02-18
[DataFeeder] Initial implementation of dataset publishing to GeoServer

groldan:
* Initial implementation of `GeorchestraOwsPublicationService`, delegating to `GeoServerRemoteService`, which in turn uses the `geoserver-rest-openapi` java client to interact with GeoServer's REST api.

* Adds geoserver-rest-openapi submodule for datafeeder, until it finds its place in a public maven repo.

* Once #3292 is merged, we could add integration tests that target the postgis database in the docker composition used for testing.

* Integration test for GeoServer publishing workflow:
    
Single integration test to publish a dataset through `GeorchestraOwsPublicationService`.
    
Expects the GeoServer instance to be running on geOrchestra's "minimum data directory", with this shapefile:
`file:data/automated_tests/public_layer.shp`
    
Verifies the workspace and datastore are automatically created and the `FeatureTypeInfo`/`Layer` is created in GeoServer.

### DIFF
--- a/datafeeder/pom.xml
+++ b/datafeeder/pom.xml
@@ -65,6 +65,11 @@
         <scope>import</scope>
       </dependency>
       <dependency>
+        <groupId>org.geoserver.openapi</groupId>
+        <artifactId>gs-rest-java-client</artifactId>
+        <version>${gs-rest-java-client.version}</version>
+      </dependency>
+      <dependency>
         <groupId>org.projectlombok</groupId>
         <artifactId>lombok</artifactId>
         <version>${lombok.version}</version>
--- a/datafeeder/pom.xml
+++ b/datafeeder/pom.xml
@@ -13,7 +13,6 @@
   <packaging>jar</packaging>
   <name>Data-Feeder microservice</name>
   <properties>
-    <maven.test.skip>false</maven.test.skip>
     <context.name>${project.artifactId}</context.name>
     <packageDatadirScmVersion>datafeeder</packageDatadirScmVersion>
 
@@ -30,9 +29,9 @@
     <formatter-maven-plugin.version>2.10.0</formatter-maven-plugin.version>
     <fmt.skip>false</fmt.skip>
     <fmt.action>format</fmt.action><!-- or fmt.action 'validate' -->
+    <gs-rest-java-client.version>2.19-SNAPSHOT</gs-rest-java-client.version>
   </properties>
   <repositories>
-    <!-- geotools -->
     <repository>
       <id>osgeo</id>
       <name>OSGeo Nexus Release Repository</name>
@@ -145,6 +144,10 @@
     </dependencies>
   </dependencyManagement>
   <dependencies>
+    <dependency>
+      <groupId>org.geoserver.openapi</groupId>
+      <artifactId>gs-rest-java-client</artifactId>
+    </dependency>
     <dependency>
       <groupId>joda-time</groupId>
       <artifactId>joda-time</artifactId>
--- a/datafeeder/pom.xml
+++ b/datafeeder/pom.xml
@@ -88,6 +88,13 @@
         <groupId>org.apache.commons</groupId>
         <artifactId>commons-vfs2</artifactId>
         <version>2.7.0</version>
+        <exclusions>
+          <!-- hadoop transitive dependency carries over an old version of okhttp that interferes with the geoserver rest client -->
+          <exclusion>
+            <groupId>com.squareup.okhttp</groupId>
+            <artifactId>okhttp</artifactId>
+          </exclusion>
+        </exclusions>
       </dependency>
       <dependency>
         <groupId>org.apache.commons</groupId>
@@ -95,15 +102,22 @@
         <version>1.20</version>
       </dependency>
       <dependency>
-        <groupId>org.geotools</groupId>
-        <artifactId>gt-geopkg</artifactId>
-        <version>${gt.version}</version>
-      </dependency>
-      <dependency>
         <groupId>org.geotools.jdbc</groupId>
         <artifactId>gt-jdbc-postgis</artifactId>
         <version>${gt.version}</version>
       </dependency>
+      <dependency>
+        <groupId>org.geotools</groupId>
+        <artifactId>gt-shapefile</artifactId>
+        <version>${gt.version}</version>
+      </dependency>
+      <!-- 
+      <dependency>
+        <groupId>org.geotools</groupId>
+        <artifactId>gt-geopkg</artifactId>
+        <version>${gt.version}</version>
+      </dependency>
+       -->
       <dependency>
         <groupId>org.geotools</groupId>
         <artifactId>gt-geojson</artifactId>
@@ -206,22 +220,20 @@
     <dependency>
       <groupId>org.geotools</groupId>
       <artifactId>gt-shapefile</artifactId>
-      <version>${gt.version}</version>
     </dependency>
     <dependency>
       <groupId>org.geotools</groupId>
       <artifactId>gt-geojson</artifactId>
-      <version>${gt.version}</version>
     </dependency>
+    <!--
     <dependency>
       <groupId>org.geotools</groupId>
       <artifactId>gt-geopkg</artifactId>
-      <version>${gt.version}</version>
     </dependency>
+    -->
     <dependency>
       <groupId>org.geotools.jdbc</groupId>
       <artifactId>gt-jdbc-postgis</artifactId>
-      <version>${gt.version}</version>
     </dependency>
     <!--SpringFox dependencies -->
     <dependency>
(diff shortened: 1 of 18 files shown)


datafeeder/pom.xml
```diff
@@ -84,22 +88,36 @@
         <groupId>org.apache.commons</groupId>
         <artifactId>commons-vfs2</artifactId>
         <version>2.7.0</version>
+        <exclusions>
+          <!-- hadoop transitive dependency carries over an old version of okhttp that interferes with the geoserver rest client -->
```
fvanderbiest, line 92: the nightmare ...

datafeeder/src/main/java/org/georchestra/datafeeder/service/publish/impl/GeoServerRemoteService.java
```diff
@@ -0,0 +1,222 @@
...
+        } finally {
+            LOCKS.writeLock(workspaceName).unlock();
+        }
+    }
+
+    private void configureNamespace(final WorkspaceInfo wsInfo) {
+        final NamespacesClient namespaces = this.client().namespaces();
+
+        final String workspaceName = wsInfo.getName();
+        String defaultNamespaceURI = "http://georchestra.org/datafeeder/" + wsInfo.getName();
```
fvanderbiest, line 126: Says, https ?

datafeeder/src/main/java/org/georchestra/datafeeder/service/publish/impl/GeorchestraOwsPublicationService.java
```diff
@@ -0,0 +1,228 @@
...
+
+        geoserver.update(fti);
+    }
+
+    private MetadataLinkInfo buildMetadataLink(String metadataRecordId) {
+        final URI recordURI = metadataPublicationService.buildMetadataRecordURI(metadataRecordId);
+        MetadataLinkInfo info = new MetadataLinkInfo();
+        info.setId(metadataRecordId);
+        info.setContent(recordURI.toString());
+        info.setMetadataType("ISO-19139");// TODO: revisit correct value
```
fvanderbiest, line 223: IIRC, ISO19115:2003 is the one GeoServer proposes.
"An internationally-adopted schema for describing geographic information and services. It provides information about the identification, the extent, the quality, the spatial and temporal schema, spatial reference, and distribution of digital geographic data. Sponsored by the International Standards Organisation, the first edition of ISO 19115 was published in 2003."

datafeeder/src/main/java/org/georchestra/datafeeder/service/publish/impl/GeorchestraPublishingServicesConfiguration.java
```diff
@@ -32,10 +38,33 @@
...
+    }
+
+    public @Bean GeoServerClient geoServerApiClient(DataFeederConfigurationProperties props) {
+        ExternalApiConfiguration config = props.getPublishing().getGeoserver();
+        String restApiEntryPoint = config.getApiUrl().toExternalForm();
+
+        GeoServerClient client = new GeoServerClient(restApiEntryPoint);
+
+        Map<String, String> authHeaders = new HashMap<>();
+        // authHeaders.put("sec-proxy", "true");
```
fvanderbiest, line 59: sec-proxy is not mandatory for this use case indeed

datafeeder/src/main/java/org/georchestra/datafeeder/service/publish/impl/GeorchestraPublishingServicesConfiguration.java
```diff
@@ -32,10 +38,33 @@
...
+        authHeaders.put("sec-username", "datafeeder-application");
+        authHeaders.put("sec-roles", "ROLE_ADMINISTRATOR");
```
fvanderbiest, line 61: Maybe we need a discussion with @pmauduit and others about the pros and cons of directly targeting the internal geoserver endpoint as opposed to going through the security-proxy via the front LB.
Usually we do the latter (eg: mapfishapp, extractorapp both have in their properties a reference to a `geoserver_privileged_user`), but I'm not opposed to changing the way we do if everyone agrees.
The downside of the current situation is that it is a potential security hole. The downside of targeting the geoserver local service is that the module configuration will depend on the implementation (localhost:8080/geoserver or geoserver:8080/geoserver).

datafeeder/src/main/java/org/georchestra/datafeeder/service/publish/impl/GeorchestraOwsPublicationService.java
landryb: beware of this setting, and eventually try setting several metadatalink values (`text/xml`, `text/html`...) , because iirc mapstore expects a different one from mapfishapp...

also see https://github.com/georchestra/mapstore2-georchestra/issues/300 for my previous findings with mapstore.

datafeeder/src/main/java/org/georchestra/datafeeder/service/publish/impl/GeorchestraPublishingServicesConfiguration.java
landryb: :+1: please no hardcoding of username/target server :)

groldan: Hi, I came to see the comments with the idea in mind of fully externalizing the configuration for the auth method, so good timing :)
i.e., we currently have
```
publishing.geoserver.api-url=http://localhost:8080/geoserver/rest
publishing.geoserver.username=admin
publishing.geoserver.password=geoserver
```

My idea is to turn it into:

```
publishing.geoserver.api-url=http://localhost:8080/geoserver/rest
publishing.geoserver.auth.type=basic|headers

publishing.geoserver.auth.basic.username=admin
publishing.geoserver.auth.basic.password=geoserver

publishing.geoserver.auth.headers.sec-username=XXX
publishing.geoserver.auth.headers.sec-roles=YYY
publishing.geoserver.auth.headers.<keyM>=<valueN>
```

Sounds better?

landryb: yes,  i think that would work for me (for historical reasons i dont use `geoserver_privileged_user` from the LDAP, and my geoserver has a local admin user not in the LDAP..)

thanks !

groldan: See https://jira.camptocamp.com/browse/GSGEODPT43-139

datafeeder/src/main/java/org/georchestra/datafeeder/service/publish/impl/GeorchestraOwsPublicationService.java
groldan: Good point, created a separate task in order to stop overloading the scope of this one https://jira.camptocamp.com/browse/GSGEODPT43-140

datafeeder/src/main/java/org/georchestra/datafeeder/service/publish/impl/GeorchestraPublishingServicesConfiguration.java
groldan: Right, superseded by https://jira.camptocamp.com/browse/GSGEODPT43-139

datafeeder/src/main/java/org/georchestra/datafeeder/service/publish/impl/GeorchestraPublishingServicesConfiguration.java
fvanderbiest: (Landry does not have access to JIRA - which is our internal bug tracker)

landryb: and that's fine with me not accessing to such monstrosity... my last experience with geoserver's JIRA was .. painful, to say the least ;)

datafeeder/src/main/java/org/georchestra/datafeeder/service/publish/impl/GeoServerRemoteService.java
groldan: done.

datafeeder/src/main/java/org/georchestra/datafeeder/service/publish/impl/GeorchestraPublishingServicesConfiguration.java
groldan: Moved this discussion/decision to its own task. Merging in order to continue working.

PR #3306, opened 2021-02-23
[DataFeeder] Dataset upload to postgis

groldan:
Integrated in the publish workflow:
- postgis upload
- geoserver publishing


groldan:
@fgravin @fgravin, merging so you can test postigs/geoserver integration in the datafeeder/integration branch.

It may not be perfect, works with the osm shapefiles I'm testing.

Run with:

```bash
docker-compose -f docker-compose-it.yml up -d
mvn spring-boot:run
```

Issue #3279, opened 2021-02-04
geoserver 2.18


fvanderbiest:
GeoServer 2.18.2 was released on January 20th, 2021.
Would be great if we could upgrade master to this version, in order to ship it with 21.0.0

fvanderbiest:
Caution when upgrading, `s3-geotiff` has been renamed into `cog` (and performances enhanced by taking advantage of the internal file structure, rather than downloading and caching locally the file).
https://docs.geoserver.org/stable/en/user/community/cog/index.html

Might need to revisit https://github.com/georchestra/datadir/tree/master/geoserver/s3-geotiff

PR #3309, opened 2021-02-24
Upgrade GeoServer to 2.18.2

fvanderbiest:
Compiles, untested for now.
See #3279


fvanderbiest:
Seems s3-geotiff and cog extensions coexist in the release. Will require a deeper look into it.

PR #3336, opened 2021-03-31
Datafeeder: georchestra integration

groldan:
Ready to use within georchestra's default docker composition.

Needs to be run from the https://github.com/georchestra/docker/tree/datafeeder branch, which in turn sets up the config submodule from https://github.com/georchestra/datadir/tree/docker-datafeeder

### DIFF
--- a/datafeeder/src/main/java/org/georchestra/datafeeder/autoconf/GeorchestraBatchStoreConfiguration.java
+++ b/datafeeder/src/main/java/org/georchestra/datafeeder/autoconf/GeorchestraBatchStoreConfiguration.java
@@ -1,0 +1,51 @@
+/*
+ * Copyright (C) 2021 by the geOrchestra PSC
+ *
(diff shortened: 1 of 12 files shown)


fvanderbiest (approved):

landryb:
you probably know where i'm going, but are there plans for integration (doc, etc) in the **non-docker** case ? ;)

but maybe i have no reasons to be worried...

groldan:
> you probably know where i'm going, but are there plans for integration (doc, etc) in the **non-docker** case ? ;)
> 
> but maybe i have no reasons to be worried...

sort of... @pmauduit mentioned that in slack, I honestly didn't think about it before, totally slipped off my mind that's also a deployment option for georchestra.

So more than plans, there's the possibility of either:
- Converting the app to be a war file
- keep it being a single jar executable and have some sort of linux service script

but we haven't gone forward with any

PR #3351, opened 2021-04-13
[datafeeder]: Load metadata template and XSL from georchestra datadir

groldan:
With the `georchestra` spring profile enabled (default), georchestra's
datadir/datafeeder/datafeeder.properties can provide the following
configuration properties to load the template record and the XSL
transformation from an external source:

```
datafeeder.publishing.geonetwork.template-record: file:${georchestra.datadir}/datafeeder/metadata_template.xml
datafeeder.publishing.geonetwork.template-transform: file:${georchestra.datadir}/datafeeder/metadata_transform.xsl
```


groldan:
superseded by #3353

PR #3354, opened 2021-04-15
security proxy encoded headers

groldan:


landryb:
such larges changes have potential for so much regression..  i'd be extra careful with that :D

security-proxy/src/main/java/org/georchestra/security/LdapUserDetailsRequestHeaderProvider.java
```diff
@@ -46,156 +55,220 @@
...
 
     @Autowired
     private GeorchestraConfiguration georchestraConfiguration;
 
-    public LdapUserDetailsRequestHeaderProvider(LdapUserSearch userSearch, String orgSearchBaseDN,
-            Map<String, String> headerMapping) {
+    public LdapUserDetailsRequestHeaderProvider(FilterBasedLdapUserSearch userSearch, String orgSearchBaseDN) {
         Assert.notNull(userSearch, "userSearch must not be null");
-        Assert.notNull(headerMapping, "headerMapping must not be null");
+        // Q: is it ok for orgSearchBaseDN to be null?
```
fvanderbiest, line 100: Don't think so... cc @pmauduit

security-proxy/src/main/java/org/georchestra/security/LdapUserDetailsRequestHeaderProvider.java
```diff
@@ -46,156 +55,220 @@
...
+                logger.warn("Cannot find associated org with cn " + orgCn);
+            }
+        }
+        return headers;
+    }
+
+    private String loadOrgCn(String username) {
+        try {
+            // Retreive memberOf attributes
+            // WARN! (groldan) looks like _userSearch is a singleton, so we could be mixing
```
fvanderbiest, line 176: whoops

security-proxy/src/main/java/org/georchestra/security/LdapUserDetailsRequestHeaderProvider.java
pmauduit: if it's null, my comprehension of the next lines of code would be to have a more global regex pattern (e.g. "member == uid=blah,.*" instead of "member == uid=blah,ou=orgs,dc=geochestra,dc=org.*") so it does not matter that much ?

groldan:
> such larges changes have potential for so much regression.. i'd be extra careful with that :D

of course. I've been. We also have a moral obligation to fix wrong where we see it, or I've been watching too many Uncle Bob youtube videos.

security-proxy/src/main/java/org/georchestra/security/LdapUserDetailsRequestHeaderProvider.java
groldan: that would end up in `Pattern.compile("([^=,]+)=([^=,]+),null.*")`

PR #3356, opened 2021-04-19
datafeeder integration - adding a maven module to build the UI

pmauduit:
See the added README.md for the considerations of why it could make sense to be able to have the ui in a separate module and provided as a WAR package.

### DIFF
--- a/datafeeder-ui/README.md
+++ b/datafeeder-ui/README.md
@@ -1,0 +1,20 @@
+# About
+
+This directory contains a maven project to build a WAR with the frontend part
+for the datafeeder.
+
+Note: the docker image makes use of an Nginx, which makes more sense to serve
+static resources as a web ui, but since this leads to make several assumptions
+on the targeted infrastructurei (Do we have a webserver ? Which port is it
+supposed to listen on ? Where is the document root ? Which webserver to begin
+with ? Which sample configuration to provide ?), is has been preferred to
+provide this UI as another webapp, as we know that at least one servlet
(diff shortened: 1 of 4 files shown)


landryb:
for consistency with other modules, why not `datafeeder-ui.war` instead of `import.war` ?

pmauduit:
> for consistency with other modules, why not datafeeder-ui.war instead of import.war ?

For now, the web ui expects to be behind `/import/`. Is it a must to be able to put it wherever you want ?

landryb:
ah then it makes sense, it's just that some things in buildbot and ansible playbook tend to assume that the war has the same name as the folder, but i might remember wrong

pmauduit:
> ah then it makes sense, it's just that some things in buildbot and ansible playbook tend to assume that the war has the same name as the folder, but i might remember wrong

I can rename the directory to "import", but I made more sense to me to have datafeeder (backend) + datafeeder-ui (frontend).

pmauduit:
> ah then it makes sense, it's just that some things in buildbot and ansible playbook tend to assume that the war has the same name as the folder, but i might remember wrong

You can customize the final generated artifact name at several places, and it depends on the involved maven plugin. For maven-war-plugin, it normally uses the <artifactid>-<version>.war convention, but if a finalName property is defined, then it will name it following the property. This is what we actually use in the debianPackage profile so that we can have a "predictible" path in the debian package setup.

groldan (approved):
LGTM

PR #3357, opened 2021-04-19
security proxy encoded headers

groldan:
## Add support for per-service header mappings

Enhancement to allow configuring request headers on a per-target service (proxified app) fashion.
`headers-mapping.properties` can now have, besides default header mappings, service-specific mappings.

For example, the following `headers-mapping.properties` file would set up `sec-email` for all services, and `sec-lastname` only for the `analytics` proxified application:

```
    sec-email=mail
    analytics.sec-lastname=sn
```    

## Add support to base64-encode header values
    
Allow header values to be defined as base64-encoded in header-mappings.properties, by appending the `base64:` prefix to  the LDAP property name.
    
For example, a mapping of:
    
    sec-firstname=base64:givenName
    
Will result in a value of:
    
    {base64}<base64 encoded string>

## Move security header names to commons' SecurityHeaders
    
Introduce commons' `org.georchestra.commons.security.SecurityHeaders`
holding the common request header names used in security-proxy.
    
Introduce a `SecurityHeaders.decode(String):String` utility method and
change all calls to `HttpServletRequest.getHeader(String)` to
`SecurityHeaders.decode(request.getHeader(...))`, to account for the
 fact that headers could come encoded to preserve non ISO-8869-1
 characters, as defined in security-proxy's `headers-mapping.properties`
 (for example, `sec-username=base64:givenName` instead of just
`sec-username=givenName`).

### DIFF
--- a/console/src/test/java/org/georchestra/console/ws/editorgdetails/EditOrgDetailsFormControllerTest.java
+++ b/console/src/test/java/org/georchestra/console/ws/editorgdetails/EditOrgDetailsFormControllerTest.java
@@ -1,6 +1,20 @@
 package org.georchestra.console.ws.editorgdetails;
 
-import com.google.common.io.ByteStreams;
+import static org.georchestra.commons.security.SecurityHeaders.SEC_ORG;
+import static org.georchestra.commons.security.SecurityHeaders.SEC_ROLES;
+import static org.georchestra.commons.security.SecurityHeaders.SEC_USERNAME;
+import static org.junit.Assert.assertEquals;
+import static org.junit.Assert.assertTrue;
+
+import java.io.ByteArrayOutputStream;
+import java.io.IOException;
+import java.io.InputStream;
+import java.io.OutputStream;
+import java.util.Arrays;
+import java.util.Base64;
+import java.util.HashMap;
+import java.util.List;
+
 import org.georchestra.console.ds.AccountDao;
 import org.georchestra.console.ds.OrgsDao;
 import org.georchestra.console.ds.RoleDao;
@@ -22,17 +36,7 @@ import org.springframework.validation.MapBindingResult;
 import org.springframework.web.bind.WebDataBinder;
 import org.springframework.web.multipart.MultipartFile;
 
-import java.io.ByteArrayOutputStream;
-import java.io.IOException;
-import java.io.InputStream;
-import java.io.OutputStream;
-import java.util.Arrays;
-import java.util.Base64;
-import java.util.HashMap;
-import java.util.List;
-
-import static org.junit.Assert.assertEquals;
-import static org.junit.Assert.assertTrue;
+import com.google.common.io.ByteStreams;
 
 public class EditOrgDetailsFormControllerTest {
     private EditOrgDetailsFormController ctrl;
@@ -101,9 +105,9 @@ public class EditOrgDetailsFormControllerTest {
 
     @Test
     public void testSetupFormChangeUrl() throws IOException {
-        request.addHeader("sec-org", "georTest");
-        request.addHeader("sec-username", "georTest");
-        request.addHeader("sec-role", "REFERENT,THE_ADMIN");
+        request.addHeader(SEC_ORG, "georTest");
+        request.addHeader(SEC_USERNAME, "georTest");
+        request.addHeader(SEC_ROLES, "REFERENT;THE_ADMIN");
         BindingResult resultErrors = new MapBindingResult(new HashMap<>(), "errors");
         ctrl = new EditOrgDetailsFormController(orgsDao, new Validation(""));
         ctrl.logUtils = mockLogUtils;
@@ -117,9 +121,9 @@ public class EditOrgDetailsFormControllerTest {
 
     @Test
     public void testSetupFormWithImage() throws IOException {
-        request.addHeader("sec-username", "georTest");
-        request.addHeader("sec-org", "georTest");
-        request.addHeader("sec-role", "MOMO,REFERENT");
+        request.addHeader(SEC_USERNAME, "georTest");
+        request.addHeader(SEC_ORG, "georTest");
+        request.addHeader(SEC_ROLES, "MOMO;REFERENT");
         BindingResult resultErrors = new MapBindingResult(new HashMap<>(), "errors");
         ctrl = new EditOrgDetailsFormController(orgsDao, new Validation(""));
         ctrl.logUtils = mockLogUtils;
--- a/security-proxy/src/main/java/org/georchestra/security/HeadersManagementStrategy.java
+++ b/security-proxy/src/main/java/org/georchestra/security/HeadersManagementStrategy.java
@@ -28,8 +28,8 @@ import static org.georchestra.security.HeaderNames.CONTENT_LENGTH;
 import static org.georchestra.security.HeaderNames.COOKIE_ID;
 import static org.georchestra.security.HeaderNames.HOST;
 import static org.georchestra.security.HeaderNames.PROTECTED_HEADER_PREFIX;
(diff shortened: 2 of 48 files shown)


groldan:
@pmauduit @fvanderbiest a note on console's [build failure](https://github.com/georchestra/georchestra/pull/3357/checks?check_run_id=2385836600#step:7:571):

A bit of debugging shows the failure is unrelated to this pr. The test in question is changing a user's org from `psc` to `cra`. `UsersController` fails at this point (probably the `cra` org doesn't exist): 

https://github.com/georchestra/georchestra/blob/536214ffa36237360d47bdd80ca417074f60995c/console/src/main/java/org/georchestra/console/ws/backoffice/users/UsersController.java#L439

The test case executes `support.perform(put("/private/users/" + userName), ...` but doesn't check the HTTP return code or anything from the put operation. After that, the call to `support.perform(get("/private/users/" + newUserName))` expects the org to have changed but that's not the case due to the above mentioned failure.

Also, by then `UsersController` has ["unlinked"](https://github.com/georchestra/georchestra/blob/536214ffa36237360d47bdd80ca417074f60995c/console/src/main/java/org/georchestra/console/ws/backoffice/users/UsersController.java#L426) the user from its current organization, but no corrective action is taken when linking it to the new org fails.

I'm not sure how to fix that though?

groldan:
`mapfish-app` fails with this, also can't fix myself:


https://github.com/georchestra/georchestra/pull/3357/checks?check_run_id=2386075943#step:8:195

```
2021-04-20T00:23:13.8377636Z jsbuild/build.sh: 49: /home/runner/work/georchestra/georchestra/mapfishapp/jsbuild/env/bin/jsbuild: not found
2021-04-20T00:23:13.8379476Z done.
2021-04-20T00:23:13.8388103Z [NOK] jsbuild failure
2021-04-20T00:23:13.9236474Z [ERROR] Command execution failed.
2021-04-20T00:23:13.9238230Z org.apache.commons.exec.ExecuteException: Process exited with an error: 1 (Exit value: 1)
```

security-proxy/src/main/java/org/georchestra/security/LdapUserDetailsRequestHeaderProvider.java
```diff
@@ -39,163 +51,323 @@
...
 
     @Autowired
     private GeorchestraConfiguration georchestraConfiguration;
 
-    public LdapUserDetailsRequestHeaderProvider(LdapUserSearch userSearch, String orgSearchBaseDN,
-            Map<String, String> headerMapping) {
+    public LdapUserDetailsRequestHeaderProvider(FilterBasedLdapUserSearch userSearch, String orgSearchBaseDN) {
         Assert.notNull(userSearch, "userSearch must not be null");
-        Assert.notNull(headerMapping, "headerMapping must not be null");
+        // Q: is it ok for orgSearchBaseDN to be null?
```
pmauduit, line 134: if null, then the pattern below (orgSeachMemberOfPattern) will just be less accurate, right ?

security-proxy/src/main/java/org/georchestra/security/LdapUserDetailsRequestHeaderProvider.java
```diff
@@ -39,163 +51,323 @@
...
         this._userSearch = userSearch;
-        this._headerMapping = headerMapping;
         this.orgSearchBaseDN = orgSearchBaseDN;
-
-        this.pattern = Pattern.compile("([^=,]+)=([^=,]+)," + orgSearchBaseDN + ".*");
+        this.orgSeachMemberOfPattern = Pattern.compile("([^=,]+)=([^=,]+)," + orgSearchBaseDN + ".*");
```
pmauduit, line 137: typo orgSearchMemberOfPattern (with a 'r' to search)

security-proxy/src/main/java/org/georchestra/security/LdapUserDetailsRequestHeaderProvider.java
```diff
@@ -39,163 +51,323 @@
...
+        }
+        return headers;
+    }
+
+    private String loadOrgCn(String username) {
+        try {
+            // Retreive memberOf attributes
+            // WARN! (groldan) looks like _userSearch is a singleton, so we could be mixing
+            // up setReturningAttributes from concurrent requests (we're synchronized on
+            // session here, not on _userSearch)
```
pmauduit, line 249: for searching the LDAP, would it make more sense to create a new "userSearch" object each time instead of using a member object of the instance, and taking the risk of a race condition ?

pmauduit (approved):

console/src/test/java/org/georchestra/console/ws/editorgdetails/EditOrgDetailsFormControllerTest.java
```diff
@@ -101,9 +105,9 @@ public void testInitForm() {
 
     @Test
     public void testSetupFormChangeUrl() throws IOException {
-        request.addHeader("sec-org", "georTest");
-        request.addHeader("sec-username", "georTest");
-        request.addHeader("sec-role", "REFERENT,THE_ADMIN");
```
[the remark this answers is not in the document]
groldan: seemed odd to me too. IIRC it changed from sec-org to sec-orgs some time ago? in any case it didn't seem to affect the test case

security-proxy/src/main/java/org/georchestra/security/LdapUserDetailsRequestHeaderProvider.java
groldan: if `null` then the pattern will be `"([^=,]+)=([^=,]+),null.*"`, and won't match anything?

groldan: I will just add a null check for the constructor to fail fast. a null orgSearchBaseDN will make an unusable pattern, and `buildStandardOrganizationHeaders` will build a dn like `cn=org,null`:
```
 String dn = "cn=" + orgCn + "," + this.orgSearchBaseDN;
```

Means null is never actually used or someone would have complained, besides being out of scope for this pr.

groldan: Done: 920d8a1bbbf0400f03b0356b365b8b39e77042de

security-proxy/src/main/java/org/georchestra/security/HeadersManagementStrategy.java
```diff
@@ -259,10 +291,12 @@ public void handleRequestCookies(HttpServletRequest originalRequest, HttpRequest
...
-        headersLog.append("\t" + COOKIE_ID);
-        headersLog.append("=");
-        headersLog.append(cookies);
-        headersLog.append("\n");
+        if (headersLog != null) {
+            headersLog.append("\t" + COOKIE_ID);
+            headersLog.append("=");
+            headersLog.append(cookies);
+            headersLog.append("\n");
+        }
```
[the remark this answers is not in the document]
groldan: good catch. Done ef7a120fe746f9897b2f8010cac607824f97dd5b

security-proxy/src/main/java/org/georchestra/security/LdapUserDetailsRequestHeaderProvider.java
groldan: Fixed. 2e283fc9e5a94395359a31956b871cab3b6999d6

security-proxy/src/main/java/org/georchestra/security/LdapUserDetailsRequestHeaderProvider.java
groldan: Fixed: 2e283fc9e5a94395359a31956b871cab3b6999d6

PR #3365, opened 2021-04-27
[datafeeder] Separate calls to build md record identifier and xml endpoint

groldan:


datafeeder/src/main/java/org/georchestra/datafeeder/service/geonetwork/GeoNetworkRemoteService.java
```diff
@@ -205,4 +197,13 @@ public URI buildMetadataRecordURI(@NonNull String recordId) {
         builder.query(queryString);
         return builder.build().toUri();
     }
+
+    // e.g.
+    // http://localhost:28080/geonetwork/srv/api/0.1/records/2bd68e79-7fb1-443b-8709-3570b19a3d6f/formatters/xml
+    public URI buildMetadataRecordXmlURI(@NonNull String recordId) {
+        URL publicURL = this.config.getPublicUrl();
+        String xmlMdRecord = String.format("%s/srv/api/0.1/records/%s/formatters/xml", publicURL, recordId);
```
fxprunayre, line 205: Don't put `0.1` in API path - it is not needed and it has been removed in 4. Also not that requesting %s/srv/api/0.1/records/%s with Accept application/xml should also return an XML document. Depending on the usage of this URL.

groldan: roger on the 0.1 path. Given the url is gonna be used as a GET request from wherever, and need to make sure the outcome is xml, we probably leave the `/formatters/xml` part, right?

fxprunayre: > we probably leave the `/formatters/xml` part, right?

Depends who is going to use that URL (not really aware of the context). If you expect apps/users to get an XML doc, then keep it.

PR #3366, opened 2021-04-27
datafeeder-ui - serves the index.html file on dynamic endpoints

pmauduit:
the JS app will manage the actual requested path in some cases, e.g.  if
`/import/<uuid>/validation` is requested, then we have to return the
index.html content to the client, which will parse JS-side the url
fragment and display the according page.

The behaviour should be slightly the same as the suggested nginx
configuration from the geonetwork-ui repository, see:

https://github.com/georchestra/geonetwork-ui/blob/georchestra/nginx-default.conf

Note: I first tried to do it in a more "bare-metal servlet" approach,
thinking that the servlet-mapping's url-pattern parameters in the
web.xml would suffice, but it actually does not support regexps and is
more of a simple globbing. Then the "fastest" way to figure out how I
could send whatever I wanted to the client was to do spring-mvc, and
introducing spring as a dependency. Maybe a urlrewrite would have been
lighter.

Note2: I could have converted the project into a spring-boot one, but
the idea is still to finish with a webapp war that could be easily
deployed into an existing servlet container for non-dockerized approach.

Tests: mainly runtime-tested via `mvn clean package jetty:run` with
loads of curl.

Also tested in a simple jetty docker image, dropping the webapp into
`/var/lib/jetty/webapps`, replacing the import container by this new one, and I
was able to publish a dataset.

### DIFF
--- a/datafeeder-ui/pom.xml
+++ b/datafeeder-ui/pom.xml
@@ -12,10 +12,42 @@
   <artifactId>datafeeder-ui</artifactId>
   <packaging>war</packaging>
   <name>Data-Feeder (web ui)</name>
-
+  <dependencies>
+      <dependency>
+          <groupId>org.springframework</groupId>
+          <artifactId>spring-webmvc</artifactId>
+      </dependency>
+      <dependency>
+          <groupId>org.springframework</groupId>
+          <artifactId>spring-context</artifactId>
+      </dependency>
+      <dependency>
+          <groupId>commons-io</groupId>
+          <artifactId>commons-io</artifactId>
+          <version>2.8.0</version>
+      </dependency>
+      <dependency>
+          <groupId>javax.servlet</groupId>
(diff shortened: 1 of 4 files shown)


groldan:
Are the UUID patterns in 
```
    @RequestMapping("/{uuid:[0-9a-fA-F]{8}-[0-9a-fA-F]{4}-[0-9a-fA-F]{4}-[0-9a-fA-F]{4}-[0-9a-fA-F]{12}}")<br class="Apple-interchange-newline">
```
to distinguish from other static resources that might be requested at the root?

Can you add the docs to the class so anyone can know what it's about?

pmauduit:
> Are the UUID patterns in [...] to distinguish from other static resources that might be requested at the root?

yes

I actually introduced spring, but maybe using urlrewrite would have been lighter, thinking of it ...

> Can you add the docs to the class so anyone can know what it's about?

Yes

PR #3378, opened 2021-05-06
[datafeeder] Fix single line multiple variables resolution on email templates

groldan:
Fix variable names resolution when there is more than one variable on
the same line of email templates.

Change templates to be closer to what the real ones shall be.

Better handle job failures.  We need to leave the spring-batch job complete normally or the
publishStatus/error won't be persisted, as spring-batch and our JPA
repositories share the same entity manager. If a publishing process failed and
the exception was re-thrown, spring-batch would abort the current
transaction.

### DIFF
--- a/datafeeder/src/main/java/org/georchestra/datafeeder/autoconf/GeorchestraEmailFactory.java
+++ b/datafeeder/src/main/java/org/georchestra/datafeeder/autoconf/GeorchestraEmailFactory.java
@@ -358,9 +358,9 @@ public class GeorchestraEmailFactory implements DatafeederEmailFactory {
         return propertyName;
     }
 
-    private Set<String> extractVariableNames(String messageTemplate) {
-        Matcher matcher = Pattern.compile("\\$\\{.*\\}").matcher(messageTemplate);
+    static Set<String> extractVariableNames(String messageTemplate) {
         Set<String> varNames = new TreeSet<>();
+        Matcher matcher = Pattern.compile("\\$(\\{(\\w+\\.)*\\w+\\})").matcher(messageTemplate);
         while (matcher.find()) {
             String varName = matcher.group(0);
             varNames.add(varName);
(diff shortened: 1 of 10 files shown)


groldan:
Also see improved templates at https://github.com/georchestra/datadir/commit/3236a8fa41ef6db7a63ff48cb1de8644eb5d91b1

datafeeder/src/main/java/org/georchestra/datafeeder/autoconf/GeorchestraEmailFactory.java
```diff
@@ -358,9 +358,9 @@ private String variableToPropertyName(final String varName) {
         return propertyName;
     }
 
-    private Set<String> extractVariableNames(String messageTemplate) {
-        Matcher matcher = Pattern.compile("\\$\\{.*\\}").matcher(messageTemplate);
+    static Set<String> extractVariableNames(String messageTemplate) {
         Set<String> varNames = new TreeSet<>();
+        Matcher matcher = Pattern.compile("\\$(\\{(\\w+\\.)*\\w+\\})").matcher(messageTemplate);
```
pmauduit, line 363: I'd have expected something more "spring-ish" for templating, but I"m pretty sure you already checked ;-) (thymeleaf wasn't doing the job ?)

pmauduit (approved):

datafeeder/src/main/java/org/georchestra/datafeeder/autoconf/GeorchestraEmailFactory.java
groldan: you're probably right. Often one opts for custom instead of learning a new technology, especially when under pressure.

pmauduit:
Runtime tested without the patch unintentionnally, then I can confirm it fixes the issue after having updated / regenerated the docker image.

PR #3382, opened 2021-05-11
sp - splits user-agents string for preemptive basic-auth

### LINKED ISSUE #3375: SP - basic-authentication by user-agent does not work if multiple user-agents are provided in sp.properties
Since this commit:
https://github.com/georchestra/datadir/commit/fc34b8142c496452d54cb57880e8109cd2e98963

we are now relying on a coma-separated list of user-agent patterns, but the whole string is considered as a unique pattern in the current state of master:

https://github.com/georchestra/georchestra/blob/master/security-proxy/src/main/webapp/WEB-INF/applicationContext-security.xml#L102

It's probably missing some SpEL to split the string by commas, e.g.:

```
#{'${userAgents:}'.split(',')}
```

pmauduit:
Fixes #3375


fvanderbiest:
Thanks a bunch !

PR #3389, opened 2021-05-12
Fix gs-importer profile

jeanpommier:
importer extension has been removed from standard GeoServer build. The
profile was kept, but did not build anymore, because the name changed.


fvanderbiest:
to be backported ?

jeanpommier:
Already applied on 20.1.x through https://github.com/georchestra/georchestra/pull/3392
20.0 should not be concerned as it's still using GS 2.17.5

PR #3407, opened 2021-05-28
[datafeeder] publish metadata records to the users org GN group

groldan:
Remaining task : "The metadata is inserted into the group which
already exists due to GeoNetwork's LDAP sync (which currently
guarantees that a geonetwork group exists for the publishing organization)"


groldan:
superseded by #3408

PR #3021, opened 2020-04-02
Remove UTF8 filter. Fix web.xml encoding in some webapps.

RemiDesgrange:

PR #3410, opened 2021-05-31
Console manager - fix encoding pb and deactivate save button on pending user

isanchez-c2c:
- set locale to UTF-8 in console : PR #3021 led to a regression in console manager : jetty's default encoding for the page is ISO-8859-* but we forced the encoding of the templates to UTF-8. When removing that, the encoding for the templates became the page default ones. With this PR we set the default encoding for the page to UTF-8.

Screenshots from 20.1.1 :
contenttypetemplates
contenttypeiso

- deactivate save button on pending user : the user needs to be confirmed as a user first (there was a possible confusion between save and confirm)


tonio (approved):
No idea for the backend side (-: looks good for the button.

fvanderbiest (approved):
LGTM, to be cherry-picked onto 20.1.x branch as well.

console/src/main/webapp/WEB-INF/web.xml
```diff
@@ -84,4 +84,164 @@
...
+        </locale-encoding-mapping>
+        <locale-encoding-mapping>
+            <locale>zh</locale>
+            <encoding>UTF-8</encoding>
+        </locale-encoding-mapping>
+        <locale-encoding-mapping>
+            <locale>zh_TW</locale>
+            <encoding>UTF-8</encoding>
+        </locale-encoding-mapping>
+    </locale-encoding-mapping-list>
```
pmauduit, line 246: I did not get why you did not use the filter from spring, as we are sure to have the necessary code already into the webapp. It'd be less code than specifiying a default charset for each locale, that the browser is not even guaranteed to send (even if it will be very unlikely).

The filter from Spring that was removed earlier will do the job with less configuration needed from the web.xml:
https://github.com/spring-projects/spring-framework/blob/main/spring-web/src/main/java/org/springframework/web/filter/CharacterEncodingFilter.java#L187-L202

isanchez-c2c: ok, I agree it will be prettier :) (I was not so sure about what was the more logic)

fvanderbiest:
@isanchez-c2c can you also fix 20.1.x ? (as I understand it, there's no use for 20.0.x)

PR #3412, opened 2021-06-02
[datafeeder] Properly handle upload files with spaces

groldan:
Using `java.nio.Path.toUri().toURL()` to create the URL argument for
GeoTools Shapefile DataStore produced URL encoded file names (e.g.
`file%20name.shp` instead of `file name.shp`, which afterwards results
in the Shapefile DataStore.getTypeNames() to return `file%20name` as the
native type name.


groldan:
Merging to proceed with other fixes on top of this one. Let me know if it fails review/manual testing.

PR #3424, opened 2021-06-07
[datafeeder] Log which charset shapefiles are being read with when importing to PostGIS

groldan:
In this patch:

* Default to UTF-8 when loading the source shapefile before importing to PostGIS
* Log which charset shapefiles are being read with when importing to PostGIS
* Use a smaller java base docker image
* Allow running datafeeder in IT docker composition
* Fix non-disposed datastore due to open FeatureReader

Tried to reproduce the reported bug (GSGEODPT43-158) where a dataset "labeled as UTF-8" in the `importer` UI ends up with mangled characters.

Initially thought I reproduced, but I was actually sending the wrong request through the Swagger UI. I.e.
```
{
 "encoding": "UTF-8",
  "datasets": [
    {
      "metadata": {
...
```
Instead of
```
{
  "datasets": [
    {
      "encoding": "UTF-8",
      "metadata": {
...
```

And was confused because it would work locally but not in docker, but I was sending a good request on the local test, and a wrong one on the docker test.

Tried several things to figure out if it was a linux/docker/jdk version issue.

My conclusion is that somehow the UI may not be sending the `encoding` property.
I still can't get past the "Select one or more tags that fit your data" page in the UI, it pops up a "Please fill the mandatory fields" JS alert dialog.

I'll need someone with a working UI to test it out.

If working properly, you should see a log entry like the following in the `datafeeder's` logs:

- Running 
```
INFO 1 --- [         task-4] o.g.datafeeder.service.DatasetsService   : Loading source data from /tmp/datafeeder/uploads/805e6f57-2849-4716-aba0-10a015819b22/t_troncon_line.shp using requested character encoding UTF-8
```

i.e.
```
Loading source data from <path> using requested character encoding UTF-8
```
Otherwise, the message will be either

```
Loading source data from <path> using native character encoding <code from .cpg file>
```
or
```
Loading source data from <path> default fallback character encoding UTF-8
```


datafeeder/src/docker/Dockerfile
```diff
@@ -1,11 +1,11 @@
-FROM openjdk:8-jre
+FROM bellsoft/liberica-openjdk-alpine-musl:11
```
fvanderbiest, line 1: I'm a bit concerned by this move for production use (image is 2 months old)
cc @pmauduit

fvanderbiest (approved):
LGTM otherwise

datafeeder/src/docker/Dockerfile
groldan: Motivated by trying a thousand things that could make the bug show in docker and not in a local app run. We can get rid of it.

groldan: @pmauduit would you be so kind of reverting the Dockerfile commit for me? I'm unavailable until afternoon

pmauduit: sure

Issue #3405, opened 2021-05-28
doc - enrich the setup doc regarding datafeeder


fvanderbiest:
https://github.com/georchestra/georchestra/tree/master/docs/setup

landryb:
cf georchestra/ansible#77 ;) and ... some github action or buildbot knob to produce wars ?

pmauduit:
> cf georchestra/ansible#77 ;) and ... some github action or buildbot knob to produce wars ?

I don't really see the point of using GH actions here if we cannot store the artifact. I mean github provides a maven repository, but it requires to be authenticated, so I would go for our own CI on our server for such things.

Or we could also use a gh actions with a secret config.xml file for maven, which would allow us to publish the DF artifact onto the artifactory ? It could be nice, just thinking out loud. If so we could probably go further with the other artifacts as well.

landryb:
or .. add the missing plumbing to produce a debian package like things are done for other modules ? :)

pmauduit:
> add the missing plumbing to produce a debian package like things are done for other modules ?

Well there would be 2 different modules to consider:
* the datafeeder (spring boot app, so it would slightly differ compared to the other georchestra modules which can already be packaged as a debian package)
* the frontend (import.war, should be similar to the other ones)

fvanderbiest:
> or .. add the missing plumbing to produce a debian package like things are done for other modules ? :)

If you're ready to fund this, that would ease things a lot :-)
(considering we already invested a lot on the datafeeder app)

landryb:
> > or .. add the missing plumbing to produce a debian package like things are done for other modules ? :)
> 
> If you're ready to fund this, that would ease things a lot :-)
> (considering we already invested a lot on the datafeeder app)

who needs funding when you can have a PR instead ? ;)

pmauduit:
Added some doc specific to the DF, see https://github.com/georchestra/georchestra/pull/3656/files

fvanderbiest:
We forgot to modify https://github.com/georchestra/georchestra/blob/master/docs/setup/apache.md

... but should we maintain this "manual install" documentation as it is ? or should we remove it, in favor of the ansible / docker setups ?

PR #3427, opened 2021-06-10
debianize datafeeder & datafeeder-ui (#3405)

landryb:
a bit WIP, but produces two debian packages:
- `datafeeder/target/georchestra-datafeeder_21.0+202106102113-1_all.deb`
-  `datafeeder-ui/target/georchestra-datafeeder-ui_99.master.202106101806~ba0ad6e-1_all.deb`

first one ships the backend and the datadir subdir, + a systemd service. versioning is wrong since the `<parent>` bit was commented out in b2495efe it cant reach `project.packageVersion` defined in the parent. @groldan, is it still necessary ?

second one ships `import.war`.

untested at runtime yet, but will allow improving the integration discussed in #3405, and rework what was done in georchestra/ansible#77

at some point will `datafeeder-ui/pom.xml` point to a fixed commit in the https://github.com/georchestra/geonetwork-ui remote repo ? otherwise the build isnt much reproducible... (yeah i know its still wip :)

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -99,7 +99,7 @@ deb-build-geowebcache: war-build-geowebcache
 	mvn package deb:package -pl geowebcache-webapp -PdebianPackage -DskipTests -Dfmt.skip=true ${DEPLOY_OPTS}
 
 deb-build-georchestra: war-build-georchestra build-deps deb-build-geoserver deb-build-geowebcache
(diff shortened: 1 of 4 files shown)


pmauduit:
> @groldan, is it still necessary ?

Unfortunately I'd expect so, the root pom is coming with an incompatible spring-boot version with the one used in the DF, that was the reason why the parent link has been commented out in the first place, and I don't think the situation evolved since then.

pmauduit (approved):
LGTM

fvanderbiest:
Thanks Landry !

PR #3428, opened 2021-06-10
[datafeeder] Add missing properties to published GeoServer layers

groldan:
Add AttributionInfo to published GeoServer layer, and client cache settings to the feature type

Depends on https://github.com/geoserver/geoserver/pull/5103

### DIFF
--- a/datafeeder/src/main/java/org/georchestra/datafeeder/service/publish/impl/GeorchestraOwsPublicationService.java
+++ b/datafeeder/src/main/java/org/georchestra/datafeeder/service/publish/impl/GeorchestraOwsPublicationService.java
@@ -49,6 +49,7 @@ import org.geoserver.openapi.model.catalog.KeywordInfo;
 import org.geoserver.openapi.model.catalog.LayerInfo;
 import org.geoserver.openapi.model.catalog.MetadataEntry;
 import org.geoserver.openapi.model.catalog.MetadataLinkInfo;
+import org.geoserver.openapi.model.catalog.MetadataLinks;
 import org.geoserver.openapi.model.catalog.MetadataMap;
 import org.geoserver.openapi.model.catalog.NamespaceInfo;
 import org.geoserver.openapi.model.catalog.ProjectionPolicy;
@@ -56,6 +57,7 @@ import org.geoserver.openapi.model.catalog.WorkspaceInfo;
 import org.geoserver.openapi.v1.model.DataStoreResponse;
 import org.geoserver.openapi.v1.model.Layer;
 import org.springframework.beans.factory.annotation.Autowired;
+import org.springframework.http.MediaType;
 
 import lombok.NonNull;
 import lombok.extern.slf4j.Slf4j;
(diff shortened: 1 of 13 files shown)


groldan:
Waiting for geoserver's nightly build to publish the changes at `https://github.com/geoserver/geoserver/pull/5103` to maven

datafeeder/src/main/java/org/georchestra/datafeeder/service/publish/impl/GeorchestraOwsPublicationService.java
```diff
@@ -223,9 +257,13 @@ private FeatureTypeInfo buildPublishingFeatureType(String workspace, String data
             ft.setProjectionPolicy(ProjectionPolicy.FORCE_DECLARED);
         }
 
+        // make the layer cacheable
+        MetadataMap mdmap = new MetadataMap();
+        mdmap.addEntryItem(new MetadataEntry().atKey("cacheAgeMax").value("3600"));
```
fvanderbiest, line 262: Can it be a setting ?

fvanderbiest (approved):
Looks good to me ! Thanks Gabriel.
So we need to patch our geoserver fork before the fixes to the rest client are released (in 2.19.x ?)

landryb:
> Looks good to me ! Thanks Gabriel.
> So we need to patch our geoserver fork before the fixes to the rest client are released (in 2.19.x ?)

and stress somewhere in the doc that datafeeder relies on georchestra-geoserver :)

groldan:
> So we need to patch our geoserver fork before the fixes to the rest client are released (in 2.19.x ?)
Nope, we use 20.0-SNAPSHOT already. It's safe cause we're the only ones using that community module

datafeeder/src/main/java/org/georchestra/datafeeder/service/publish/impl/GeorchestraOwsPublicationService.java
groldan: Done: 25c8c2e

groldan:
> `cacheSeconds.intValue() <= 0` ?

ok, b61fbaf. Merging.

landryb:
> Nope, we use 20.0-SNAPSHOT already. It's safe cause we're the only ones using that community module

well that had lead to a build failure of master before this PR (need to recheck) so it isnt that safe :)

```
[ERROR] /data/src/georchestra/georchestra/datafeeder/src/main/java/org/georchestra/datafeeder/service/publish/impl/GeorchestraOwsPublicationService.java:[269,13] cannot find symbol
[ERROR]   symbol:   method setId(java.lang.String)
[ERROR]   location: variable info of type org.geoserver.openapi.model.catalog.MetadataLinkInfo
```

groldan:
The question was if "we need to patch our geoserver fork before the fixes to the rest client are released (in 2.19.x ?)"
And it's safe in that regard, we don't need to patch geoserver, because we're using the client library, which doesn't require any server-side change.
It's also safe to use the -SNAPSHOT dependency instead of waiting for a stable release because, by one side, we're the only ones using that client library, and by the other, even if there were a stable geoserver release, the community module wouldn't be part of it, so we'd still be using the SNAPSHOT version.
But of course, if we needed to make API changes to the client library in order to fix this issue, we need to get it in sync (i.e. master vs master).

PR #3438, opened 2021-06-14
[datafeeder] Allow overriding native CRS on getBounds/getSampleFeature

groldan:
When an uploaded shapefile doesn't provide a .prj file, reprojecting to
the target CRS is impossible.

This API change allows the getBounds and getSampleFeature end points to
provide a "srsOverride" query parameter in addition to the "srs"
parameter, in order to define the source and target CRS for the
reprojection.

It also simplifies the semantics by removing the "srs_reproject"
parameter:

- If the "srs" parameter is provided, reprojection DOES HAPPEN.
- If there's no native CRS, it can be overridden with the "srsOverride"
  parameter. Otherwise, an error is thrown since it's not possible to
reproject from an unknown CRS.

Client-side logic shall be updated to:

For:

`GET /upload/{jobId}/{typeName}/bounds`
`GET /upload/{jobId}/{typeName}/sampleFeature`

Based on `GET /upload/{jobId}`'s:

* If `datasets[0].nativeBounds.crs` OR `datasets[0].nativeBounds.crs.srs` is `null`, then provide a `srsOverride` parameter.
* `GET /upload/{jobId}/{typeName}/bounds?srsOverride=EPSG:4326&srs=EPSG:3857` ignores the native CRS and interprets the native dataset as `EPSG:4326`, then reprojects to `EPSG:3857`
* `GET /upload/{jobId}/{typeName}/bounds` also returns the native CRS, so the returned `BoundingBox`'s `crs` or `crs.srs` could be `null`. In which case, `srsOverride` is mandatory to perform reprojection to `srs`.
*

### DIFF
--- a/datafeeder/api.yaml
+++ b/datafeeder/api.yaml
@@ -129,6 +129,11 @@ paths:
       operationId: getSampleFeature
       description: Obtain a sample dataset feature in GeoJSON format, optionally specifying a feature index, crs, and/or dataset's character encoding.
                    The response encoding is always UTF-8. The 'encoding' parameter can be used to force reading the native data in a different charset.
+                   The returned geometry CRS is controlled by the "srs", and "srsOverride" query parameters.
+                   If none is provided, the geometry is returned as-is, in the dataset's native CRS (possibly null).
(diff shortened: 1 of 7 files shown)


fvanderbiest (approved):
Thanks Gabriel, this looks good !

Issue #3459, opened 2021-06-23
20.1.3 / master - geoserver dependencies version mismatch


pmauduit:
the `spring-boot-dependencies` parent pom defined in our root pom is forcing some of the versions used in our artifacts:
https://github.com/georchestra/georchestra/blob/master/pom.xml#L5-L13

The problem is with the GeoServer 2.18.3, we are supposed to use geotools 24.3, which - for the geopkg module- should transitively make use of a version 3.34.0 of sqlite-jdbc. But the version is forced by the previous spring-boot-dependencies to v3.15.1, which is incompatible.

If you have a GS datadir which some geopackage datastores, it is possible that the mentioned geoserver version from georchestra won't start, with this error (at webapp startup or when trying to query a layer):

```
[stacktrace]
Caused by:
java.lang.NoSuchMethodError: org.sqlite.Function.create(Ljava/sql/Connection;Ljava/lang/String;Lorg/sqlite/Function;II)V
```

Also, the GF version in our pom does not match the one from upstream GS:
https://github.com/georchestra/georchestra/blame/20.1.3/geoserver/pom.xml#L16
vs
https://github.com/georchestra/geoserver/blob/2.18.3-georchestra/src/pom.xml#L122

pmauduit:
Build failed because of parameters coming from the georchestra root pom.

pmauduit:
on both 20.1.x and latest docker image, the wrong sqlite-jdbc jar is still present.

pmauduit:
Build passed on master as well, closing.

PR #3460, opened 2021-06-23
Fixing geoserver dependencies calculation (#3459)

pmauduit:
See issue #3459  for the context of this commit. Main proposal here, apart from fixing the versions required by GS upstream, is to unplug the geoserver-georchestra root pom from the georchestra root pom.

Note: checking the dependencies against an official GS webapp from sourceforge, there are still some differences after build between geor &  upstream:

```
upstream: commons-collections-3.2.2.jar
geor:     commons-collections-3.2.1.jar

upstream: slf4j-api-1.6.4.jar
geor:     slf4j-api-1.7.21.jar

upstream: spring-tx-5.1.20.RELEASE.jar
geor:     spring-tx-4.2.5.RELEASE.jar

upstream: xml-apis-1.4.01.jar
geor:     xml-apis-1.3.04.jar
```

These differences go beyond my maven understanding (asking @groldan for help :)), as I cannot track down why they appear in these specific versions in our custom build.

Also I cannot get why we have in the georchestra version junit as a dependency into the classpath, but it seems to be a dependency from gt-geojson.

### DIFF
--- a/.github/workflows/geoserver.yml
+++ b/.github/workflows/geoserver.yml
@@ -54,13 +54,13 @@ jobs:
     - name: "Building docker image with native security"
       if: github.repository == 'georchestra/georchestra'
       working-directory: geoserver/webapp
-      run: ../../mvnw --no-transfer-progress clean package docker:build -Pdocker,colormap,mbtiles,wps-download,app-schema,control-flow,csw,feature-pregeneralized,importer,inspire,libjpeg-turbo,monitor,pyramid,wps,css,s3-geotiff,jp2k,log4j-logstash,sentry-log4j,authkey,mapstore2,sldservice -DdockerImageName=georchestra/geoserver:${{ steps.version.outputs.VERSION }} -DskipTests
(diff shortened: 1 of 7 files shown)


pmauduit:
I don't feel very comfortable with removing the parent to georchestra root pom ... Maybe forcing the problematic dependencies in the gs/pom.xml should make it, FDYT @groldan ?

pmauduit:
I actually don't think that unplugging GS from the root pom is the right way here: It will require to copy/paste some things that was implicetely provided by the georchestra root pom (url to maven repos, some shared properties across modules ...)

In more mid/long term, I think we should think about splitting the huge georchestra repositories into several different ones, to make integration of single components easier.

pmauduit:
rebased, changed strategy (reattaching to geor root pom), and forced pushed.

pmauduit:
console failing because in 20.1.x, some LDAP schemas introduced in master (but not yet in our branch) don't exist anymore.

pmauduit:
Ok, one test stays a bit flacky (had to relaunch 3 times the CI, was able to reproduce locally 1 or 2 times)

```
[ERROR] Failures: 
[ERROR]   PasswordRecoverySurvivesDatabaseRestartIT.testSurvivesDatabaseRestartPOST:89 Status expected:<200> but was:<404>

```

I cannot explain why, but @groldan : care to review it ?

groldan (approved):
forcing those couple dep versions seems easy enough, but it'll also add some maintenance burden, whenever you want to upgrade the gs version, you have to remember to repeat the whole process and amend, right?

I'd vote for the couple geor specific changes to the geoserver pom that you can just cherry-pick when upgrading to another version, instead of the maintenance burden of figuring out dep mismatches again and patching them in the pom.

That said, go for it, and treat it as a separate issue.

pmauduit:
> That said, go for it, and treat it as a separate issue.

I suggest we go for this one, and go with unplugging the root pom into master. I'll prepare a PR for this.

PR #3466, opened 2021-07-06
[datafeeder] Fix (workaround) issue with GeoServer REST API

groldan:
Change GeoServer datastore names from `datafeeder` to
`<workspace>_datafeeder`.

Using `datafeeder` as the GeoServer data store name for all workspaces
should be possible, but GeoServer has a bug where the requested
workspace/datastore name pair is ignored, and tries to use the first
datastore with the requested name ignoring the requested workspace,
when, for example, posting a FeatureType.

This resulted in unpredictable errors when publishing from georchestra
accounts in different organizations.

### DIFF
--- a/datafeeder/src/main/java/org/georchestra/datafeeder/autoconf/GeorchestraNameNormalizer.java
+++ b/datafeeder/src/main/java/org/georchestra/datafeeder/autoconf/GeorchestraNameNormalizer.java
@@ -51,6 +51,19 @@ public class GeorchestraNameNormalizer {
     }
 
     /**
+     * The datastore name is composed of {@code <workspaceName>_datafeeder} to make
+     * it unique across workspaces (with {@code workspaceName} as resolved by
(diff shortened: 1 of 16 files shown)


datafeeder/src/main/java/org/georchestra/datafeeder/autoconf/GeorchestraNameNormalizer.java
```diff
@@ -50,6 +50,19 @@ public String resolveWorkspaceName(String orgName) {
         return normalizeName(orgName).toLowerCase();
     }
 
+    /**
+     * The datastore name is composed of {@code <workspaceName>_datafeeder} to make
```
pmauduit, line 54: wouldn't it be `<datafeeder_workspaceName>` relying on the code below ?

pmauduit (commented):
Looks good

PR #3469, opened 2021-07-12
[datafeeder] Add INSPIRE topic categories to metadata record

groldan:
Simplify and fix template to add topicCategory metadata elements based on keywords
    
As gmd:keywords come from the INSPIRE thesaurus, a gmd:topicCategory is
addedd for each kewyword that matches an entry in themes.rdf.
    
For example, for a `themes.rdf` entry like:
```xml
    <rdf:Description rdf:about="3">
        <rdf:type rdf:resource="http://rdfdata.eionet.europa.eu/inspirethemes/ontology/InspireTheme"/>
        ...
        <skos:definition xml:lang="en">Names of areas, regions, ...</skos:definition>
        <skos:prefLabel xml:lang="en">Geographical names</skos:prefLabel>
        ...
    </rdf:Description>
```
    
And a mapping definition like:
```xml
      <xsl:variable name="topiccat-map">
        <entry topiccatcode="location" rdf:about="3" />
        <entry topiccatcode="boundaries" rdf:about="4" />
        ....
```

A keyword `Geographical names` will add an element like:

```xml
      <gmd:topicCategory>
        <gmd:MD_TopicCategoryCode>location</gmd:MD_TopicCategoryCode>
      </gmd:topicCategory>
```


datafeeder/src/main/resources/md_record_template.xml
```diff
@@ -190,11 +190,8 @@
       <gmd:language>
         <gmd:LanguageCode codeList="http://www.loc.gov/standards/iso639-2/" codeListValue="eng" />
       </gmd:language>
-      <!-- 
       <gmd:topicCategory>
```
fgravin, line 193: What is this file used to ? Because the template comes from the datadir doesn't it ?

fgravin (approved):
Tested and works well.
Thanks for the implementation

PR #3475, opened 2021-08-03
mapfishapp - results_with_summary is unsupported with GN4

fvanderbiest:
Untested for now - there's no garantee this will be enough for mapfishapp's layer search through CSW to work as expected.
What I can say is that it's a required move.


fvanderbiest:
Superseded by #3477

PR #3476, opened 2021-08-09
[WIP] Upgrade to GeoNetwork 4.0.5

groldan:
Branch to upgrade to GeoNetwork 4.0.5


landryb (commented):
what's the intent of this header ? time of last user modification by an admin in the console ? why a hash then ?

just wondering, because (at least with openldap) there are operational attributes related to creation/modification time and authorship.. as they're operational, they're not returned by default but only if you explicitely ask for them.

```
creatorsName: cn=admin,dc=craig,dc=fr
createTimestamp: 20181004190250Z
modifiersName: cn=admin,dc=craig,dc=fr
modifyTimestamp: 20181022132807Z
```

PR #3483, opened 2021-08-17
sp - implements webdav specific http verbs

pmauduit:
This is required in order to be able to proxify webdav requests, e.g. to interact with a NextCloud instance behind the SP.

### DIFF
--- a/security-proxy/src/main/java/org/georchestra/security/Proxy.java
+++ b/security-proxy/src/main/java/org/georchestra/security/Proxy.java
@@ -110,6 +110,8 @@ import org.georchestra.ogcservstatistics.log4j.OGCServiceMessageFormatter;
 import org.georchestra.ogcservstatistics.log4j.OGCServicesAppender;
 import org.georchestra.security.permissions.Permissions;
 import org.georchestra.security.permissions.UriMatcher;
+import org.georchestra.security.webdav.PropFindMethod;
+import org.georchestra.security.webdav.SearchMethod;
 import org.springframework.beans.factory.BeanInitializationException;
 import org.springframework.beans.factory.annotation.Autowired;
 import org.springframework.beans.factory.annotation.Qualifier;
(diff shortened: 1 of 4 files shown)


security-proxy/src/main/java/org/georchestra/security/Proxy.java
```diff
@@ -957,7 +958,25 @@ private HttpRequestBase makeRequest(HttpServletRequest request, String sURL) thr
...
+                pfm.setEntity(entity);
+                return pfm;
+            } else if (SearchMethod.METHOD_NAME.equalsIgnoreCase(method)) {
+                SearchMethod sm = new SearchMethod(uri);
+                int contentLength = request.getContentLength();
+                ServletInputStream inputStream = request.getInputStream();
+                HttpEntity entity = new InputStreamEntity(inputStream, contentLength);
+                sm.setEntity(entity);
+                return sm;
+            }
```
RemiDesgrange, line 977: I'm wondering in which case the two conditions couldn't be fullfiled?

pmauduit: if the coming request is a completely custom HTTP verb ? But then I'd expect an exception to be thrown here:
https://github.com/georchestra/georchestra/pull/3483/files#diff-557c2eb148c461ecfcde83d38082b5270605be793cd92a707660c2f38e527adfR1053

pmauduit: do you think it deserves an extra test ?

pmauduit: trying an extra test shows that `switch(null)` throws an npe.

pmauduit: I added a test & a guard to avoid the NPE.

pmauduit:
superseded by #3528

RemiDesgrange:
closed since #3569 superseed this.

PR #3488, opened 2021-08-24
Makefile: also pass -DskipTests to maven in war-build-georchestra

landryb:
otherwise docker-compose is required when using the Makefile..

cf https://packages.georchestra.org/bot/#/builders/3/builds/3/steps/1/logs/stdio, datafeeder only skips tests when `skipTests` is set, `maven.test.skip=true` isnt enough.


pmauduit (approved):
I'm ok with the change, but I find curious that the unit tests require docker-compose, @groldan any idea ?

PR #3471, opened 2021-07-19
deb - improve version calculation

pmauduit:
* using git-commit-id plugin to get the closest tag from the current commit
* injecting the value as a system property
* Accessing it while overriding the packageVersion to set it according to the previously obtained version.

runtime-tested on extractorapp, this produced a package named after the following file:

```
[WARNING] dpkg-deb: building package 'georchestra-extractorapp' in '/home/pmauduit/projects/georchestra/georchestra/extractorapp/target/georchestra-extractorapp_20.1.3.202107191450~903ced7-1_all.deb'.
```

PR #3490, opened 2021-08-25
Update release procedure

landryb:
Cf https://github.com/georchestra/georchestra/pull/3471#issuecomment-904786963

while here add the pom.xml bit adding git tags properties so that it ends in master branch for the next branching.

### DIFF
--- a/RELEASE_PROCEDURE.md
+++ b/RELEASE_PROCEDURE.md
@@ -103,6 +103,7 @@ Other tasks:
  * Manually update the files mentionning the current release version (```README.md``` and ```RELEASE_NOTES.md```)
  * Update the branch name for the Travis status logo
  * Change the `packageDatadirScmVersion` parameter in the root `pom.xml` to `20.1`
+ * Replace `99.master` by `${build.closest.tag.name}` in the root `pom.xml` so that debian packages have the right version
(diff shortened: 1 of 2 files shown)


RELEASE_PROCEDURE.md
```diff
@@ -103,6 +103,7 @@ Other tasks:
  * Manually update the files mentionning the current release version (```README.md``` and ```RELEASE_NOTES.md```)
  * Update the branch name for the Travis status logo
  * Change the `packageDatadirScmVersion` parameter in the root `pom.xml` to `20.1`
+ * Replace `99.master` by `${build.closest.tag.name}` in the root `pom.xml` so that debian packages have the right version
```
fvanderbiest, line 106: IIUC, this line should be removed after first execution.

landryb: i think it stays forever in the stable branches, we're talking about https://github.com/georchestra/georchestra/pull/3471/files#diff-9c5fb3d1b7e3b0f54bc5c4182965c4fe1f9023d449017cece3005d3f90e8e4d8R633 right ?

from my understanding only master should 'hardcode' `99.master` and the other branches should have `${build.closest.tag.name}`

PR #3496, opened 2021-09-08
ldap - introducing unique object identifier

pmauduit:
OpenLDAP provides a UUID as extra attribute, but this is vendor specific and unstable in case of dumping the LDAP using ldapsearch (slapcat will do, though).

Since we have a hand on the georchestra objectClasses already introduced, the idea is to introduce a (optional) ldap attribute which follows the UUID syntax on both georchestra classes (georchestraUser & georchestraOrg).

Tests: tested in the docker image. For existing LDAP trees, the LDAP schema for geOrchestra will have to be updated, but it should be possible to produce a LDIF modification spec.

### DIFF
--- a/ldap/docker-root/georchestra.ldif
+++ b/ldap/docker-root/georchestra.ldif
@@ -27,6 +27,7 @@ description: USER
 userPassword:: e1NIQX1SY1Z4b1ZiZHp2UVRVYWNUdk4zdVc2ZnBWR0E9
 knowledgeInformation: Internal CRM notes on testuser
 cn: testuser
+georchestraObjectIdentifier: 048b2f38-6ee7-4eec-9bed-349cc6eb13c3
 
 # testreviewer, users, georchestra.org
 dn: uid=testreviewer,ou=users,dc=georchestra,dc=org
@@ -43,6 +44,7 @@ description: Reviewer
 userPassword:: e1NIQX1Nb3IzdXZ5cnpISWpHK0crSEcvblhxZW8reWc9
 mail: psc+testreviewer@georchestra.org
 cn: testreviewer
+georchestraObjectIdentifier: 5f6ce744-4f36-4432-9543-7f6025fab02c
 
 # testeditor, users, georchestra.org
 dn: uid=testeditor,ou=users,dc=georchestra,dc=org
@@ -59,6 +61,7 @@ description: editor
 userPassword:: e1NIQX1mVTFvSmdzV0FEZ1ZtTHBHeHBxdFBVa2RiekU9
 mail: psc+testeditor@georchestra.org
 cn: testeditor
+georchestraObjectIdentifier: e75b336d-a9fc-4671-8002-e1a49c2acf2c
 
 # testadmin, users, georchestra.org
 dn: uid=testadmin,ou=users,dc=georchestra,dc=org
@@ -80,6 +83,7 @@ userPassword:: e1NIQX1kREU1SkEvMkVpVTRGMFFObUt5eXpuazUrN1E9
 mail: psc+testadmin@georchestra.org
 knowledgeInformation: Internal CRM notes on testadmin
 cn: testadmin
+georchestraObjectIdentifier: 0c6bb556-4ee8-46f2-892d-6116e262b489
 
 # geoserver_privileged_user, users, georchestra.org
 dn: uid=geoserver_privileged_user,ou=users,dc=georchestra,dc=org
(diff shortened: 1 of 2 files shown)


fvanderbiest:
Since we're at it, why not having it on roles too ? Would make sense IMO.

fvanderbiest:
To be documented on https://github.com/georchestra/georchestra.github.io/blob/master/oid.md too

pmauduit:
> Since we're at it, why not having it on roles too ? Would make sense IMO.

I don't think the roles will move as often as the organizations and users can (who will rename ROLE_ADMINISTRATOR to ROLE_BLAH ?). Also, roles are not `georchestraSomething` object class, so it will require to create a georchestraRole object class with the attribute onto it. It is feasible, but I wonder if this is necessary or wanted. @groldan any opinion ?

fvanderbiest:
> It is feasible, but I wonder if this is necessary or wanted

You know, there are geOrchestra instances which sync LDAP orgs with GN groups and others which sync LDAP roles with GN groups. So we need a symetrical way to address LDAP object renaming IMO.

groldan:
> > It is feasible, but I wonder if this is necessary or wanted
> 
> You know, there are geOrchestra instances which sync LDAP orgs with GN groups and others which sync LDAP roles with GN groups. So we need a symetrical way to address LDAP object renaming IMO.

.... no, I didn't know... :/

pmauduit:
> You know, there are geOrchestra instances which sync LDAP orgs with GN groups and others which sync LDAP roles with GN groups. So we need a symetrical way to address LDAP object renaming IMO.

Then we probably need to have a `georchestraRole` objectClass as well in the openldap, and having the roles attached to it.

groldan:
> > You know, there are geOrchestra instances which sync LDAP orgs with GN groups and others which sync LDAP roles with GN groups. So we need a symetrical way to address LDAP object renaming IMO.

I'm not sure I follow though.
Let's say we have the following user:
```json
{
  "id": "0c6bb556-4ee8-46f2-892d-6116e262b489",
  "username": "testadmin",
  "lastUpdated": "1a76a7ad7f472be2e43a0fe34a3048ce7b0dc7bb9ab8ff3b9256f092020ca424",
  "roles": [
    "ADMINISTRATOR",
    "SUPERUSER",
    "GN_ADMIN"
  ],
  "organization": {
    "id": "bddf474d-125d-4b18-92bd-bd8ebb6699a9",
    "shortName": "PSC",
    "name": "Project Steering Committee",
    "linkage": "https://www.georchestra.org/",
    "lastUpdated": "7751e4cc6118eeb98a578bdda4319435ee99343db4683d87b612294fc28a0cb0"
  },
  "firstName": "Test",
  "lastName": "ADMIN",
  "email": "psc+testadmin@georchestra.org"
}
```

So, the common case is to sync the `PSC` organization to a `PSC` group and have the GN user belong to that GN group.
But it could also be that the GN user should be attached to GN groups   `ADMINISTRATOR`, `SUPERUSER`, and  `GN_ADMIN` instead?

pmauduit:
> But it could also be that the GN user should be attached to GN groups `ADMINISTRATOR`, `SUPERUSER`, and `GN_ADMIN` instead?

At the very begining of geOrchestra, we did not have `Roles` and `Organizations` as a separated notion, and we had only a `ou=groups` branch in the LDAP. The trick at that time was to use a prefixes (`EL_...` for custom / instance-based groups like `EL_DEPARTEMENT_68`, and `SV_*` for "service levels" groups (which will become the roles later on) to distinguish roles from "organizations".

Some years ago we split to 2 different branches in the LDAP, but some users in the community (at least INRAE, but I think other ones are also willing to stick with the roles) kept the usage of the roles and this `EL_` convention (you can also create custom ones from the console), and we document how to synchronize "the former way" here: https://github.com/georchestra/georchestra/blob/master/docs/geonetwork-synchronization.md#synchronization-between-geonetwork-and-the-ldap

fvanderbiest:
Those who stick to the older behavior have this specific requirement where people should belong to several geonetwork groups (mapped to roles). This is not possible when geonetwork groups are synced with LDAP orgs, since users can belong to a single Org.

groldan:
So the answer to my question would be "yes", right?
Like in, that is consistent with `default.properties:ldapRolesRdn=ou=roles`
So we just need a way to tell the new sync process to use the roles instead of the org to sync GN groups.
Something like:
`geonetwork.properties`:
```
groups.sync-mode=roles|orgs
groups.sync-filter=EL_(.*)
```

pmauduit:
yes, I can't remember if they still use this EL_ prefix, but the way you suggest should be sufficiently configurable anyway

fvanderbiest:
> yes

This indeed looks good Gabriel.

> I can't remember if they still use this EL_ prefix, but the way you suggest should be sufficiently configurable anyway

Some of them even use the `(&(objectClass=groupOfMembers)(|(cn=EL_*)(cn=SP_*)))` pattern for group sync !

PR #3497, opened 2021-09-13
Georchestra gn4 4.x dev

groldan:


security-proxy-spring-integration/src/main/java/org/georchestra/security/client/console/GeorchestraConsoleOrganizationsApiImpl.java
```diff
@@ -0,0 +1,64 @@
...
+        Organization organization = client.get("/internal/organizations/id/{id}", Organization.class, id);
+        return Optional.ofNullable(organization);
+    }
+
+    @Override
+    public Optional<Organization> findByShortName(String shortName) {
+        Organization organization = client.get("/internal/organizations/name/{name}", Organization.class, shortName);
+        return Optional.ofNullable(organization);
+    }
+
```
pmauduit, line 63: I did not get yet how these endpoints will be secured, compared to the "public ones"

security-proxy/src/main/java/org/georchestra/security/UserDetailsJSONRequestHeaderProvider.java
```diff
@@ -0,0 +1,264 @@
...
+     */
+    private Map<String, Boolean> enabledServices = Collections.singletonMap(GLOBAL_KEY, Boolean.FALSE);
+
+    /**
+     * Cache time-to-live in milliseconds, can be given through an external
+     * configuration property {@code security-proxy.ldap.cache.ttl}
+     */
+    private static final long DEFAULT_CACHE_TTL = 2000;
+
+    public UserDetailsJSONRequestHeaderProvider() {
```
pmauduit, line 133: Checking the maximum size for http headers (not clear if it is for a single one or for multiple ones), it depends across webserver implementations:
https://www.tutorialspoint.com/What-is-the-maximum-size-of-HTTP-header-values
But we are quite limited (min 8KB), so we won't be able to transmit the jpegPhoto attribute using this one ;-)

commons/pom.xml
```diff
@@ -38,6 +46,19 @@
...
+        <executions>
+          <execution>
+            <id>attach-sources</id>
+            <phase>verify</phase>
+            <goals>
+              <goal>jar-no-fork</goal>
+            </goals>
+          </execution>
+        </executions>
+      </plugin>
```
pmauduit, line 61: :+1: for generating the sources jar (even if the idea decompiler plugin is so nice than I generally don't have the need for them, or the "get the sources from the Internet" feature from idea works pretty well also).

Any motivation to add them from your side ?

console/src/main/java/org/georchestra/console/dto/EventType.java
```diff
@@ -2,6 +2,8 @@
 
 /**
  * Type discriminator for application events
+ * <p>
+ * TODO: dead code!
```
pmauduit, line 6: grepping the code, it does not seem to be used somewhere indeed, I think we can remove the class.

pmauduit (commented):
Ok finished to read it ; it looks promising from my side and I think there are lots of improvements in here, so here are my summarized thoughts so far:

1. How the "internal API" to grab the users / roles / ... from the console is protected ? I did not see any change in the SP's security-mappings.xml nor I remember having seen some spring annotation to protect a regular user to consume them
2. I really like the idea of splitting the code from the DF to 2 new modules (security-proxy-spring-integration,ldap-account-management) + georchestra-commons for sharing purposes across components, but does this integrate well with spring-boot apps (DF) as well as the regular spring ones ? (see one of my comments about the spring version, I fear that it's not harmonized across geOrchestra modules).
3. As we are at LDAP schema modifications, shall we also create a georchestraRole along with an UUID ?
4. I also like the idea of a base64-encoded JSON document for the sec-user, but skimming the net about it, it seems that we are limited and I was worrying to reach the 8KB limit of some webserver implementations.

security-proxy-spring-integration/README.md
```diff
@@ -0,0 +1,293 @@
+# Data Feeder
```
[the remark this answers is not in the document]
groldan: done.

postgresql/130-geonetwor_georchestra.sql
```diff
@@ -0,0 +1,7 @@
+-- Create schema for georchestra-geonetwork user links
```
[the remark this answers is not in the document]
groldan: good catch. file's gone anyways

security-proxy-spring-integration/src/main/java/org/georchestra/security/client/console/GeorchestraConsoleOrganizationsApiImpl.java
groldan: Secured in proxy's `security-mappings.xml` (`docker-gn4` branch)

commons/pom.xml
groldan: well, eclipse doesn't seem to be that smart

ldap-account-management/pom.xml
```diff
@@ -0,0 +1,127 @@
...
+      <artifactId>spring-security-core</artifactId>
+    </dependency>
+    <dependency>
+      <groupId>com.fasterxml.jackson.core</groupId>
+      <artifactId>jackson-annotations</artifactId>
+    </dependency>
+    <dependency>
+      <groupId>com.fasterxml.jackson.core</groupId>
+      <artifactId>jackson-databind</artifactId>
+    </dependency>
```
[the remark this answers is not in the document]
groldan: console and security-proxy do use the same spring version afaik. And nobody else should use that module (accessing ldap)

ldap/docker-root/georchestraSchema.ldif
```diff
@@ -6,16 +6,21 @@ olcObjectClasses: ( 1.3.6.1.4.1.53611.1.1.1
...
+    DESC 'A UUID identifying the geOrchestra object'
+    SYNTAX 1.3.6.1.1.16.1
+    SINGLE-VALUE  )
 olcObjectClasses: ( 1.3.6.1.4.1.53611.1.1.2
     NAME 'georchestraOrg'
     DESC 'georchestra org'
     SUP top
     AUXILIARY
-    MAY (jpegphoto $ labeledURI $ knowledgeInformation))
+    MAY (jpegphoto $ labeledURI $ knowledgeInformation $ georchestraObjectIdentifier))
```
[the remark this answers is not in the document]
groldan: I think it's ok now? https://github.com/georchestra/georchestra/blob/georchestra-gn4-4.x-dev/ldap/docker-root/georchestraSchema.ldif

security-proxy/src/main/java/org/georchestra/security/UserDetailsJSONRequestHeaderProvider.java
groldan: yeah, but a user won't come near close to that, especially now that the roles and organization are just strings, not the full objects. The jpegPhoto is not included in the `GeorchestraUser` class either, and IMHO it shouldn't. If it comes to requiring it, we should add a method to `UsersApi` to fetch a users picture instead.

groldan:
also note this will work with  [docker/gn4](https://github.com/georchestra/docker/tree/gn4) branch, which's been rebased on top of master already (and so docker/config)

PR #3528, opened 2021-11-10
Supplement to web dav pr #3483

RemiDesgrange:
In order to nextcloud to fully work with georchestra, I added a bunch of other webdav verbs.

### DIFF
--- a/security-proxy/src/main/java/org/georchestra/security/Proxy.java
+++ b/security-proxy/src/main/java/org/georchestra/security/Proxy.java
@@ -24,7 +24,6 @@ import static org.georchestra.commons.security.SecurityHeaders.SEC_ROLES;
 import static org.springframework.web.bind.annotation.RequestMethod.GET;
 import static org.springframework.web.bind.annotation.RequestMethod.POST;
 
-
 import java.io.File;
 import java.io.FileInputStream;
 import java.io.IOException;
--- a/security-proxy/src/main/java/org/georchestra/security/Proxy.java
+++ b/security-proxy/src/main/java/org/georchestra/security/Proxy.java
@@ -956,7 +956,7 @@ public class Proxy {
 
             // handles webdav specific verbs
             String[] webdavVerb = { "COPY", "LOCK", "UNLOCK", "MKCOL", "MOVE", "PROPFIND", "PROPPATCH", "UNLOCK",
-                    "REPORT" };
+                    "REPORT", "SEARCH" };
             boolean isWebdav = Arrays.stream(webdavVerb).anyMatch(x -> x.equalsIgnoreCase(method));
             if (isWebdav) {
                 HttpEntityEnclosingRequestBase heerb = new HttpEntityEnclosingRequestBase() {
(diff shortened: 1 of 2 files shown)


pmauduit (approved):

security-proxy/src/main/java/org/georchestra/security/Proxy.java
```diff
@@ -956,7 +955,7 @@ private HttpRequestBase makeRequest(HttpServletRequest request, String sURL) thr
 
             // handles webdav specific verbs
             String[] webdavVerb = { "COPY", "LOCK", "UNLOCK", "MKCOL", "MOVE", "PROPFIND", "PROPPATCH", "UNLOCK",
-                    "REPORT" };
+                    "REPORT", "SEARCH" };
```
pmauduit, line 958: As discussed during the conf-call, we are not sure if these are all entity/body based (as `POST` is) ; for sure the ones from the former PR were (`PROPFIND` & `SEARCH`).

security-proxy/src/main/java/org/georchestra/security/Proxy.java
```diff
@@ -973,22 +972,6 @@ public String getMethod() {
...
-//                return pfm;
-//            } else if (SearchMethod.METHOD_NAME.equalsIgnoreCase(method)) {
-//                SearchMethod sm = new SearchMethod(uri);
-//                int contentLength = request.getContentLength();
-//                ServletInputStream inputStream = request.getInputStream();
-//                HttpEntity entity = new InputStreamEntity(inputStream, contentLength);
-//                sm.setEntity(entity);
-//                return sm;
-//            }
-
```
pmauduit, line 991: can be removed

security-proxy/src/main/java/org/georchestra/security/Proxy.java
```diff
@@ -956,7 +955,7 @@ private HttpRequestBase makeRequest(HttpServletRequest request, String sURL) thr
 
             // handles webdav specific verbs
             String[] webdavVerb = { "COPY", "LOCK", "UNLOCK", "MKCOL", "MOVE", "PROPFIND", "PROPPATCH", "UNLOCK",
```
pmauduit, line 957: * COPY does not seem to take any http entity:
https://docs.microsoft.com/en-us/previous-versions/office/developer/exchange-server-2003/aa142816(v=exchg.65)
* LOCK does
* UNLOCK does
* MKCOL does not
* MOVE does not
* PROPPATCH does
* REPORT does (https://datatracker.ietf.org/doc/html/rfc3253#section-3.7.1)

pmauduit: UNLOCK is defined twice in the array

pmauduit: it might work with HttpEntityEnclosingRequestBase anyway, but I think a more relevant one (the same as the base for GET requests for instance) should be used instead

Issue #3515, opened 2021-10-29
[header] broken link to GN4 admin page


landryb:
when logged in, in the header there's a link to geonetwork admin page in https://github.com/georchestra/georchestra/blob/master/header/src/main/webapp/WEB-INF/jsp/index.jsp#L321

from testing here, GN4 doesnt support anymore the two-letter codes (eg `/geonetwork/srv/fr/admin.console` gives a 404) so the code should use the 3-letters codes (`/geonetwork/srv/fre/admin.console` works)

pmauduit:
Should be fixed by the mentioned PR

PR #3550, opened 2021-12-06
header - using 3 letters for GN admin UI (#3515)

pmauduit:
tests: runtime into an ansible deployment:

* locales FR => /fre/
* locales EN => /eng/


fvanderbiest:
LGTM.

We should also remove these lines nowadays:
```
<!-- GN2 or GN3 -->
<!--li class="active"><a href="/geonetwork/srv/<%= lang %>/admin"><fmt:message key="catalogue"/></a></li-->
```

landryb:
tested locally, works fine !

PR #3560, opened 2021-12-20
mapfishapp: Fix build

groldan:
- Force System property `org.geotools.referencing.forceXY=true` for the application itself
- Set `org.geotools.referencing.forceXY=true`  on tests that require it and force forking the jvm when surefire runs tests so prior runs don't interfer with each other re geotools initialization
- Update SLDClassifierTest online WFS test with working parameters
- Improvements to CI build workflow


fvanderbiest:
Excellent, thanks Gabe !

fvanderbiest:
backported to 20.1.x as well

PR #3563, opened 2021-12-31
remove CKAN-related roles

fvanderbiest:
CKAN is an optional module, which is rarely deployed.
At some point we were thinking it made sense to have these roles in the main repo, but it does not anymore.

### DIFF
--- a/ldap-account-management/src/test/resources/defaultRoles.json
+++ b/ldap-account-management/src/test/resources/defaultRoles.json
@@ -10,31 +10,6 @@
 		]
 	},
 	{
-		"id": "7637a7bc-e540-4a5a-b828-dc2530540017",
-		"name": "CKAN_ADMIN",
-		"description": "This role grants admin rights in CKAN, scoped to user's org",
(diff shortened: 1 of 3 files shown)


pmauduit (approved):

pmauduit:
This PR is actually needed to go forward with the GN4 integration

pmauduit:
rebased (hence the push-force)

pmauduit:
For the record, it turned out it was not that simple, the following resources had to be updated:
* https://github.com/georchestra/georchestra/blob/master/ldap-account-management/src/test/resources/defaultUsers.json
* https://github.com/georchestra/geonetwork/blob/georchestra-gn4-4.x/georchestra-integration/georchestra-authnz/src/test/resources/org/georchestra/geonetwork/security/integration/defaultUsers.json

Because the lastUpdated is actually a hash depending on the users properties. To update these files, one need to simply launch a docker composition (with at least console + ldap, maybe the db ?), then call the `console:8080/console/internal/users`. The json array obtained there can be indented by piping it into `jq`.

pmauduit:
I forgot to thanks @groldan for the hint

PR #3568, opened 2022-01-18
Adding migration notes for the LDAP / GN4

pmauduit:
Following the LDAP revamp, it is necessary to update existing LDAP trees prior to be able to run the current master (coming 21.x) georchestra version.

This aims to document how to "easily" migrate an existing LDAP tree.

Thanks @landryb for having commented my gist (https://gist.github.com/pmauduit/c743ac13cc4b08dd4cda150d6279ef16#gistcomment-3977590), which this PR has been mainly inspired from.

I would have liked to find a way to do it in one ldapmodify file spec on the schema, but I could not manage to have it, so it requires 3 different ldapmodify files / command, unfortunately.

### DIFF
--- a/migrations/21.0/README.md
+++ b/migrations/21.0/README.md
@@ -9,3 +9,157 @@ For existing _Geonetwork 4_ installations, it can be done by unticking the check
 ```
 ☐ Authentication
 ```
+
+### authtype
+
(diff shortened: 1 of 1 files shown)


fvanderbiest:
:+1:

Issue #3175, opened 2020-10-12
GeoServer login from header


fvanderbiest:
When clicking the login link on geoserver home page, the generated link is https://xxx/geoserver/web/?0&login (which is logical since the page url is https://xxx/geoserver/web/?0)

The problem is, after successful login, one gets redirected to https://xxx/geoserver/web/?0= where the login is not effective.
One has to click any geoserver link to get the geoserver admin page.

PR #3577, opened 2022-01-26
header - fix login from geoserver web

fvanderbiest:
Proposal for #3175


fvanderbiest:
Also changing logging in from any cas page, to redirect to the defined homepage

PR #3578, opened 2022-01-26
Remove the uid query string from change password form

### LINKED ISSUE #3138: console - the "change password" page should not require a UID parameter
Currently the "change password" page is accessible at `/console/account/changePassword?uid=fvanderbiest`.
Without the `uid` parameter, one gets "Required String parameter 'uid' is not present"
Instead, `uid` should be obtained from the security-proxy-provided `sec-username` header.

This paves the way for https://web.dev/change-password-url/ ...

RemiDesgrange:
Fix #3138


fvanderbiest:
LGTM

Issue #2489, opened 2019-02-11
Update jetty images to JRE 11


fvanderbiest:
There are jetty images running with jre11 see https://hub.docker.com/_/jetty

fvanderbiest:
Will require GeoServer 2.15 at least

fvanderbiest:
Gabriel commented elsewhere "Also make use of the (new in Java 10) -XX:InitialRAMPercentage, -XX:MaxRAMPercentage, and -XX:MinRAMPercentage JVM options to control the heap size relative to the container memory limits"

RemiDesgrange:
Will also allow use to test different GC option. Java 11 has 3 GC (ParrallelGc, G1, ZGC).

landryb:
Fwiw java 11 is the only version supported in debian 10/buster.

RemiDesgrange:
I will update the doc accordingly but I am still clarifiying stuff here:

**java 11 means nothing when it comes to runtime**. 

Java11 is a language specification. It runs on JVM. Since the new release process of Oracle (with JDK9) everything change. We can't talk about java11 when we talk about the runtime, the only runtime we're interest on is OpenJDK. OpenJDK have multiple builds. For example github actions (CI/CD) uses Azul build of OpenJDK. I don't know which build Debian uses. 
Georchestra _should_ run fine on any OpenJDK builds, but since we only test it on Azul+Debian+AdoptOpenJDK one we cannot know in advance.

Jetty docker images are using AdoptOpenJDK: https://github.com/docker-library/openjdk/blob/master/11/jre/Dockerfile#L36

And for the record, relying on debian package for JRE is a ticking bomb...

landryb:
> And for the record, relying on debian package for JRE is a ticking bomb...

can you clarify ? because that's what everyone (not using docker) is doing.

RemiDesgrange:
Two things:

* It happened in the past that Debian broke OpenJDK, without any notice. It took me out of my bed 1 time a 3am. Not doing this mistake again. 
* Debian does not _always_ follow the exact release version. People start noticing when JDK8u212 was released (this update bring cgroup support in the OpenJDK JVM) and some bugs remain on the debian version, because debian maintainer didn't use the same commit (tag)  the release uses. (cannot found the blog post from docker that talks about it).

Still Debian remain a distribution heavily tested and with a huge user base. I'm really not a fan of the release management at debian but here I stop claiming fact and start putting some personal opinion :-)

[...]

PR #3581, opened 2022-01-26
Jdk 11 compat

RemiDesgrange:
Supersede #3183


pmauduit:
Current status with `mvn clean install -fae`:
```
[INFO] Mapfishapp Webapp 21.0-SNAPSHOT .................... FAILURE [ 32.495 s]
[INFO] Analytics webapp 21.0-SNAPSHOT ..................... FAILURE [  6.690 s]
[INFO] Data-Feeder microservice 21.0-SNAPSHOT ............. FAILURE [01:25 min]
[INFO] Data-Feeder (web ui) 21.0-SNAPSHOT ................. FAILURE [ 13.983 s]
```

I'm still wondering if we go this way, if geonetwork also would need to migrate to jdk11. I prepared a PR about it here:
https://github.com/georchestra/geonetwork/pull/191/files
but it will require to disable the failing test (workaround for date/time parsing/calculation on one harverster), and be suggested upstream as well.

pmauduit:
testsuite for mfapp is broken in java11 because of xerces / xercesImpl, something resembling as described here: https://issues.apache.org/jira/browse/XERCESJ-1706
The dependency is transitively added by mfprint (org.mapfish.print:print-lib). We might need to add a custom profile for jdk11 redefining the dependency to the print with an extra exclude:

```
[INFO] +- org.mapfish.print:print-lib:jar:2.1.1-georchestra:compile
[...]
[INFO] |  +- xerces:xercesImpl:jar:2.4.0:compile
```

pmauduit:
It will also require to *check every dockerfiles* and make sure they are using a jetty based image with a sufficiently recent jre before merging.

RemiDesgrange:
Fix #2489

pmauduit (approved):

PR #3594, opened 2022-01-27
Remove cas-server-webapp

RemiDesgrange:
Remove cas-server-webapp code from the repo. We now need to use the CAS version 6 in [this specific repo](https://github.com/georchestra/georchestra-cas-server).


landryb:
don't forget to update the Makefile at the root ?

pmauduit (approved):

PR #3608, opened 2022-02-03
[header] add link to ms2 home on /mapstore/#/home

landryb:
fixes georchestra/mapstore2-georchestra#459

to be discussed wether this should replace the direct link to the viewer or next to it, for the default header i think it makes sense to have both for discoverability and ppl are free to comment one or the other. @fvanderbiest ?

while here i noticed the `active` mechanism is partly broken, doesnt seem to work all the time, clicking on 'services' properly yields to geoserver but the `services` div doesnt end up with the `active` class, same thing for mapstore home.

### DIFF
--- a/header/src/main/webapp/WEB-INF/jsp/index.jsp
+++ b/header/src/main/webapp/WEB-INF/jsp/index.jsp
@@ -289,7 +289,7 @@ if(sec_roles != null) {
             <li class="active"><a><fmt:message key="mapstore.home"/></a></li>
             </c:when>
             <c:otherwise>
-            <li><a href="/mapstore/#/home"><fmt:message key="mapstore.home"/></a></li>
+            <li><a href="/mapstore/#home"><fmt:message key="mapstore.home"/></a></li>
             </c:otherwise>
         </c:choose>
 
--- a/header/src/main/webapp/WEB-INF/jsp/index.jsp
+++ b/header/src/main/webapp/WEB-INF/jsp/index.jsp
@@ -285,6 +285,15 @@ if(sec_roles != null) {
         </c:choose>
 
         <c:choose>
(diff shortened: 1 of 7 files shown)


landryb:
> In order to have the "active" CSS tab working, it is the responsability of the webapp which integrates the header to set it according to what the header expects, for mapstore, it seems that the expected value is "mapstore" (by default, or whatever is passed to the function): https://github.com/georchestra/mapstore2-georchestra/blob/master/js/plugins/Header.jsx#L23

right, i think i see, thanks for the cluebat :)

header/src/main/webapp/WEB-INF/jsp/index.jsp
```diff
@@ -284,6 +284,15 @@ if(sec_roles != null) {
             </c:otherwise>
         </c:choose>
 
+        <c:choose>
+            <c:when test='<%= active.equals("mapstore_home") %>'>
+            <li class="active"><a><fmt:message key="mapstore.home"/></a></li>
+            </c:when>
+            <c:otherwise>
+            <li><a href="/mapstore/#home"><fmt:message key="mapstore.home"/></a></li>
```
fvanderbiest, line 292: seems like the correct link is /mapstore/#/home (though a redirect is made from /mapstore/#home)

fvanderbiest:
I agree on having by default a link to mapstore home from the header BUT i'd like to find a better name than "mapstore home"... which evokes nothing to a normal user.

It's difficult to summarize in a word all one can find on mapstore home ... maps, stories, dashboards ... how can we name this conveniently ?

fvanderbiest:
I know : "Maps" ;-)

fvanderbiest:
@landryb I think we want this for 22.0 ?
Do you want me to take over or will you push more commits to it ?

PR #3610, opened 2022-02-04
lower loglevel to debug, no need to spam logs

landryb:
noticed while looking at my loglevels, otherwise triggers a line at each request if the logger is at INFO lvl.


pmauduit (approved):

pmauduit:
I don't know why the df / gn do not pass, but given the modification, I allow myself to merge.

Issue #3574, opened 2022-01-24
LDAP backend


fvanderbiest:
Reviewing what we have in the repository, it seems we are using the older `hdb` backend for our docker images, while the newer `mdb` is used when installing through debian packages.
see eg https://github.com/georchestra/georchestra/blob/930ab7c30525c85b26de0576a1cfb97454dd174b/ldap/docker-root/docker-entrypoint.d/01-populate#L36

Why not using `mdb` everywhere ? cc @Vampouille

fvanderbiest:
Yep, this is what's recommended indeed.

pmauduit:
https://github.com/georchestra/georchestra/pull/3612

jmbourdaret:
> Reviewing what we have in the repository, it seems we are using the older `hdb` backend for our docker images, while the newer `mdb` is used when installing through debian packages. see eg
> 
> https://github.com/georchestra/georchestra/blob/930ab7c30525c85b26de0576a1cfb97454dd174b/ldap/docker-root/docker-entrypoint.d/01-populate#L36
> 
> Why not using `mdb` everywhere ? cc @Vampouille

see #1498 , looks like some time ago, we had two different LDIF for docker and non-docker installs.

pmauduit:
Funny, but it seems that with the new version (or because of switching to mdb), the cn=admin user still exists but is not queriable. We can still bind onto the LDAP using cn=admin,dc=georchestra,dc=org, but it won't appear in the result of a ldapsearch anymore.
This is why the tests are failing (because a preflight request is made and checks the existence of cn=admin in the tree).

pmauduit:
rebased / testsuite updated, hoping it will pass now.

PR #3612, opened 2022-02-04
ldap - migrating to mdb (#3574)

pmauduit:
This mainly affects the docker image, which was still making use of hdb
as backend for the openldap database.

tests: docker build & run:

* Ldapsearch reveals:
  * the georchestra schema is still here
  * the memberOf overlay still functions
* checking by hand into /etc/ldap/slapd.d proves that the georchestra db
  is now making use of mdb.


pmauduit:
I also took the opportunity to migrate to debian bullseye.

pmauduit:
Tests on the console are failing, need to investigate why (esp. "no such object" querying the LDAP, it's likely to be related to the changes here).

pmauduit:
> Tests on the console are failing, need to investigate why (esp. "no such object" querying the LDAP, it's likely to be related to the changes here).

Fixed with https://github.com/georchestra/georchestra/issues/3574#issuecomment-1036296082

PR #3628, opened 2022-02-08
Make OrgsDaoImpl.findAllWithExt() org to ext matching case insensitive

### LINKED ISSUE #3626: ldap-account-mgmt - case-sensivity (java-side) vs case-insensivity (openldap)
Given the following ldap objects to describe an organization:
https://github.com/georchestra/georchestra/blob/master/ldap/docker-root/georchestra.ldif#L378-L398

OpenLDAP o / cn fields are case insensitive, which means we can ldapsearch on o=c2c or o=C2C which will return the same result.

The issue is that the ldap-account-management will try to figure out one object to the other using a lookup into a map:
https://github.com/georchestra/georchestra/blob/master/ldap-account-management/src/main/java/org/georchestra/ds/orgs/OrgsDaoImpl.java#L330 which will return the expected object only if the case matches.

groldan:
Fixes #3626 

OpenLdap seach (e.g. `o=c2c`) is case insensitive, meaning
org's id may be stored as `C2C` but its link to the `OrgExt`
as `seeAlso: o=c2c,ou=orgs,dc=georchestra,dc=org`.

`findAllWithExt()` hence needs to match Org to OrgExt ids
in a case insensitive way.

Single-object searches though will match without special
treatment because they'll use the case-sensitiveness matching
of the underlying LDAP server.

### DIFF
--- a/ldap-account-management/src/main/java/org/georchestra/ds/orgs/OrgsDaoImpl.java
+++ b/ldap-account-management/src/main/java/org/georchestra/ds/orgs/OrgsDaoImpl.java
@@ -58,8 +58,6 @@ import org.springframework.ldap.filter.Filter;
 import org.springframework.ldap.support.LdapNameBuilder;
 import org.springframework.util.StringUtils;
 
-import lombok.Setter;
-
 /**
  * This class manage organization membership
  */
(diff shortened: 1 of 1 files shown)


pmauduit:
What about also modifying the sample ldif file so that we can make sure this will work (like having `o=c2c` along with `cn=C2C`) ?

pmauduit (commented):
LGTM

PR #3642, opened 2022-02-15
gn - updating gn submodule

landryb:
- java11 compat (georchestra/geonetwork#191)
- ms2 geopublisher compat (georchestra/geonetwork#197)


pmauduit (approved):

pmauduit:
build fails because of the docker login action which requires the docker-hub username (which is not available from your fork)

PR #3663, opened 2022-03-02
extract user and pass from jdbcUrl in pool creation

jmbourdaret:
same as  #3662  for the `20.1` branch

_description_:
Sometimes, `pgsqlDatabase` contains special parameters string.
like `$pgsqlDatabase=georchestra?sslmode=require`

This can confuse the `jdbcUrl` connection string like this (for data pools): 
```
<property name="jdbcUrl" value="jdbc:postgresql://${pgsqlHost}:${pgsqlPort}/${pgsqlDatabase}?user=${pgsqlUser}&amp;password=${pgsqlPassword}"/>
=>
"jdbc:postgresql://10.0.0.42:5432/georchestra?sslmode=require?user=georchestra&password=nope"
```
(note the double character `?`)

There is some workaround by setting in *.properties* files :
```
pgsqlUser=georchestra&sslmode=require
```
it takes care of the double `?` but is not really readable and then confuse https://github.com/georchestra/georchestra/blob/48fb57feda89b38ff83926e96477757c3e052008/commons/src/main/java/org/georchestra/commons/WaitForDb.java#L25

So we will probably be better off splitting this jdbcUrl into small chunks...

### DIFF
--- a/analytics/src/main/webapp/WEB-INF/ws-servlet.xml
+++ b/analytics/src/main/webapp/WEB-INF/ws-servlet.xml
@@ -17,7 +17,10 @@
     </bean>
 
     <bean id="dataSource" class="com.mchange.v2.c3p0.ComboPooledDataSource" depends-on="waitForDb">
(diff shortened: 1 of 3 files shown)


jmbourdaret:
I think this PR and #3662 can be merged safely but I'm no expert.

pmauduit (approved):

Issue #3627, opened 2022-02-08
[datafeeder] hardcodes resourceType=series


landryb:
I dunno why this has been made this way, but metadata created via datafeeder hardcode the `series` resource type ('Collection de données'), while so far most of the catalogs we know use `dataset` (eg 'Jeu de données') cf for example https://public.sig.rennesmetropole.fr/geonetwork/srv/fre/catalog.search#/home or https://ids.craig.fr/geocat/srv/fre/catalog.search#/home

why this choice ?

i'm asking this, because by default mapstore2 CSW search applies a filter on `dataset` (cf https://mapstore.readthedocs.io/en/latest/user-guide/catalog/#static-filter-and-dynamic-filter) so not all records are shown from the catalog. Sure, the default CSW filter can be amended....

fvanderbiest:
Good catch, thanks !
We'll have a look at it. Makes no sense to me.

jahow:
The `series` value is hardcoded here FYI: https://github.com/georchestra/georchestra/blob/10c251d197e62244e467fb16b5926f0953a1ad6c/datafeeder/src/main/resources/default_iso_2005_gmd.xsl#L215-L218

fgravin:
Thanks @landryb for pointing this out.

It's weird that the hierachy level is overridden in the XSL transfo
https://github.com/georchestra/datadir/blob/master/datafeeder/metadata_transform.xsl#L223

while correctly set in the template https://github.com/georchestra/datadir/blob/master/datafeeder/metadata_template.xml#L22

I think it is just a mistake while taking a XSL file as inspiration for the transformation.
I let @groldan confirm that, thank you !

pmauduit:
See https://github.com/georchestra/datadir/pull/233 & https://github.com/georchestra/georchestra/pull/3671

pmauduit:
Merged both PR above.

PR #3671, opened 2022-03-10
datafeeder - being able to override the resource type (#3627)

pmauduit:
1. Taking back basically what has been done as PR onto the [datadir](https://github.com/georchestra/datadir/pull/233), but adding a test & modifying the default xsl bundled into the DF.
2. adds the possibility to set the default resource type into the properties file from the datadir.

tests: testsuite updated, runtime tested in a vagrant / ansible setup.


fvanderbiest:
:+1:

PR #3693, opened 2022-03-18
console Dockerfile: remove jetty embedded mail system

### LINKED ISSUE #3692: Broken password recovery after Java 11 upgrade
Handler dispatch failed; nested exception is java.lang.NoClassDefFoundError: javax/activation/DataSource on console/account/passwordRecovery

jmbourdaret:
fixes #3692 
Both "user creation" and "forgot password" features failed with :
```
NoClassDefFoundError: javax/activation/DataSource
```

it turns out *jetty9-jre11* has some issues with `javax.activation` package:
https://github.com/eclipse/jetty.docker/issues/10

This PR removes the jetty *embedded* mail libs, and therefore its `JETTY_HOME` restricted classpath.
we don't need those because we bring our own mail libs :
https://github.com/georchestra/georchestra/blob/05c3864a4ca7cdcb65ba0caca2341c79e3cbdaf2/pom.xml#L157-L159

### DIFF
--- a/console/src/docker/Dockerfile
+++ b/console/src/docker/Dockerfile
@@ -1,5 +1,12 @@
 FROM jetty:9-jre11
 
+# remove jetty embedded "JavaMail", we have our own.
+# fix #3692 : ClassNotFoundException: javax.activation.DataSource
+# https://github.com/eclipse/jetty.docker/issues/10
+USER root
+RUN rm -r /usr/local/jetty/lib/mail
+USER jetty
+
 ENV XMS=512M XMX=1G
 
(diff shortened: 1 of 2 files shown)


jmbourdaret:
Some checks are failing, but they did also fail on other pipelines. 
Are they relevant here ?

pmauduit:
> Are they relevant here ?

I guess we are better of fixing them, as it will be cumbersome in the future otherwise ... Reading the output, it should be just about using http instead of git protocol to grab some npm dependencies. I cannot remember the reason why we had to go through dep resolution via the git protocol anyway.

Issue #3548, opened 2021-12-01
Removes GN_USERADMIN role


fvanderbiest:
GN_USERADMIN was defined in #3399 

To be added to the default LDAP + migration notes + to console's "protected roles" (to prevent deletion)

landryb:
fwiw, since its not fully clear to me, the 'user admin' profile in GN has this rights (per https://geonetwork-opensource.org/manuals/3.10.x/en/administrator-guide/managing-users-and-groups/index.html#users-groups-and-roles):

> User Administrator Profile
> The User Administrator is the administrator of his/her own group with the following privileges:
> *  Full rights on creating new users within the own group
> *  Rights to change users profiles within the own group
> *  Full rights on creating/editing/ deleting new/old data within the own group


in the georchestra context, the first right doesnt make sense, so it boils down to 'the useradmin is able to grant editor/reviewer rights to users in his group and fully manage MD published in his group' ? Did i summarize that right ? will that mean that GN_USERADMIN will be able to grant GN_EDITOR and GN_REVIEWER in the ldap too, as to me the 'profile by group' info is local to geonetwork only ? This really need proper clear documentation :)

so it's a bit like 'delegated admin' for the console ?

fvanderbiest:
I have to admit that it does not seem to fit very well.
I'll ask again @gryckelynck about the original purpose.

fvanderbiest:
"L’objectif initial était de pouvoir déclencher le moissonnage d’un catalogue via l’API (visiblement ce n’est possible que par les admins de groupes dans GN)."

fvanderbiest:
Based on the above, I think we should not advertise this role in the console app. 
It should stay confidential, if not being removed.

pmauduit:
fixed testsuite after having merged PR above.

PR #3694, opened 2022-03-22
removing testuseradmin & GN_USERADMIN

fvanderbiest:
See https://github.com/georchestra/georchestra/issues/3548#issuecomment-1072224795 for the motivations


pmauduit:
LGTM but modifying this likely causes the test to fail (see https://github.com/georchestra/georchestra/issues/3574#issuecomment-1036296082)

pmauduit:
Merging for now, we will spy on the failing CI processes and repair afterwards (we'll probably need to have a look on GN).

PR #3698, opened 2022-03-23
correct mapstore label in Admin Header panel

jmbourdaret:
follow-up to #3695 
While logged in as an admin user, the _mapstore admin_ link was displayed as `???mapstore???` because it couldn't find a proper translation.


fvanderbiest:
Good catch !
Instead of "map" I would favor "viewer" though.

Thanks

PR #3709, opened 2022-04-06
df,gs - fixing debian package calculation

pmauduit:
See https://github.com/georchestra/georchestra/issues/3701#issuecomment-1081828335
This shall harmonize the calculation on the datafeeder + geoserver
package, as these modules are disconnected from the root pom. We
basically need to copy the missing maven configuration / plugin parts
defined in the root pom.

tests: locally tested a `mvn package deb:package -PdebianPackage`:

* df generated: `georchestra-datafeeder_22.0.0.202204061354~2851446-1_all.deb`
* gs generated: `georchestra-geoserver_22.0.0.202204061357~2851446-1_all.deb`

Note: we probably should port the same modification to master.

PR #3711, opened 2022-04-07
refix debian package version, matching on branch name instead of SNAPSHOT

landryb:
followup to 993e7c7e, 0715664 and #3709

with that, the logic is:
- on stable branches (eg matching \d{2}\.\d\.x$, so 20.1.x, 22.0.x...) start the package version with the closest tag name
- on other branches (master, pull requests, experimental branches..) use 99.master

for geonetwork and cas-server, the upstream version is used (eg 4.0.6-georchestra or 6.3.7.4)

another PR targetting 22.0.x with the same login is coming.

### DIFF
--- a/datafeeder/pom.xml
+++ b/datafeeder/pom.xml
@@ -632,9 +632,9 @@
                             <exportAntProperties>true</exportAntProperties>
                             <target>
                                 <condition property="project.packageVersion"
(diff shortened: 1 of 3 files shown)


pmauduit:
LGTM, sorry for having messed up the calculation in the first place.

pmauduit (approved):

PR #3722, opened 2022-04-13
gs - upgrade to 2.18.6

pmauduit:
Note: geofence testsuite is broken, I don't expect this to work at
runtime (at least with geofence enabled). It compiles with skipped
tests though.

Note2: we need to add a maven repository to geo-solutions, because this
is the only place where hibernate-spatial-postgis is available (in
version 1.1.3.2).

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -31,9 +31,9 @@ docker-build-geoserver: docker-pull-jetty
 
 docker-build-geoserver-geofence: docker-pull-jetty
 	cd geoserver; \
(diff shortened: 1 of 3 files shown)


groldan:
Should be fixed as of https://github.com/georchestra/geoserver/commit/c63493728c4428ad666a774826417b2e73645f61
Note I've force-pushed to this branch to remove the note in the commit about the geofence failure

pmauduit:
Thanks @groldan

landryb:
fwiw this doesnt work at runtime with tomcat here, cf #3729

PR #3723, opened 2022-04-15
GS - moving to GS 2.20.4

pmauduit:
compilation OK using the Makefile (incl. geofence), untested at runtime.

### DIFF
--- a/.github/workflows/geoserver.yml
+++ b/.github/workflows/geoserver.yml
@@ -33,11 +33,6 @@ jobs:
         distribution: 'adopt'
         cache: 'maven'
 
-    - name: "Configuring Maven"
-      run: |
-          mkdir -p $HOME/.m2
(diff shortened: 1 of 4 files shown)


pmauduit:
Weird, the build fails because it can't find the `26-SNAPSHOT` version of `gt-mbtiles`. I have it locally in my .m2 but since I don't usually compile geotools by hand, I have no idea where it comes from. I can't find it either on osgeo's maven repository.

pmauduit:
@groldan I reverted back to the upstream geofence / geofence server module in my 2.20.4-georchestra, so we probably miss https://github.com/geoserver/geoserver/pull/5109

groldan (approved):
as mentioned on slack, cherry-picked the geofence stuff and it works. I guess for the time being we can live with this, but it'd be nicer to split the changes into single-purpose, sensible commits for the sake of maintainability

PR #3761, opened 2022-06-20
datahub: update CI to use node 16.15

fgravin:
Force the usage of `npm` version `<= 8.5.5` otherwise the `npm install` fails cause of `peerDependencies`.


fvanderbiest:
Thanks !

PR #3766, opened 2022-06-29
Fix geoserver pom file to include sentry and logstash dependencies correctly

jemacchi:
Profiles defined for logstash and sentry at pom.xml root level 
```
<profile>
      <id>log4j-logstash</id>
      <dependencies>
        <dependency>
          <groupId>net.logstash.log4j</groupId>
          <artifactId>jsonevent-layout</artifactId>
          <version>1.7</version>
        </dependency>
      </dependencies>
    </profile>
...
    <profile>
      <id>sentry-log4j</id>
      <dependencies>
        <dependency>
          <groupId>io.sentry</groupId>
          <artifactId>sentry-log4j</artifactId>
          <version>1.7.30</version>
        </dependency>
      </dependencies>
    </profile>
```

are not really doing anything related to including dependencies on the Geoserver build process as in here _georchestra/.github/workflows/geoserver.yml_ :
```
run: ../../mvnw ... clean package docker:build -Pdocker,${{ env.GEOSERVER_EXTENSION_PROFILES }},geofence,log4j-logstash,sentry-log4j .... 
```
Adding the profiles in geoserver/webapp/pom.xml  fixed the build, including these jar files in the resulting WEB-INF/lib folder:

- /WEB-INF/lib/commons-lang-2.6.jar
- /WEB-INF/lib/jsonevent-layout-1.7.jar
- /WEB-INF/lib/json-smart-1.1.1.jar
- /WEB-INF/lib/sentry-1.7.30.jar
- /WEB-INF/lib/sentry-log4j-1.7.30.jar

### DIFF
--- a/geoserver/webapp/pom.xml
+++ b/geoserver/webapp/pom.xml
@@ -1474,5 +1474,28 @@
         </dependency>
       </dependencies>
     </profile>
+    
+    
+    <profile>
+      <id>log4j-logstash</id>
+      <dependencies>
+        <dependency>
+          <groupId>net.logstash.log4j</groupId>
+          <artifactId>jsonevent-layout</artifactId>
+          <version>1.7</version>
+        </dependency>
+      </dependencies>
+    </profile>
+    <profile>
+      <id>sentry-log4j</id>
+      <dependencies>
+        <dependency>
+          <groupId>io.sentry</groupId>
+          <artifactId>sentry-log4j</artifactId>
+          <version>1.7.30</version>
+        </dependency>
+      </dependencies>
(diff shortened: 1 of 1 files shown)


pmauduit (approved):

pmauduit:
I don't know why we lost these profiles, maybe because we detached GS from the root pom ? And we have to be cautious when we will upgrade GS to a new version to either remove the profile (at least for log4j2, because I think that we have a json output right into the log4j2 core, and no need to add an optional module anymore).
for sentry, there is something for log4j2 but it would be another dependency

PR #3772, opened 2022-07-13
df-frontend - harmonizing build

pmauduit:
For now, we are building the docker image for the datafeeder UI into a separate repository, and the generic war / debian package into the main (e.g.this) repository.

this PR aims to harmonize the way we build, by building the docker image at the same place.

Tests: `mvn clean package docker:build -Pdocker` into datafeeder-ui generated a docker image which is quite the same as georchestra/datafeeder-frontend:latest, having a quick look inside it.

If this PR is accepted, then the github action located into the geonetwork-ui fork (branch georchestra-datafeeder) won't be necessary anymore, and there won't be any need of setting a tag there either.

This should not change anything to the other builds (war,deb).


pmauduit:
This PR targets master, it makes more sense to set the milestone onto 22.1.0

Issue #3759, opened 2022-06-14
ldap upgrade notes for 22.0 should be fixed


landryb:
following https://github.com/georchestra/georchestra/tree/master/migrations/22.0#from-20x-to-22x the two ldap steps to add `georchestraRole` *then* replace `georchestraOrg` & `georchestraUser` result in a schema without `georchestraRole`.

ideally, both operations should be merged - eg replace the `olcObjectClasses` which previously had 2 classes without `georchestraObjectIdentifier` by 3 classes with a `georchestraObjectIdentifier` attribute.

landryb:
to be clearer, https://github.com/georchestra/georchestra/tree/master/migrations/22.0#adding-the-georchestrarole-objectclass-to-the-schema should be done after https://github.com/georchestra/georchestra/tree/master/migrations/22.0#adding-the-identifier-to-the-2-existing-object-classes, or merged in the same ldif..

fvanderbiest:
:+1:

pmauduit:
PR has been merged, closing

PR #3773, opened 2022-07-13
doc - fixing upgrade notes for 22.0 (#3759)

pmauduit:
See #3759 

It sounded interesting to keep both LDIF separated to me because the purpose of both of them is different (adding an objectClass definition vs adding an attribute on existing objectclasses).


pmauduit:
Will probably require to be ported to 22.0.x branch as it is mainly required here.

PR #3774, opened 2022-07-19
Add objectClass georchestraUser on ldap users in README

arnaud-morvan:
As there is no Python interpreter in LDAP image, so I did not added the Python script in image, but use a venv to run it locally.


fvanderbiest:
Very elegant. :+1:

pmauduit (commented):
:+1:

Issue #3779, opened 2022-08-01
[geonetwork] ldap<->gn4 sync broken with large directories/many users/orgs using default sync settings


landryb:
trying to upgrade our instance to 22.0, which includes gn4 and the new "integration" mechanism with ldap by @groldan - sadly, it doesn't work at all with large directories (1k orgs, 1.5k users), i guess because it needs some paging mechanism or larger timeouts or something.

once all the db migrations more or less managed to pass, GN4 tries to sync the orgs & users from the console, which apparently takes too much time:
```
2022-07-29 15:51:08,619 INFO  [org.geonetwork.security.external.integration] - No organizations were deleted.
2022-07-29 15:51:08,623 INFO  [org.geonetwork.security.external.integration] - Reconciling existing GN group AFB with canonical group (id: 6a9655b1-94c2-464a-b1c4-d15c6d67b8c2)
...
2022-07-29 15:51:15,651 INFO  [org.geonetwork.security.external.integration] - Reconciling existing GN group COM_ST_JULIEN_MONTDENIS with canonical group (id: 0d50fc31-d6cf-4cdf-9db2-ba55180393df)
2022-07-29 15:51:21,739 WARN  [org.geonetwork.security.external.integration] - Synchronization failed, retrying in 10 SECONDS.
org.springframework.web.client.ResourceAccessException: I/O error on GET request for "http://localhost:9480/console/internal/users": Read timed out; nested exception is java.net.SocketTimeoutException: Read timed out
        at org.springframework.web.client.RestTemplate.doExecute(RestTemplate.java:748)
        at org.springframework.web.client.RestTemplate.execute(RestTemplate.java:674)
        at org.springframework.web.client.RestTemplate.getForEntity(RestTemplate.java:342)
        at org.georchestra.security.client.console.RestClient.get(RestClient.java:40)
        at org.georchestra.security.client.console.RestClient.getAll(RestClient.java:55)
        at org.georchestra.security.client.console.GeorchestraConsoleUsersApiImpl.findAll(GeorchestraConsoleUsersApiImpl.java:45)
        at org.georchestra.geonetwork.security.integration.GeorchestraAccountsRepository.findAllUsers(GeorchestraAccountsRepository.java:54)
        at org.geonetwork.security.external.integration.UserSynchronizer.findCanonicalUsers(UserSynchronizer.java:78)
```
[...]

PR #3780, opened 2022-08-04
Raise RestClient read timeout to 60s 

landryb:
As discussed in #3779 this allows the GN4<->LDAP sync to work with ~1k users. To be debated: change the defaults in the datadir repo. And backport to 22.0.


landryb:
fwiw, on another instance on slower hardware, fetching http://localhost:8480/console/internal/users takes 80s. That timeout definitely has to be configurable without modifying the code...

fvanderbiest:
Thanks Landry.
We should take this into account for 22.0.4

fvanderbiest:
Ported to 22.0.x with 4811a196a

PR #3782, opened 2022-08-16
console - account creation - improve the error message if e-mail already in use

pmauduit:
For now, when trying to create an account with a mail address which is already used by another account, the console just displays an error message. This PR provides the URL to the password recovery onto the message, to improve the user experience.

Notes:
* This has already been applied on the console from DataGrandEst.
* I am not really happy with hardcoding the `/console/` endpoint, but we don't have much templating capabilities from here ...
* I have no clue for translating this for the `es` and `nl` translations (poke @groldan || José Macchi).

### DIFF
--- a/console/src/main/java/org/georchestra/console/ws/newaccount/NewAccountFormController.java
+++ b/console/src/main/java/org/georchestra/console/ws/newaccount/NewAccountFormController.java
@@ -61,6 +61,7 @@ import org.georchestra.ds.users.AccountFactory;
 import org.georchestra.ds.users.DuplicatedEmailException;
 import org.georchestra.ds.users.DuplicatedUidException;
 import org.springframework.beans.factory.annotation.Autowired;
+import org.springframework.beans.factory.annotation.Value;
 import org.springframework.stereotype.Controller;
 import org.springframework.ui.Model;
 import org.springframework.validation.BindingResult;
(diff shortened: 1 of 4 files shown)


fvanderbiest:
> * I am not really happy with hardcoding the `/console/` endpoint, but we don't have much templating capabilities from here ...

Being on `/console/account/new` you can try to use `../passwordRecovery` instead

fvanderbiest:
up ?

arnaud-morvan:
Is it not possible to use errorArgs and MessageFormat to bind URL in string:

https://github.com/georchestra/georchestra/blob/d8a3d95566a07faa8a3cadf89e87faa3ca74868d/console/src/main/java/org/georchestra/console/ws/newaccount/NewAccountFormController.java#L323

https://docs.spring.io/spring-framework/docs/current/javadoc-api/org/springframework/validation/Errors.html#rejectValue-java.lang.String-java.lang.String-java.lang.Object:A-java.lang.String-

pmauduit:
> Is it not possible to use errorArgs and MessageFormat to bind URL in string:

sounds possible indeed

pmauduit:
I can't approve because I'm the original author of the PR, but we pair-programmed on this one, so if I could, I'd approve it

arnaud-morvan (approved):

PR #3795, opened 2022-09-07
Log username lowercase in ogcstatistics

arnaud-morvan:
https://jira.camptocamp.com/browse/GEO-5609


pmauduit (approved):
pair-progged together, so LGTM

Issue #3776, opened 2022-07-22
datafeeder - missing namespace alias for WFS layer name


fvanderbiest:
datafeeder generates metadata XML :
![Capture d’écran du 2022-07-22 14-25-10](https://user-images.githubusercontent.com/265319/180439097-ad969c31-5bd4-4bab-b841-a7f4945e2a2e.png)

WFS link does not work with GeoServer, due to missing namespace alias.

Got : `EPCI_SHAPEFILE`
Expected : `sidi:EPCI_SHAPEFILE`

pmauduit:
> WFS link does not work with GeoServer, due to missing namespace alias.

if the gmd:URL is referencing the workspace, then the namespace is not necessary, geoserver is able to find back the layer name.

fvanderbiest:
> if the gmd:URL is referencing the workspace, then the namespace is not necessary, geoserver is able to find back the layer name.

True with WMS, not true with WFS, to my knowledge.

groldan:
Jesus

fvanderbiest:
@jemacchi did you make progress on this one ? Thanks !

jemacchi:
> @jemacchi did you make progress on this one ? Thanks !
@fvanderbiest 
yes, so far now I'm more familiar with the datafeeder codebase and project itself. 
As result of research, found that at this [class](https://github.com/georchestra/georchestra/blob/536214ffa36237360d47bdd80ca417074f60995c/datafeeder/src/main/java/org/georchestra/datafeeder/service/publish/impl/MetadataRecordProperties.java#L113-L118)  we should be defining the qualifiedName instead of using simply the name, so we can ensure that at the definition of [onlineResource name definition](https://github.com/georchestra/georchestra/blob/b6575a3d3911d86dddd65fc7df6b7512d2f4c9f2/datafeeder/src/main/resources/default_iso_2005_gmd.xsl#L327-L348) we get the namespace included into the name (so WFS will find the layer name)

Now I need to check how to apply those changes to datafeeder.

pmauduit:
If we only consider the "GN to Mapstore" scenario, in order to have a MD inserted with the datafeeder working correctly with WFS, I had to edit the WFS section of the MD, so that:

1. I got rid of the parameters in the URL, e.g. keeping the simplest form possible : "https://.../geoserver/psc/wfs"
2. The layer name is fully qualified (with its namespace)

[...]

PR #3799, opened 2022-09-19
datafeeder - fix online resources creation (#3776)

pmauduit:
See issue #3776 for more info. The goal is to fix the different links provided by GeoNetwork on the metadata, to actually be able to access the datas from the catalogue to the (mapstore) viewer.

Note: I am not sure this is the right move in regards to metadata editing best practice but at least generated links work in geOrchestra, and I think this is better to keep the (geOrchestra) consistency.

Tests:
* Runtime tests OK (WMS, WFS, OnlineResources / shapezip)
* `mvn verify` OK
* backward compatibility with mapfishapp *untested*

### DIFF
--- a/datafeeder/src/main/java/org/georchestra/datafeeder/service/publish/impl/GeorchestraMetadataPublicationService.java
+++ b/datafeeder/src/main/java/org/georchestra/datafeeder/service/publish/impl/GeorchestraMetadataPublicationService.java
@@ -276,15 +276,17 @@ public class GeorchestraMetadataPublicationService implements MetadataPublicatio
     private OnlineResource wmsOnlineResource(DatasetUploadState d) {
         String protocol = "OGC:WMS";
         String description = d.getPublishing().getTitle() + " - WMS";
-        String queryString = "SERVICE=WMS&REQUEST=GetCapabilities";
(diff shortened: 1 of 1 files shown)


fvanderbiest:
IIUC, this PR makes datafeeder generate MD XML with data linkage :

- WMS /geoserver/nsalias/ows - layername "layer"
- WFS /geoserver/nsalias/ows - layername "nsalias:layer"

LGTM.

pmauduit:
> IIUC, this PR makes datafeeder generate MD XML with data linkage :

it also removes the parameters in the linkage URL for WMS & WFS (the ones for the shapefile download via WFS are left intact), because Mapstore does not need them and it actually makes mapstore unable to load a layer if they are present.

fvanderbiest:
> it also removes the parameters in the linkage URL for WMS & WFS

:+1: 



> the ones for the shapefile download via WFS are left intact

This linkage is less useful nowadays - we should consider removing it at some point.

fvanderbiest:
Ported to `22.0.x` with ca2a25f9d

pmauduit:
ok, the testsuite is a bit flacky and requires to be launched several times

Issue #3802, opened 2022-09-20
[geonetwork] 404 errors accessing geonetwork with all logins containing a dot


landryb:
sometimes (havent figured when/how yet) when a user access the catalog there's a 404, from my understanding coming from the code added in georchestra for the gn<->ldap sync
```
Etat HTTP 500 – Erreur interne du serveur

Type Rapport d'exception

message 404 : [no body]

description Le serveur a rencontré une erreur interne qui l'a empêché de satisfaire la requête.

exception

org.springframework.web.client.HttpClientErrorException$NotFound: 404 : [no body]
	org.springframework.web.client.HttpClientErrorException.create([HttpClientErrorException.java:113](http://httpclienterrorexception.java:113/))
	org.springframework.web.client.DefaultResponseErrorHandler.handleError([DefaultResponseErrorHandler.java:170](http://defaultresponseerrorhandler.java:170/))
	org.springframework.web.client.DefaultResponseErrorHandler.handleError([DefaultResponseErrorHandler.java:112](http://defaultresponseerrorhandler.java:112/))
	org.springframework.web.client.ResponseErrorHandler.handleError([ResponseErrorHandler.java:63](http://responseerrorhandler.java:63/))
	org.springframework.web.client.RestTemplate.handleResponse([RestTemplate.java:782](http://resttemplate.java:782/))
	org.springframework.web.client.RestTemplate.doExecute([RestTemplate.java:740](http://resttemplate.java:740/))
	org.springframework.web.client.RestTemplate.execute([RestTemplate.java:674](http://resttemplate.java:674/))
	org.springframework.web.client.RestTemplate.getForEntity([RestTemplate.java:342](http://resttemplate.java:342/))
	[org.georchestra.security](http://org.georchestra.security/).client.console.RestClient.get([RestClient.java:40](http://restclient.java:40/))
	[org.georchestra.security](http://org.georchestra.security/).client.console.GeorchestraConsoleUsersApiImpl.findByUsername([GeorchestraConsoleUsersApiImpl.java:57](http://georchestraconsoleusersapiimpl.java:57/))
	[org.georchestra.geonetwork.security](http://org.georchestra.geonetwork.security/).integration.GeorchestraAccountsRepository.findUserByUsername([GeorchestraAccountsRepository.java:58](http://georchestraaccountsrepository.java:58/))
[...]

PR #3817, opened 2022-10-03
unbreak users with a . in their login (#3802)

landryb:
per https://www.baeldung.com/spring-mvc-pathvariable-dot spring's @PathVariable annotation strips what's after a dot, thinking its a file extension


pmauduit (commented):
LGTM ...

Issue #3823, opened 2022-10-07
gh actions - disable login to docker-hub if dependabot is the actor


pmauduit:
The CI will generally fail at the "login in to docker hub" step when triggered by dependabot, because it does not have access to the secrets from the repository. e.g.:
https://github.com/georchestra/georchestra/actions?query=branch%3Adependabot%2Fmaven%2Fmaster%2Fgt.version-26.4

we probably have to skip the steps related to sending a docker image when `github.actor == 'dependabot[bot]'`, relying on the documentation (https://docs.github.com/en/code-security/dependabot/working-with-dependabot/automating-dependabot-with-github-actions).

pmauduit:
Missing the backport to 22.0.x to comply with the milestone

pmauduit:
backported.

PR #3837, opened 2022-10-28
cicd - skip docker-related steps when actor is dependabot (#3823)

pmauduit:
see #3823 for the motivation


fvanderbiest:
LGTM

fvanderbiest:
We should probably backport to stable branches too

Issue #3755, opened 2022-05-31
GeoServer 2.21


fvanderbiest:
https://geoserver.org/announcements/2022/05/24/geoserver-2-21-0-released.html

Points of attention when upgrading:

- log4j2 move
- The gdal-output extension is tested against GDAL 3.x series.

Regarding log4j, `georchestra/pom.xml` (root pom) declares dependencies which probably need an update for log4j2 (thanks for the hint @groldan):

```
    <profile>
      <id>log4j-logstash</id>
      <dependencies>
        <dependency>
          <groupId>net.logstash.log4j</groupId>
          <artifactId>jsonevent-layout</artifactId>
          <version>1.7</version>
        </dependency>
      </dependencies>
    </profile>
...
    <profile>
      <id>sentry-log4j</id>
      <dependencies>
        <dependency>
          <groupId>io.sentry</groupId>
          <artifactId>sentry-log4j</artifactId>
          <version>1.7.30</version>
        </dependency>
      </dependencies>
    </profile>
```

fvanderbiest:
https://geoserver.org/announcements/2022/12/28/geoserver-2-21-3-released.html

fvanderbiest:
geOrchestra 23.0 comes with GeoServer 2.22.2: https://github.com/georchestra/georchestra/blob/2766a33a16837e5b18f35262ed86e5af2f542b88/geoserver/pom.xml#L11

fvanderbiest:
@pmauduit would you be able to check that geoserver 2.22.2 on master does not provide several log4j versions ?
Thanks !

PR #3844, opened 2022-11-06
Add GeoServer plugin geopkg-output

arnaud-morvan:


fvanderbiest:
This extension is not available before GeoServer 2.21 => depends on #3755

PR #3851, opened 2022-11-24
sec-proxy: Allow to configure cookie affinity between backend services

groldan:
## Cookie Affinity Mapping

Sometimes, cookies sent by one backing service need to be readable by another. The Security-Proxy will set a cookie path to all backend service cookies to match the service base path (for example, all cookies sent by the `console` application will have their path set to `/console`. This makes it impossible for other applications to read them. A clear case is when the `datahub` application, under the `/datahub` context path, needs access to the GeoNetwork `XSRF-TOKEN` issued cookie.

**Cookie Affinity Mapping** allows to duplicate cookies set to one path with another path. For the example above, we need to make it so the GeoNetwork `XSRF-TOKEN` cookie is sent twice to the client, once with `Path=/geonetwork` and once with `Path=/datahub`.

To this end, the file `<datadir>/security-proxy/cookie-mappings.json` can be used to configure such cookie affinity. It shall contain an array of objects like the following:

```json
[
	{
		"name": "XSRF-TOKEN",
		"from": "/geonetwork",
		"to": "/datahub"
	},
	{
		"name": "my-custom-cookie",
		"from": "/console",
		"to": "/analytics"
	},
]
```

The `name` property indicates the cookie name, the `from` property indicates from which original path the cookie will be duplicated, and the `to` property which path to duplicate the cookie with.

## Refactor HeadersManagementStrategy cookie handling for clarity
Refactor HeadersManagementStrategy cookie hanlding to use Cookie objects and HttpServletResponse.setCookie(Cookie) instead of string manging of header response values.

### DIFF
--- a/security-proxy/src/main/java/org/georchestra/security/Proxy.java
+++ b/security-proxy/src/main/java/org/georchestra/security/Proxy.java
@@ -125,6 +125,8 @@ import com.google.common.annotations.VisibleForTesting;
 import com.google.common.collect.ImmutableMap;
 import com.google.common.collect.Maps;
 
+import lombok.NonNull;
+
 /**
  * This proxy provides an indirect access to a remote host to retrieve data.
  * Useful to overcome security constraints on client side.
--- a/security-proxy/README.md
+++ b/security-proxy/README.md
@@ -70,6 +70,41 @@ Logout entrypoint is `/logout`.
 Password recovery form is available from `/console/account/passwordRecovery`.
 Account creation form can be found at `/console/account/new`.
 
+
+#### Cookie Affinity Mapping
+
+Sometimes, cookies sent by one backing service need to be readable by another.
+The Security-Proxy will set a cookie path to all backend service cookies to match the service base path (for example,
+all cookies sent by the `console` application will have their path set to `/console`.
+This makes it impossible for other applications to read them.
+A clear case is when the `datahub` application, under the `/datahub` context path, needs access to the
+GeoNetwork `XSRF-TOKEN` issued cookie.
+
+**Cookie Affinity Mapping** allows to duplicate cookies set to one path with another path. For the example above,
+we need to make it so the GeoNetwork `XSRF-TOKEN` cookie is sent twice to the client, once with `Path=/geonetwork`
+and once with `Path=/datahub`.
+
+To this end, the file `<datadir>/security-proxy/cookie-mappings.json` can be used to configure such
+cookie affinity. It shall contain an array of objects like the following:
+
+```json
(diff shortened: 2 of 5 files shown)


fvanderbiest:
Thanks Gabriel ! Sounds like a perfect fit.

security-proxy/src/main/java/org/georchestra/security/Proxy.java
```diff
@@ -125,6 +125,8 @@
 import com.google.common.collect.ImmutableMap;
 import com.google.common.collect.Maps;
 
+import lombok.NonNull;
```
pmauduit, line 128: unused ?

pmauduit (approved):

pmauduit:
LGTM, I can't recall why I could not use the cookie parsing library from javax (and doing string parsing by hand instead) though, when I reworked this some years ago.

groldan:
> LGTM, I can't recall why I could not use the cookie parsing library from javax (and doing string parsing by hand instead) though, when I reworked this some years ago.

FWIW, I didn't see a cookie parsing utility in the javax api

pmauduit:
> FWIW, I didn't see a cookie parsing utility in the javax api

I meant this package:
https://github.com/georchestra/georchestra/pull/3851/files#diff-2277cfdae3cdd7c948248da5e767abcff14c909065050fc3bce300d479dfb873R53

landryb:
> Cookie Affinity Mapping allows to duplicate cookies set to one path with another path. For the example above, we need to make it so the GeoNetwork XSRF-TOKEN cookie is sent twice to the client, once with Path=/geonetwork and once with Path=/datahub.

I know it was merged etc, but for more clarity, can you explain what user-facing issue this is solving ? some wrong behaviour when datahub tries to load data from geonetwork ?

With your PR, reading the code i understand that the sp is responsible of duplicating the cookie sent to the client ? that's.. clever :)

pmauduit:
> I know it was merged etc, but for more clarity, can you explain what user-facing issue this is solving ? some wrong behaviour when datahub tries to load data from geonetwork ?

the datahub (a purely JS / browser app) was in the need of accessing a cookie from GeoNetwork (for "favorite items" or so, I honestly don't know much of the details). But something in the SP I reworked several years ago forbids this, because every cookies intercepted by the SP were enforced with a path to be set from the originating webapp. So a cookie originating from geonetwork was enforced on the "/geonetwork" path, making it unavailable from another path (e.g. /datahub).

landryb:
thanks for the details, that's .. twisted :)

PR #3853, opened 2022-12-02
Ldap logo to gn

cmangeat:
bring an entry point at console / security side so to retrieve an organisation logo, given its uuid, canonical, org lastUpdated hash taking into account logo updates.

logo can be retrieved from org uuid, but do not travel with common module organisation dto over network.

linked with https://github.com/georchestra/geonetwork/pull/218.


fgravin (approved):
Great work @cmangeat !
LGTM

PR #3869, opened 2022-12-30
console: add area of competence API endpoint for the request's user

groldan:
Add a new `GET /console/account/areaofcompetence` API endpoint to return the calling user's area of competence geometry as a GeoJSON geometry.

The user's area of competence is computed from the list of city identifiers in the Organization it belongs to.

It uses GeoTools' GeoJSON data store to query the features configured in `console.properties` under the `AreaUrl` key, just like `AreaController` (which instead returns the `cities.geojson` file as-is).

From the deafult georchestra configuration, the `testadmin` user has access to two cities, and the result is like:

image

while the `testeditor` user has access to all 360 cities, resulting in:

image

## Performance

The operation consists of three steps:
* Obtain the city ids from the user's organization, this is negligible
* Obtain all the geometries from `cities.geojson` whose `INSEE_COM` attribute matches one of the city ids.
* Union all geometries to return a single multipolygon

Setting the following logger level in `<datadir>/console/log4j/log4j.properties`: `log4j.logger.org.georchestra.console.bs.areas=DEBUG`, you'll get something like this:

```
console_1                            | 2022-12-30 23:37:50 AreasService [DEBUG] Computing area of competence for user testeditor, org Camptocamp
console_1                            | 2022-12-30 23:37:50 AreasService [DEBUG] Queried 360 geometries in 31.56 ms
console_1                            | 2022-12-30 23:37:50 AreasService [DEBUG] Unioned 360 geometries in 108.0 ms
```
The geometry union operation runs in parallel, which reduces the union of the 360 polygons from > 2 seconds to ~100 ms in my machine.

```
export url="https://georchestra-127-0-1-1.traefik.me/console/account/areaofcompetence"
ab -A testeditor:testeditor -n 100 -c 10 $url
This is ApacheBench, Version 2.3 <$Revision: 1879490 $>
Copyright 1996 Adam Twiss, Zeus Technology Ltd, http://www.zeustech.net/
Licensed to The Apache Software Foundation, http://www.apache.org/

Benchmarking georchestra-127-0-1-1.traefik.me (be patient).....done


Server Software:        
Server Hostname:        georchestra-127-0-1-1.traefik.me
Server Port:            443
SSL/TLS Protocol:       TLSv1.2,ECDHE-RSA-AES256-GCM-SHA384,2048,256
Server Temp Key:        X25519 253 bits
TLS Server Name:        georchestra-127-0-1-1.traefik.me

Document Path:          /console/account/areaofcompetence
Document Length:        28914 bytes

Concurrency Level:      10
Time taken for tests:   2.726 seconds
Complete requests:      100
Failed requests:        0
Total transferred:      2944698 bytes
HTML transferred:       2891400 bytes
Requests per second:    36.68 [#/sec] (mean)
Time per request:       272.602 [ms] (mean)
Time per request:       27.260 [ms] (mean, across all concurrent requests)
Transfer rate:          1054.90 [Kbytes/sec] received

Connection Times (ms)
              min  mean[+/-sd] median   max
Connect:        6   10   6.3      8      29
Processing:   157  238  32.1    236     364
Waiting:      157  238  31.9    235     356
Total:        168  249  33.3    246     370

Percentage of the requests served within a certain time (ms)
  50%    246
  66%    257
  75%    264
  80%    270
  90%    288
  95%    313
  98%    349
  99%    370
 100%    370 (longest request)
```


groldan:
NOTE: for non-logged in users to receive a `404` when calling this service, we need to add the following line to `security-proxy/security-mappings.xml`: `<intercept-url pattern="/console/account/areaofcompetence" access="IS_AUTHENTICATED_ANONYMOUSLY" />`, which we can do once this is reviewed and approved.

fgravin (approved):
It's clean, great work thanks @groldan !
Thanks for the refactoring of the tests as well.

What is the error code if we are not logged in ATM ?
You propose a 404 by adding a rule to the config-security mappers but i don't know what code would be the best.

fvanderbiest:
Currently we have this rule:
https://github.com/georchestra/datadir/blob/1579ddcb959f90fc91aeeb6f14b63fea349b3de0/security-proxy/security-mappings.xml#L22 which means that the request gets a 302 redirect to /cas/login 

I think it's OK like this, since the service should not be called for anonymous visitors of the datahub (to be double checked though).

Issue #2271, opened 2018-11-22
Record user last connection timestamp


fvanderbiest:
This information could be displayed in the console.
This is useful to some platform admins.

We currently record their use of webservices, not their latest CAS login action.

fvanderbiest:
This would be a "nice to have" for the Geo2France project, ping @bchartier

fvanderbiest:
Also interested: GeoGrandEst

MaelREBOUX:
Me ! Me ! Me ! Me !
Since 2016 :p

pmauduit:
Since the talk took place on IRC, there are several technical ways of achieving this:

* Using the slapo-lastbind overlay (https://manpages.ubuntu.com/manpages/bionic/man5/slapo-lastbind.5.html) :
  * :+1: :-1: :question: this will also track users authenticated via basic-auth
  * :-1: it won't work for setups making use of something else than the geOrchestra OpenLDAP (wink wink @MaelREBOUX)
* logging onto cas via log4j2 config (suggested by @jusabatier):
  * :+1: :-1: :question: will only track users authenticating via CAS (web-based)

jusabatier:
- logging onto Security Proxy via log4j2 config : 
  - :+1: this will also track users authenticated via basic-auth
  - :-1: this will generate more logs (each request) / :+1: those logs can be used to generate metrics

landryb:
i thought a bit about the pros/cons of enabling the lastbind overlay, was a bit concerned at first that it might generate *tons* of writes to the ldap server (eg if you have many connected users using an authenticated WMS...) **but** after looking at the code a bit, i'm relieved to say that it has been thought ahead and is handled by a configurable timeout/precision, per comment in https://github.com/openldap/openldap/blob/master/contrib/slapd-modules/lastbind/lastbind.c#L25

it has been recently changed in https://github.com/openldap/openldap/commit/7a4b42600b4fc149e1d61784fa19a63be63412dc (so for openldap 2.6.1, per https://lists.openldap.org/hyperkitty/list/openldap-announce@openldap.org/thread/VGSH3HMNASR6PP72CHV5BVNID5WOUXUN/) , previously there was a dedicated `olcLastBindPrecision` attribute in the overlay to configure the precision, but now apparently it's supposed to be "the backend's precision", whatever that is.. i'll dig further.

so i think in the coming days i'll have a look at:
[...]

PR #3889, opened 2023-02-02
Display last user connection & account creation date from OpenLDAP attributes

landryb:
will fix/implement #2271 but more work needed.


jusabatier:
Is it possible to also retrieve and display account creation date ?

It should help to identify recently created accounts with no connection.

jusabatier:
Maybe you should just add some documentation on how to install the LDAP overlay in install doc : 

https://github.com/georchestra/georchestra/blob/master/docs/setup/openldap.md

landryb:
> Maybe you should just add some documentation on how to install the LDAP overlay in install doc :
> 
> https://github.com/georchestra/georchestra/blob/master/docs/setup/openldap.md

like https://github.com/georchestra/georchestra/blob/e0e4590a533d55472929de65808d41d8717a8924/docs/setup/openldap.md#the-lastbind-overlay ? i've rewritten a bit the doc while here in c11e41f

fvanderbiest:
So, you say that ldap writes are "buffered" in some way ? I'm really concerned about this for the basic-auth case specifically...

landryb:
> So, you say that ldap writes are "buffered" in some way ? I'm really concerned about this for the basic-auth case specifically...

what is your concern ? the code of the overlay only writes if the attribute value is older than the configured precision (cf https://github.com/openldap/openldap/blob/master/contrib/slapd-modules/lastbind/lastbind.c#L25 & https://github.com/openldap/openldap/blob/master/contrib/slapd-modules/lastbind/lastbind.c#L139), to avoid updating the attribute at each hit of a basic-auth ogc service for example. If the last auth was more recent than the precision, no write happens, there's no "buffering".

landryb:
image

fvanderbiest (approved):
LGTM once translations have been provided

Issue #3887, opened 2023-01-27
datafeeder - integration tests are flacky


pmauduit:
The CI (esp. the IT on the Datafeeder) sometimes fails, but this is not systematic, when they fail, here are the tests in error:

```
Error:  Errors: 
Error:    GeorchestraOwsPublicationServiceIT.addMetadataLinks:239 » BadRequest Status 40...
Error:    GeorchestraOwsPublicationServiceIT.testPublishSingleShapefile:189 » BadRequest
Error:    GeorchestraOwsPublicationServiceIT.testPublish_sets_cache_settings:216 » BadRequest
Error:    GeorchestraOwsPublicationServiceIT.testPublish_sets_layer_attribution_to_orgname_and_url:203 » BadRequest
[INFO] 
Error:  Tests run: 23, Failures: 0, Errors: 4, Skipped: 0
``` 

generally one or 2 retries and we are good to go, but for the 22.0.5 release, I needed to relaunch the CI 5 times.

pierrejego:
All PR are impact, no github action checks are ok. 

For me, if nobody have time to correct them they should be ignore
@ignore on specific tests

pmauduit:
> All PR are impact, no github action checks are ok.

this is weird, as far as I remember, only the datafeeder github action was suffering from this.

PR #3898, opened 2023-02-09
Add /userdetails

arnaud-morvan:


arnaud-morvan:
@pmauduit
Still an error in CI with datafeeder, but seems already the case on master.

pmauduit:
> Still an error in CI with datafeeder, but seems already the case on master.

yes, this one is known: https://github.com/georchestra/georchestra/issues/3887

arnaud-morvan:
@jeanmi151 
As this is just an add, no change in existing behavior, I suppose you can merge it without risk.

PR #3899, opened 2023-02-13
Added configurable password requirements

emmdurin:
PR for GIP [password strength check](https://github.com/georchestra/improvement-proposals/issues/2)

[Link to Datadir PR#272](https://github.com/georchestra/datadir/pull/272)

Remaining opened issues :
1. ~~redefine behavior and content of Javascript function "isPasswordValid" which was doing some checks on the password when it is being typed and was showing some messages to the user~~ replaced original length check by a non-empty check similar to other fields
2. ~~a question about UX : when password requirements are not met, which is better between showing only missing requirements or showing all requirements ? For now first one is implemented~~ will stay on initial implementation


fvanderbiest:
> 2\. a question about UX : when password requirements are not met, which is better between showing only missing requirements or showing all requirements ? For now first one is implemented

Can you share a screenshot ? Thanks !

emmdurin:
I gave a password with only lowers and special characters, and I get the following output telling me that the password must contain uppers and digits.
password_error

Should the output tell that the password must also contain lowers and special characters ? We may imagine cases where user enters a password with one digit, gets an error about missing special characters, replaces the digit by a special character, then gets a new error about missing digits.

fvanderbiest:
I think the UX is fine, thanks Emmanuel.

landryb:
Nice ! i know this is still a draft, but there should also be a PR for the datadir repo adding sensible discoverable defaults to console.properties ?

console/src/main/webapp/WEB-INF/views/validation.jsp
```diff
@@ -272,11 +272,7 @@ function isNotEmpty(str) {
  * @returns {boolean}
  */
 function isPasswordValid(password) {
```
cmangeat, line 274: dead code ?

cmangeat (approved):

console/src/main/webapp/WEB-INF/views/validation.jsp
emmdurin: replaced original length check by a non-empty check similar to other fields

pmauduit (approved):

jeanmi151:
go to me 
image

Issue #3468, opened 2021-07-08
[Console] Problem persisting empty area for an org to the LDAP


jusabatier:
I tested with a war from 20.1.x and an empty area for an org isn't persisted to the LDAP.

Same behavior as this old issue : https://github.com/georchestra/georchestra/issues/2471

To reproduce : 

1. Assign an area to an org and save
2. Remove the area from the org and save
3. Refresh and the area still assigned to the org

fvanderbiest:
Thanks for the report Julien.
This means #2869 did not fix the original issue IIUC.
Bumping.

fvanderbiest:
Feel free to provide a patch. Thanks !

pmauduit:
see PR above (codacy check fixed in a second commit).

PR #3913, opened 2023-03-20
console - fix aera of competence removal (#3468)

pmauduit:
When removing all the aeras of competence from an organization, the 'description' attribute of the OrgExt has to be removed from the LDAP object. Without this modification, it is kept intact.

Tests:

* IT added
* runtime tested

Thanks @cmangeat for the help / pair programming.


pmauduit:
Tested at runtime in a docker composition:

console-3468

fvanderbiest:
Looks like it is also eligible to a backport on latest stable ...

Issue #3923, opened 2023-03-30
org.geowebcache.mime.MimeException: Unsupported format request: text/javascript


fvanderbiest:
With the `ENABLE_JSONP=true` option, the integrated geowebcache fills the logs with the following message:
```
30 Mar 13:57:54 WARN   [gwc.layer] - Unsupported format request: text/javascript
org.geowebcache.mime.MimeException: Unsupported format request: text/javascript
```

This was also reported [elsewhere](https://sourceforge.net/p/geoserver/mailman/geoserver-users/thread/5c025fbb1b9d4091b1eaa1bdfdfedfa7@digipolis.gent/)

PR #3924, opened 2023-03-30
Geoserver: disable JSONP

fvanderbiest:
see #3923 for the motivation


landryb:
i remember jsonp was being used by a previous project we had, but that site was decommissioned... still, dragons might be lurking here and there and this might be in use at some places :)

fvanderbiest:
"only affects docker" - you can relax ;-)

landryb:
i know i know, but we should avoid discrepancies between setup types :)

Issue #3932, opened 2023-04-07
build stuck on datafeeder-ui


landryb:
since some recent activity, some builds are failing/getting stuck on datafeeder-ui:
```
[INFO] --- frontend-maven-plugin:1.11.3:npm (npm run build) @ datafeeder-ui ---
[INFO] Running 'npm run build datafeeder -- --base-href=/import/' in /var/lib/buildbot/workers/georchestra/buildjob-master/build/datafeeder-ui/target/datafeeder-ui
[INFO]
[INFO] > geonetwork-ui@1.0.1 build /var/lib/buildbot/workers/georchestra/buildjob-master/build/datafeeder-ui/target/datafeeder-ui
[INFO] > nx affected:build "datafeeder" "--base-href=/import/"
[INFO]
[INFO]
[INFO]  >  NX   Affected criteria defaulted to --base=remotes/origin/georchestra-datafeeder --head=HEAD
[INFO]
[INFO]
[INFO]  >  NX   Running target build for 12 projects:
[INFO]
[INFO]     - api-metadata-converter
[INFO]     - metadata-converter
[INFO]     - metadata-editor
[INFO]     - datafeeder
[INFO]     - demo
[INFO]     - util-data-fetcher
[INFO]     - datahub
[INFO]     - webcomponents
[INFO]     - search
[INFO]     - map-viewer
[INFO]     - data-platform
[INFO]     - util-types
[INFO]
[INFO]    With additional flags:
[INFO]      datafeeder
[INFO]
[INFO]
```
and nothing after that, see the failing logs at  https://packages.georchestra.org/bot/#/builders/3/builds/296/steps/2/logs/stdio (and previous/next builds)

looking at the running processes when the build is waiting/idling, i can see a "defunct" shell process:
```
buildbot  411079  0.1  0.1 727248 30536 pts/3    Sl+  09:51   0:00 npm
buildbot  411093  0.0  0.0   2496   508 pts/3    S+   09:51   0:00 sh -c nx affected:build "datafeeder" "--base-href=/import/"
buildbot  411094  0.5  0.2 633892 43924 pts/3    Sl+  09:51   0:01 node /var/lib/buildbot/workers/georchestra/buildjob-master/build/datafeeder-ui/target/datafeeder-ui/node_modules/.bin/nx affected:build datafeeder --base-href=/import/
[...]

PR #3933, opened 2023-04-07
df-ui - attempt to fix js build (#3932)

pmauduit:
* upgrading node version in the maven plugin
* adapting build commands to what is documented in the upstream README.md

Tests: checked a georchestra/datafeeder-ui:latest against a locally generated one, and the documentroot on both sides were taking the same size.

Note: as the same repository is used in the 22.0.x, this might require a backport.

### DIFF
--- a/datafeeder-ui/pom.xml
+++ b/datafeeder-ui/pom.xml
@@ -91,17 +91,17 @@
                       </configuration>
                   </execution>
                   <execution>
-                      <id>npm run build</id>
+                      <id>build</id>
                       <goals>
-                          <goal>npm</goal>
+                          <goal>npx</goal>
                       </goals>
                       <configuration>
-                          <arguments>run build datafeeder -- --base-href=/import/</arguments>
+                          <arguments>nx build datafeeder -- --base-href=/import/</arguments>
                       </configuration>
                   </execution>
               </executions>
               <configuration>
-                  <nodeVersion>v12.22.1</nodeVersion>
+                  <nodeVersion>v18.15.0</nodeVersion>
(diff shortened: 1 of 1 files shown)


landryb:
i've been able to build master with [af6e1df](https://github.com/georchestra/georchestra/pull/3933/commits/af6e1dfe607c0d0c146c320d48f8f435b8e6c56a) so that should indeed fix #3932

pmauduit:
Still in a WIP state, we are currently losing track of the GHA run after ~10 / 15 minutes. It works locally on my dev env, but repairing the ITests here sounds mandatory for the release.

pmauduit:
Also wondering if Kibana is necessary in the docker composition, not sure we need it, a commit was removing it (fb93c7c539cd28e3af05e9848cf813b4fa99e4b5), but there are still some leftovers here:
https://github.com/georchestra/georchestra/tree/master/datafeeder/src/test/resources/datadir/kibana

jahow:
Well for what it's worth at least the issues with datafeeder-ui seem resolved, right?

pmauduit:
> Well for what it's worth at least the issues with datafeeder-ui seem resolved, right?

yes but we will need to have all green for the release anyway

pmauduit:
Build fixed, it required to cap the XMX on the ES container, at least to get back to have the worker not disappearing after some minutes (after the IT docker compo has been launched).

Tests are still a bit flacky and requires to be relaunch several times sometimes (see https://github.com/georchestra/georchestra/issues/3887), but the situation is better than the previous one.

PR #3943, opened 2023-04-24
new passwords management policy -ldap

marwanehcine:
A new policy management for password called rotation is implemented  at georchestra, where user has to set password periodically. rotation policy is not activated by default. to build a docker image with this new configuration run : docker-build-ldap-withrotation.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -10,6 +10,11 @@ docker-build-ldap:
 	docker pull debian:bullseye
 	cd ldap; \
 	docker build -t georchestra/ldap:${BTAG} .
+	
+docker-build-ldap-withrotation:
+	docker pull debian:bullseye
+	cd ldap; \
+	docker build -t georchestra/ldap:${BTAG} --build-arg PM_POLICY=rotation .
 
 docker-build-database:
 	docker pull postgres:15
--- a/ldap/docker-root/pwd_no_expire_users.ldif
+++ b/ldap/docker-root/pwd_no_expire_users.ldif
@@ -1,0 +1,9 @@
+dn: uid=geoserver_privileged_user,ou=users,dc=georchestra,dc=org
+changetype: modify
+add: pwdPolicySubentry
+pwdPolicySubentry: cn=pwd-no-expire,ou=pwpolicy,dc=georchestra,dc=org
+
+dn: uid=idatafeeder,ou=users,dc=georchestra,dc=org
+changetype: modify
+add: pwdPolicySubentry
+pwdPolicySubentry: cn=pwd-no-expire,ou=pwpolicy,dc=georchestra,dc=org
--- a/migrations/master/README.md
+++ b/migrations/master/README.md
@@ -1,2 +1,66 @@
 # From 23.0 to master
 
+## cas
+From 23.0 to 23.1
+To activate password rotation policy for ldap, user need to run these commands:
+
+```
+ldapmodify -H ldap://localhost:389 -D cn=admin,dc=georchestra,dc=org -w secret
+dn: olcOverlay={0}ppolicy,olcDatabase={1}mdb,cn=config
+changetype: modify
+replace: olcPPolicyDefault
+olcPPolicyDefault: cn=default,ou=pwpolicy,dc=georchestra,dc=org
+
+```
+```
+ldapadd -H ldap://localhost:389 -D cn=admin,dc=georchestra,dc=org -w secret
+dn: ou=pwpolicy,dc=georchestra,dc=org
+objectClass: organizationalUnit
+objectClass: top
+ou: pwpolicy
+
+dn: cn=default,ou=pwpolicy,dc=georchestra,dc=org
+objectClass: person
+objectClass: pwdPolicyChecker
+objectClass: pwdPolicy
+cn: pwpolicy
+sn: pwpolicy
+pwdAttribute: userPassword
+pwdMinAge: 0
+pwdMaxAge: 31536000
+pwdGraceAuthnLimit: 0
+pwdExpireWarning: 2592000
+
+dn: cn=pwd-no-expire,ou=pwpolicy,dc=georchestra,dc=org
+objectClass: person
+objectClass: pwdPolicyChecker
+objectClass: pwdPolicy
+cn: pwpolicy
+cn: pwd-no-expire
+sn: pwpolicy
+pwdAttribute: userPassword
+pwdMinAge: 0
+pwdMaxAge: 0
+
+```
+To activate rotation password policy for all users, please run script 'set_rotation_policy_for_all_users.sh' located at 'georchestra/migrations/23.0' :
+```
+sh set_rotation_policy_for_all_users.sh 
+```
+To desactivate rotation password policy for non humain users, please run :
+```
+ldapmodify -H ldap://localhost:389  -D cn=admin,dc=georchestra,dc=org -w secret
+dn: uid=geoserver_privileged_user,ou=users,dc=georchestra,dc=org
+changetype: modify
+add: pwdPolicySubentry
+pwdPolicySubentry: cn=pwd-no-expire,ou=pwpolicy,dc=georchestra,dc=org
+
+dn: uid=idatafeeder,ou=users,dc=georchestra,dc=org
+changetype: modify
+add: pwdPolicySubentry
+pwdPolicySubentry: cn=pwd-no-expire,ou=pwpolicy,dc=georchestra,dc=org
+```
+
+
+
+
--- a/ldap/Dockerfile
+++ b/ldap/Dockerfile
@@ -4,6 +4,7 @@
 
 FROM debian:bullseye
 
+ARG PM_POLICY=default
 ENV OPENLDAP_VERSION 2.4.57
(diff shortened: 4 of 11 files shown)


pmauduit (commented):
Just a question related to `olcPPolicyHashCleartext` attribute, otherwise looks good.

I was also wondering if it would be worth to enable/disable this feature using an environment variable, or if we want to enforce the policy by default. Maybe it would require some extra documentation here: https://github.com/georchestra/georchestra/blob/master/docs/setup/openldap.md

ldap/docker-root/etc/ldap.dist/modules/ppolicy.ldif
```diff
@@ -7,7 +7,7 @@ dn: olcOverlay=ppolicy,olcDatabase={1}mdb,cn=config
 objectClass: olcOverlayConfig
 objectClass: olcPPolicyConfig
 olcOverlay: ppolicy
-olcPPolicyDefault: PPOLICY_DN
-olcPPolicyHashCleartext: FALSE
+olcPPolicyDefault: cn=default,ou=pwpolicy,dc=georchestra,dc=org
+olcPPolicyHashCleartext: TRUE
```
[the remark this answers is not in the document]
landryb: *looks at calendar*, wait, is it 1990 ? ;)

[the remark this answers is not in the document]
pmauduit: > looks at calendar, wait, is it 1990 ? ;)

By default, slapd accepts unencrypted passwords anyway

fvanderbiest:
> I was also wondering if it would be worth to enable/disable this feature using an environment variable, or if we want to enforce the policy by default

The [GIP](https://github.com/georchestra/improvement-proposals/issues/3) defines the feature as disabled by default.
So, we either need documentation to enable the feature, or an admin-level switch.
Thanks !

ldap/docker-root/pwpolicyoverlay.ldif
```diff
@@ -0,0 +1,27 @@
...
+
+dn: cn=default,ou=pwpolicy,dc=georchestra,dc=org
+objectClass: person
+objectClass: pwdPolicyChecker
+objectClass: pwdPolicy
+cn: pwpolicy
+sn: pwpolicy
+pwdAttribute: userPassword
+pwdMinAge: 0
+pwdMaxAge: 600
```
fvanderbiest, line 14: rather 1 year by default

ldap/docker-root/etc/ldap.dist/modules/ppolicy.ldif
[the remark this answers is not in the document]
marwanehcine: > 

No, I will keep it FALSE, I think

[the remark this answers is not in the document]
marwanehcine: > Why allowing cleartext passwords ? is it to ensure SASL capabilities ?

No I will keep it FALSE I think!

Makefile
```diff
@@ -9,7 +9,7 @@ docker-pull-jetty:
 docker-build-ldap:
 	docker pull debian:buster
 	cd ldap; \
-	docker build -t georchestra/ldap:${BTAG} .
+	docker build -t georchestra/ldap:${BTAG} --build-arg PM_POLICY=rotation .
```
fvanderbiest, line 12: I would propose 2 makefile targets : one `docker-build-ldap`, unchanged. A new one `docker-build-ldap-withrotation`.

fvanderbiest (changes requested):
LGTM, please have a look at my comment.
Also, please make sure that PRs have a full description before submitting. It helps a lot.
Thanks !

Makefile
marwanehcine: code updated

fvanderbiest:
> Maybe it would require some extra documentation here: https://github.com/georchestra/georchestra/blob/master/docs/setup/openldap.md

Agreed, a bit of documentation is required for those who want to activate the feature (eg when not using docker).

fvanderbiest:
As privately reported by @jeanmi151, (kudos !) we also need to disable password rotation on `geoserver_privileged_user` since it is not owned by any real user

ldap/docker-root/pwd_no_expire_users.ldif
```diff
@@ -0,0 +1,4 @@
+dn: uid=geoserver_privileged_user,ou=users,dc=georchestra,dc=org
```
fvanderbiest, line 1: we also need no-expire policy on `idatafeeder` user, see the recent #3953

fvanderbiest:
Targeting 23.1 for this one. 
Do not merge before 23.0

fvanderbiest:
We may need a documentation for users to upgrade the LDAP to have this new policy
The documentation should go to https://github.com/georchestra/georchestra/tree/master/migrations/master

migrations/master/README.md
```diff
@@ -61,3 +61,78 @@ services:
...
+## cas
+From 22.x to 23.x
+To activate password rotation policy for ldap, user need to run these commands:
+
+```
+ldapmodify -H ldap://  -D cn=admin,dc=georchestra,dc=org -w secret
+dn: cn=module{0},cn=config
+changetype: modify
+add: olcModuleLoad
+olcModuleLoad: ppolicy.la
```
landryb, line 75: fwiw in 23.0 ppolicy module is already loaded since https://github.com/georchestra/georchestra/blob/31d5a963d5d949b30e36295f133ccc3af8c86aa2/ldap/docker-root/lastbind.ldif#L6 so im not sure this bit is needed

you have some hardcoded `172.23.0.11`  below :)

marwanehcine: yep, will update code and check if necessary

migrations/master/README.md
```diff
@@ -61,3 +61,78 @@ services:
 ```
 
 You might need to reuse or adapt those changes for your setup.
+
+
+## cas
+From 22.x to 23.x
```
fvanderbiest, line 67: From 23.0 to 23.1

migrations/master/README.md
```diff
@@ -61,3 +61,78 @@ services:
...
+objectClass: pwdPolicy
+cn: pwpolicy
+cn: pwd-no-expire
+sn: pwpolicy
+pwdAttribute: userPassword
+pwdMinAge: 0
+pwdMaxAge: 0
+```
+```
+ldapmodify -H ldap://localhost:389 -D cn=admin,dc=georchestra,dc=org -w secret
```
fvanderbiest, line 125: useless, since was already done from 22.0 to 23.0

marwanehcine: Here, I'm changing policy to no expire policy so these users will not be included in rotation password policy. I think, this is necessary @fvanderbiest.

fvanderbiest: You're 100% right !
Sorry Marwane. I was too quick in my review

Makefile
```diff
@@ -10,6 +10,11 @@ docker-build-ldap:
 	docker pull debian:bullseye
 	cd ldap; \
 	docker build -t georchestra/ldap:${BTAG} .
+	
+docker-build-ldap-withrotation:
+	docker pull debian:buster
```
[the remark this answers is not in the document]
marwanehcine: Image pulled changed, please check.
Regarding new makefile target, it was required by @fvanderbiest , please check https://github.com/georchestra/georchestra/pull/3943/files#diff-76ed074a9305c04054cdebb9e9aad2d818052b07091de1f20cad0bbac34ffb52

I agree with him, since we need to add option if we need rotation password policy.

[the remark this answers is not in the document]
pmauduit: > I agree with him, since we need to add option if we need rotation password policy.

But the option will be consumed at runtime, I'd have expected to have the same ldap docker image being generated, and the option at bootstrap which will decide if we want password rotation or not, this is why I don't get why we need a new makefile target (anyway since I don't use the makefile, I don't mind, if you both think this is needed, I'm fine)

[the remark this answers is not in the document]
fvanderbiest:
```
ARG PM_POLICY=default
ENV SLAPD_PASSWORD_MGT_POLICY ${PM_POLICY}
```
My understanding of these 2 lines from the PR (take it with a pinch of salt since I'm not actively involved in devs these days) is that we want to enable/disable password rotation at build time & at runtime, but we may miss some edge cases for the runtime change.

Indeed, considering the runtime change of the SLAPD_PASSWORD_MGT_POLICY option, we currently only deal with switching from "no password rotation" to "enabled password rotation". The symetrical use case does not seem to be dealt with.

Again, this is something I noticed from a very quick look at the code. Please bear with me if my findings are incorrect.

So, Pierre is right : if we have a working runtime option to switch back and forth between rotation / no rotation, there's no use having a specific makefile target.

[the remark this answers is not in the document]
fvanderbiest: > But the option will be consumed at runtime

After a call with Marwane, we agreed to say that we do not intend to change the behavior at runtime.
It's build time **only**.

pmauduit (approved):

Issue #3942, opened 2023-04-19
rename geoserver_privileged_user


fvanderbiest:
Datafeeder requires a valid georchestra user to publish to geonetwork:
https://github.com/georchestra/datadir/blob/68a730fb5c6d0bfcdf10592149133fc66ff2bcee/datafeeder/datafeeder.properties#L63

Proposal is to rename `geoserver_privileged_user` to `georchestra_privileged_user` and use it for the datafeeder, so that datafeeder works ootb, including when testadmin is being removed.

fvanderbiest:
Also, we would like to disable cas6 login for this user

fvanderbiest:
> Also, we would like to disable cas6 login for this user

see https://github.com/georchestra/georchestra/issues/3950

PR #3946, opened 2023-04-25
renaming geoserver_privileged_user to georchestra_privileged_user

jeanmi151:
ref: #3942


ldap/README.md
```diff
@@ -37,7 +37,7 @@ The basic users:
  * ```testeditor``` has the USER & GN_EDITOR roles. The password is **testeditor**.
  * ```testadmin``` has the USER, GN_ADMIN, ADMINISTRATOR and MOD_* roles. The password is **testadmin**.
  * ```testdelegatedadmin``` has the USER role. Is able to grant the EXTRACTORAPP & GN_EDITOR roles to members of the psc & c2c orgs. The password is **testdelegatedadmin**.
- * ```geoserver_privileged_user``` is a required user. It is internally used by the geofence module. The default password is ```gerlsSnFd6SmM``` (you should change it, and update the datadir as explained in its [README](https://github.com/georchestra/datadir/blob/18.06/README.md)).
+ * ```georchestra_privileged_user``` is a required user. It is internally used by the geofence module. The default password is ```gerlsSnFd6SmM``` (you should change it, and update the datadir as explained in its [README](https://github.com/georchestra/datadir/blob/18.06/README.md)).
```
fvanderbiest, line 40: "It is internally used by the datafeeder module"

fvanderbiest (changes requested):
LGTM, just remove the ref to geofence.
Thanks !

PR #3961, opened 2023-05-03
[geonetwork] fix add to mapstore template URL issue

### LINKED ISSUE #3958: [geonetwork] configure the externalViewer url template
GN 4.x supports configuring an external viewer, like mapstore, to view the WMS/WFS resources.

On page ```/geonetwork/srv/fre/admin.console#/settings/ui```, section "viewer" > "external viewer" > "viewer url pattern"

Depending on the geOrchestra instances, one out of the 2 following URL templates work:

- pattern that should match [GN](https://github.com/geonetwork/core-geonetwork/blob/main/web-ui/src/main/resources/catalog/js/CatController.js#L758) and [mapstore's documentation](https://docs.mapstore.geosolutionsgroup.com/en/v2021.02.02/developer-guide/map-query-parameters/#add-layers):  ```/mapstore/#/?actions=[{"type":"CATALOG:ADD_LAYERS_FROM_CATALOGS","layers":["${service.name}"],"sources":[{"type":"${service.type}","url":"${service.url}"}]}]```
- pattern actually working for some production instances ```/mapstore/#/?actions=[{"type":"CATALOG:ADD_LAYERS_FROM_CATALOGS","layers":[${service.name}],"sources":[${service.url}]}]```

In the second case, it seems that the `${service.url}` block is producing the text content that the first case outputs with `{"type":"${service.type}","url":"${service.url}"}`. Which is confirmed when looking at the URL we get, if we use the first template on an instance that supports the second.

This seems to be due to the `georchestra` view in geonetwork's configuration. I'm particularly suspecting the following code snippet in the geonetwork/catalog/views/georchestra/module.js file: https://github.com/georchestra/geonetwork/blob/georchestra-gn4.2.x/web-ui/src/main/resources/catalog/views/georchestra/module.js#L132-L134

I've checked on the georchestra instances that work with the first template, they are not loading the module.js from the georchestra view (probably not loaded at all). If asking to load the view by adding `?debug&view=georchestra` in the URL, I'm getting the behaviour associated with the instances that support template 2, which seems to confirm my hypothesis.

Which raises at least 2 questions:
- how come some georchestra instances are using the georchestra view, and some the default view ?
- there seem to be some deprecated and broken code in the geonetwork/catalog/views/georchestra/module.js file. Which parts should be removed ?

jeanpommier:
Should fix #3958


jahow (approved):

fvanderbiest:
Thanks !
submodule updates are taken care of by dependabot.
Closing here.

PR #3972, opened 2023-05-16
password rotation for gateway

marwanehcine:
A PR that add rotation password policy to gateway  (georchestra sub-module) to show alert and force password change for user periodically.


fvanderbiest:
Thanks for sharing this code early.
I'm not sure we want to integrate this as is, but good for testing purposes.

pmauduit (approved):

PR #4011, opened 2023-08-09
Update required steps to add new OAuth2ProviderID field to LDAP

marwanehcine:
An important update to steps required to add  new field OAuth2ProviderId to LDAP where we need to check for correct dn before running final script.

### DIFF
--- a/migrations/master/README.md
+++ b/migrations/master/README.md
@@ -75,7 +75,33 @@ The `oAuth2ProviderId` user attribute was added to the ldap schema.
 
 This attribute will contain OAuth2 user id when user is connected using external identity provider.
 
-To upgrade the ldap, use the following command with the [ldap_migration.ldif](ldap_migration.ldif) file:
+
+To upgrade the ldap, you need first to find the georchestra schema definition using the following command : 
+
+```
+ldapsearch -H ldap://localhost:389 -D cn=admin,dc=georchestra,dc=org -w secret -b cn=schema,cn=config '(cn=*georchestra)' dn
+```
+
+Commands provided in [ldap_migration.ldif](ldap_migration.ldif) assume that required dn is :
+
+dn: cn={7}georchestra,cn=schema,cn=config
+
+If you find a different dn please update Commands provided in [ldap_migration.ldif](ldap_migration.ldif) file
+
+Also, required olcObjectClasses "georchestraUser" to be changed should be localized using the following command : 
+
+```
+ldapsearch -H ldap://localhost:389 -D cn=admin,dc=georchestra,dc=org -w secret -b "cn=schema,cn=config" | grep 1.3.6.1.4.1.53611.1.1.1
+```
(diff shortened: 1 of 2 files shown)


fvanderbiest:
Thanks Marwane. 
Asking for a review from our LDAP expert @pmauduit when he's back.

migrations/master/README.md
```diff
@@ -75,7 +75,33 @@ The `oAuth2ProviderId` user attribute was added to the ldap schema.
 
 This attribute will contain OAuth2 user id when user is connected using external identity provider.
 
-To upgrade the ldap, use the following command with the [ldap_migration.ldif](ldap_migration.ldif) file:
+
+To upgrade the ldap, you need first to localize required dn using the following command :
```
landryb, line 79: s/localize required dn/find the georchestra schema definition/ ?

landryb:
looking at the full ldif now, i'm just puzzled by the two new oauth-related attributes in https://github.com/georchestra/georchestra/blob/4611fc310b62d6f6e99aa659afebec78976ee03c/ldap/docker-root/georchestraSchema.ldif#L21 and https://github.com/georchestra/georchestra/blob/4611fc310b62d6f6e99aa659afebec78976ee03c/ldap/docker-root/georchestraSchema.ldif#L39 - do we need **both** ? they dont have the same def but share the same DESC..

this was added in https://github.com/georchestra/georchestra/commit/f5f65f1f0ac462148364f37ae7382da29f98d3e0#diff-37859fff1ca585f4fc456db553f1203729f538ffbd8910cb94b0cced191f5b8e

pmauduit (approved):
Apart from @landryb 's comment (replacing "to localize" by "find the georchestra schema definition"), it looks good to me.
We are aware that managing OpenLDAP can be cumbersome, as the database number may vary

marwanehcine:
> looking at the full ldif now, i'm just puzzled by the two new oauth-related attributes in
> 
> https://github.com/georchestra/georchestra/blob/4611fc310b62d6f6e99aa659afebec78976ee03c/ldap/docker-root/georchestraSchema.ldif#L22
> 
> and
> https://github.com/georchestra/georchestra/blob/4611fc310b62d6f6e99aa659afebec78976ee03c/ldap/docker-root/georchestraSchema.ldif#L40
> 
> - do we need **both** ? they dont have the same def but share the same DESC..
> and the toplevel class only mentions `oAuth2ProviderId` in MAY line 9.
> this was added in [f5f65f1#diff-37859fff1ca585f4fc456db553f1203729f538ffbd8910cb94b0cced191f5b8e](https://github.com/georchestra/georchestra/commit/f5f65f1f0ac462148364f37ae7382da29f98d3e0#diff-37859fff1ca585f4fc456db553f1203729f538ffbd8910cb94b0cced191f5b8e)

@landryb  "oAuth2ProviderId" is "olcAttributeTypes" and "georchestraOAuth2ProviderId" is "olcObjectClasses".
I have update description of georchestraOAuth2ProviderId to make difference between both.
Thanks

PR #4012, opened 2023-08-13
Inform admin when new OAuth2 account user is created

marwanehcine:
On this PR, when a user account with OAuth2 is created, en entry in journal is added and an email is sent to admin user.

### DIFF
--- a/console/src/main/webapp/manager/app/assets/lang/fr.json
+++ b/console/src/main/webapp/manager/app/assets/lang/fr.json
@@ -263,6 +263,7 @@
   "logs.roleremoved"      : "a été enlevé de l'utilisateur",
   "logs.emailrecoverysent": "Email de récupération du mot de passe envoyé",
   "logs.rolecreated"      : "Role créé",
+  "logs.oauth2usercreated": "Utilisateur OAuth2 créé",
   "logs.noresults"        : "Aucune entrée pour les filtres sélectionnés",
   "analytics.title"       : "Statistiques",
   "analytics.all"         : "Tout le monde",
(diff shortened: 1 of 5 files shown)


console/src/main/webapp/manager/app/assets/lang/fr.json
```diff
@@ -263,6 +263,7 @@
   "logs.roleremoved"      : "a été enlevé de l'utilisateur",
   "logs.emailrecoverysent": "Email de récupération du mot de passe envoyé",
   "logs.rolecreated"      : "Role créé",
+  "logs.oauth2usercreated": "Utilisateur OAuth2 créé",
```
fvanderbiest, line 266: Thanks !
All lang files should be updated by the PR (using google translate if required)

fvanderbiest:
OK, so if I understand correctly, we now log 3 different "events" on user creation: 
* `pendingusercreated` for local user (ldap backend) when "moderatedSignup=true"  
* `usercreated` for local user when "moderatedSignup=false"
* `oauth2usercreated` when user connects with an external identity provider

Is this correct ?

marwanehcine:
> OK, so if I understand correctly, we now log 3 different "events" on user creation:
> 
> * `pendingusercreated` for local user (ldap backend) when "moderatedSignup=true"
> * `usercreated` for local user when "moderatedSignup=false"
> * `oauth2usercreated` when user connects with an external identity provider
> 
> Is this correct ?

Yes, I guess

console/src/main/webapp/manager/app/assets/lang/fr.json
marwanehcine: ok, will be done, thanks

PR #4020, opened 2023-08-31
update procedure notes for new releases and with correction about geonetwork

jeanmi151:
This are modification of what I really do during release patch update to correctly put GeoNetwork tag (for correct compilation).

This is also to not make the same mistakes next release ;)


fvanderbiest (approved):
LGTM !

PR #4022, opened 2023-08-31
Revert "Bump geonetwork from `3b8dae6` to `674bea1`"

fvanderbiest:
Reverts georchestra/georchestra#4021


fvanderbiest:
There's an issue here - master shouldn't have this change

jeanmi151:
the master branch of georchestra is set with on georchestra-gn4.2.X (as specified here : https://github.com/georchestra/georchestra/blob/master/.gitmodules#L4)
so for me it is "normal" with get this PR

PR #4025, opened 2023-09-03
Adding Header authentication to Gateway (Georchestra) 

marwanehcine:
On this PR, we add update header allow to authenticate user using headers fields received from Apache.
logout link is removed is Apache header authentication is applied.

### DIFF
--- a/header/src/main/webapp/WEB-INF/jsp/index.jsp
+++ b/header/src/main/webapp/WEB-INF/jsp/index.jsp
@@ -98,6 +98,8 @@ Boolean analyticsadmin = false;
 Boolean msadmin = false;
 String sec_roles = request.getHeader("sec-roles");
 String sec_remaining_days=request.getHeader("sec-ldap-remaining-days");
+String sec_organisation=request.getHeader("sec-org");
+
 
 if(sec_roles != null) {
     String[] roles = sec_roles.split(";");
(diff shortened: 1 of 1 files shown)


header/src/main/webapp/WEB-INF/jsp/index.jsp
```diff
@@ -402,7 +404,11 @@ if(sec_roles != null) {
                        <c:if test='<%= sec_remaining_days != null && !sec_remaining_days.equals("") %>'>
                             <span id="ldap_user_alert" style="display:none;" ><fmt:message key="remaining_days_msg_part1"/> <%= sec_remaining_days %> <fmt:message key="remaining_days_msg_part2"/> <a href="${consolePublicContextPath}/account/changePassword"><fmt:message key="remaining_days_msg_part3"/> </a></span>
                        </c:if>
-                    <a href="${consolePublicContextPath}/account/userdetails"><%=request.getHeader("sec-username") %></a><span class="light"> | </span><a href="/logout" onclick="logout()"><fmt:message key="logout"/></a>
+                        <a href="${consolePublicContextPath}/account/userdetails"><%=request.getHeader("sec-username") %></a>
+                       <c:if test='<%= sec_organisation != null && !sec_organisation.equals("INRAE") %>'>
```
pmauduit, line 408: This ("INRAE") is too specific to land into the main project, maybe we should have a customized header and leave it here as it is. Anyway, as far as I know this code ("header webapp") should disappear with a header revamp.

pmauduit (changes requested):

fvanderbiest:
I agree with @pmauduit - closing.

header/src/main/webapp/WEB-INF/jsp/index.jsp
marwanehcine: @pmauduit , header is not removed here since it is updated after first header revamp!
I believe we can use this code :
<c:if test='<%= is_apache_autehticated %>'
where is_apache_autehticated is computed as spring app and we  can put "INREA" string dans prop file pour que ça soit generique.

pmauduit:
After having discussed with @marwanehcine we might need a special treatment in order to be able to logout the users who would be pre-authenticated by something in front of the gateway and/or SP, but this could be handled later on.

Issue #4027, opened 2023-09-05
upgrade to debian bookworm


pmauduit:
Ansible playbook is affected, and @landryb already began a WIP here:
https://github.com/georchestra/ansible/pull/125

Docker-side, we are still referencing / using oldstable / bullseye as a base, mainly for the openldap image, but git grepping also indicates that the documentation should be updated

pmauduit:
I have a very weird behaviour with the new debian version of openldap (slapd 2.5.13+dfsg-5), I can't use a filter on the testadmin email, yet it is in the ldap tree:

```
# ldapsearch -x -bou=users,dc=georchestra,dc=org mail | grep '^mail: '
mail: psc+testuser@georchestra.org
mail: psc+testadmin@georchestra.org
mail: psc+testeditor@georchestra.org
mail: psc+idatafeeder@georchestra.org
mail: psc+testreviewer@georchestra.org
mail: psc+testdelegatedadmin@georchestra.org
mail: psc+geoserver_privileged_user@georchestra.org

# ldapsearch -x -bou=users,dc=georchestra,dc=org 'mail=psc+testeditor@georchestra.org'
[...]
dn: uid=testeditor,ou=users,dc=georchestra,dc=org
objectClass: georchestraUser
objectClass: inetOrgPerson
objectClass: organizationalPerson
objectClass: person
objectClass: shadowAccount
objectClass: top
uid: testeditor
sn: EDITOR
description: editor
mail: psc+testeditor@georchestra.org
cn: testeditor
georchestraObjectIdentifier: e75b336d-a9fc-4671-8002-e1a49c2acf2c
[...]

# ldapsearch -x -bou=users,dc=georchestra,dc=org 'mail=psc+testadmin@georchestra.org'
# extended LDIF
#
# LDAPv3
# base <ou=users,dc=georchestra,dc=org> with scope subtree
# filter: mail=psc+testadmin@georchestra.org
# requesting: ALL
#

# search result
search: 2
result: 0 Success

# numResponses: 1

```

I only encounter the issue with the testadmin user, no matter I'm using the cn=admin account to query the LDAP or if I do the search anonymously. it works with e.g. testeditor default e-mail.

pmauduit:
It works back after having used the `su -c openldap slapindex` command, I have now idea how I did to get my indexes corrupted.

PR #4028, opened 2023-09-06
[ldap,console] Adding mail attribute on organizations

pmauduit:
This PR allows to add a "mail" attribute on the organizations into the LDAP (more specifically on the "orgExt" objects). LDAP-wise, it means adding an objectClass `extensibleObject` so that the `mail` attribute becomes possible onto the `organization`s entries from the `ou=orgs` branch.

It also updates the LDAP base docker image to debian bookworm (see first commit for the details), this at least to partially address #4027.

Tests:
* testsuite on modified components looks ok
* runtime tested on a minimalist docker composition along with a console, a LDAP and a PostGreSQL db, no issue encountered so far using the console admin ui or the "create account" feature.


pmauduit:
Note: I had to patch the `security-proxy-spring-integration` module in order to have it built, else getting the following error:

```
[INFO] -------------------------------------------------------------
[ERROR] COMPILATION ERROR : 
[INFO] -------------------------------------------------------------
[ERROR] /...georchestra/security-proxy-spring-integration/src/main/java/org/georchestra/security/client/console/GeorchestraConsoleUsersApiImpl.java:[35,8] org.georchestra.security.client.console.GeorchestraConsoleUsersApiImpl is not abstract and does not override abstract method findByOAuth2ProviderId(java.lang.String) in org.georchestra.security.api.UsersApi
```

It should be related to the ongoing work on oauth2 support in the gateway. I just implemented the missing methods in the class as follows:

```
diff --git a/security-proxy-spring-integration/src/main/java/org/georchestra/security/client/console/GeorchestraConsoleUsersApiImpl.java b/security-proxy-spring-integration/src/main/java/org/georchestra/security/client/console/GeorchestraConsoleUsersApiImpl.java
index d4155efd6..4e55862fd 100644
--- a/security-proxy-spring-integration/src/main/java/org/georchestra/security/client/console/GeorchestraConsoleUsersApiImpl.java
+++ b/security-proxy-spring-integration/src/main/java/org/georchestra/security/client/console/GeorchestraConsoleUsersApiImpl.java
@@ -57,4 +57,14 @@ public class GeorchestraConsoleUsersApiImpl implements UsersApi {
         return client.get("/console/internal/users/username/{name}", GeorchestraUser.class, username);
     }
 
+    @Override
+    public Optional<GeorchestraUser> findByEmail(String username) {
+        return Optional.empty();
+    }
+
+    @Override
+    public Optional<GeorchestraUser> findByOAuth2ProviderId(String oauth2ProviderId) {
+        return Optional.empty();
+    }
+
 }
```
the master branch is currently broken because of these two missing methods, ping @emmdurin / @marwanehcine

pmauduit:
Also this PR lacks some migration notes, as every `organization`s objects from the `ou=orgs` ldap branch will require to be patched, in order to add the `objectclass: extensibleObject`.

pmauduit:
Also wondering if this PR deserves its own GIP, but since this was not needed to add the `lastLogin` date in the console web ui (#3889), and the modification here sounds pretty similar, I don't think it is actually necessary.

console/src/main/java/org/georchestra/console/ws/utils/LogUtils.java
```diff
@@ -257,6 +257,10 @@ public void logOrgChanged(Org org, JSONObject json) throws IOException {
         if (!org.getNote().equals(json.optString(Org.JSON_NOTE))) {
             createAndLogDetails(id, Org.JSON_NOTE, org.getNote(), json.optString(Org.JSON_NOTE), type);
         }
+        // mail addres of org changed
+        if (!org.getMail().equals(json.optString(Org.JSON_MAIL))) {
```
pmauduit, line 261: I haven't tested if this could lead to a NPE

pmauduit:
Added some documentation migration in the last commit, I think this PR is now ready.

There will still be some work to take advantage of the mail into GeoNetwork, but could be done in another one.

landryb:
> the master branch is currently broken because of these two missing methods, ping @emmdurin / @marwanehcine

i had seen the builds from master being broken on the buildbot but hadnt looked into it, that might be exactly that.

pmauduit:
There is still mising some code for the next feature related to the geonetwork organization sync, but I could not find a way to get this PR back to draft.

console/src/main/java/org/georchestra/console/ws/editorgdetails/EditOrgDetailsFormController.java
```diff
@@ -108,6 +110,7 @@ public String edit(Model model, @ModelAttribute EditOrgDetailsFormBean formBean,
         validation.validateOrgField("url", formBean.getUrl(), resultErrors);
         validation.validateOrgField("address", formBean.getAddress(), resultErrors);
         validation.validateOrgField("description", formBean.getDescription(), resultErrors);
+        // TODO validate mail address for the organization ?
```
fvanderbiest, line 113: which kind of validations are performed on the other fields here ? Would it be relevant for the email too ?

ldap-account-management/src/main/java/org/georchestra/ds/orgs/OrgExt.java
```diff
@@ -35,71 +35,16 @@
...
-
-    public void setNote(String note) {
-        this.note = note;
-    }
+    private @Getter @Setter String address = "";
+    private @Getter @Setter String description = "";
+    private @Getter @Setter String url = "";
+    private @Getter @Setter String logo = "";
+    private @Getter @Setter String note = "";
+    private @Getter @Setter String mail = "";
```
f-necas, line 47: It may be less explicit but more clear if getter/setter annotations to the class level `@Getter @Setter class OrgExt` ? What do you think ?

console/src/main/java/org/georchestra/console/ws/utils/LogUtils.java
f-necas: I suppose it won't as it is defined by default to empty string in `OrgExt.java`. Maybe it should be a good thing to switch variables in order to be sure to avoid it ? `optString` returns an empty string if not found so it should never lead to an NPE.

f-necas (approved):
Nice work ! 

I confirm the compilation error you mentioned. Instead of this everything LGTM and works.
We may discuss about the optional/debatable lombok and NPE things. But it's not blocking for me.

f-necas (approved):

pmauduit:
rebased on master

ldap-account-management/src/main/java/org/georchestra/ds/orgs/OrgExt.java
pmauduit: maybe just `@Data` then, if there are no side-effects (https://projectlombok.org/features/Data)

pmauduit: actually, having @getter/@setter at the class level sounds good, will do.

PR #4032, opened 2023-09-11
console - customize the welcome message if moderatedSignup=false

pmauduit:
No need to mention an unecessary moderation of the account, one can log in just after having validated the form.

Tests:

* testsuite OK
* runtime OK

image

Note: I first tried with a login link as shown on the screenshot above, but it should already be in the header, so it would have been a bit redundant. I am not an expert in regards to these user experience matters.


fvanderbiest:
Rather than a big "Request submitted" message, would it be possible to display "Account created" ?
Thanks !

pmauduit:
> Rather than a big "Request submitted" message, would it be possible to display "Account created" ? Thanks !

Not the same message identifier, so it requires to add an extra variable

fvanderbiest:
Missing translations fixed with d6fe8c142

Issue #4029, opened 2023-09-07
builds from master broken since #4005


landryb:
extracted from https://github.com/georchestra/georchestra/pull/4028#issuecomment-1708292492 and also experienced on the community buildbot:
```
[ERROR] /var/lib/buildbot/workers/georchestra/buildjob-master/build/security-proxy-spring-integration/src/main/java/org/georchestra/security/client/console/GeorchestraConsoleUsersApiImpl.java:[35,8] org.georchestra.security.client.console.GeorchestraConsoleUsersApiImpl is not abstract and does not override abstract method findByOAuth2ProviderId(java.lang.String) in org.georchestra.security.api.UsersApi
```

for the buildbot, the build failure happens since https://github.com/georchestra/georchestra/commit/f5f65f1f0ac462148364f37ae7382da29f98d3e0 from #4005

landryb:
while #4037 is still open, the build issue itself was fixed by #4042

PR #4037, opened 2023-09-19
commons - attempt to fix build (#4029)

pmauduit:
We have 2 implementations of the `UsersApi` interface, which has recently been modified for the Oauth2 support:

* https://github.com/georchestra/georchestra/blob/master/ldap-account-management/src/main/java/org/georchestra/ds/security/UsersApiImpl.java
* https://github.com/georchestra/georchestra/blob/master/security-proxy-spring-integration/src/main/java/org/georchestra/security/client/console/GeorchestraConsoleUsersApiImpl.java

Only one contains an implementation for the `findByEmail` & `findByOAuth2ProviderId` methods. This PR suggests to have a default implementation for both missing methods on the interface level.

Tests:

compilation using the following:

```
% mvn clean package -P-all,console 
[...]
[INFO] Executed tasks
[INFO] ------------------------------------------------------------------------
[INFO] Reactor Summary for Root project of the geOrchestra SDI 23.1-SNAPSHOT:
[INFO] 
[INFO] Root project of the geOrchestra SDI ................ SUCCESS [  5.274 s]
[INFO] ogc-server-statistics .............................. SUCCESS [  7.641 s]
[INFO] geOrchestra Commons ................................ SUCCESS [  5.347 s]
[INFO] security-proxy-spring-integration .................. SUCCESS [  5.401 s]
[INFO] www.testcontainers.org containers for some geOrchestra dockerized apps SUCCESS [  3.095 s]
[INFO] LDAP users and groups management library ........... SUCCESS [  7.024 s]
[INFO] Console ............................................ SUCCESS [ 34.129 s]
[INFO] ------------------------------------------------------------------------
[INFO] BUILD SUCCESS
[INFO] ------------------------------------------------------------------------
[INFO] Total time:  01:08 min
[INFO] Finished at: 2023-09-19T11:09:59+02:00
[INFO] ------------------------------------------------------------------------

```

### DIFF
--- a/commons/src/main/java/org/georchestra/security/api/UsersApi.java
+++ b/commons/src/main/java/org/georchestra/security/api/UsersApi.java
@@ -18,10 +18,10 @@
  */
 package org.georchestra.security.api;
 
+import org.georchestra.security.model.GeorchestraUser;
+
 import java.util.List;
 import java.util.Optional;
-
-import org.georchestra.security.model.GeorchestraUser;
 
 public interface UsersApi {
 
@@ -35,9 +35,13 @@ public interface UsersApi {
      */
     Optional<GeorchestraUser> findByUsername(String username);
 
-    Optional<GeorchestraUser> findByEmail(String username);
+    default Optional<GeorchestraUser> findByEmail(String username) {
+        return Optional.empty();
+    };
 
-    Optional<GeorchestraUser> findByOAuth2ProviderId(String oauth2ProviderId);
+    default Optional<GeorchestraUser> findByOAuth2ProviderId(String oauth2ProviderId) {
+        return Optional.empty();
+    };
 
     List<GeorchestraUser> findAll();
 }


f-necas (commented):
Thanks Pierre for this working fix. 

As we don't have a lot of implementations, I suggest to keep the interface simple and to move changes in `GeorchestraConsoleUsersApiImpl.java` with something like : 

```java
    @Override
    public Optional<GeorchestraUser> findByEmail(String username) {
        return Optional.empty();
    }

    @Override
    public Optional<GeorchestraUser> findByOAuth2ProviderId(String oauth2ProviderId) {
        return Optional.empty();
    }
```
In my opinion, we should avoid to implement behavior in interfaces.
What do you think ?

pmauduit:
Let's go for having it onto the implementation instead, I'll rework the PR.

marwanehcine:
> > In my opinion, we should avoid to implement behavior in interfaces.
> > What do you think ?
> 
> This is what I suggested in the other PR before making it default onto the interface. I don't have that many pros/cons for one solution or the other, to be honest. It is true that it does not make that much sense to have implementation code onto the interface, but they are just default methods, and since the Java language allows it, it sounded to me like the right usecase for it.

I 'm with adding implementation into classes and not interface, because define default behavior means that we need it by default. I would not prefer to make return "Optional.empty();"  as default behavior.
I'm OK with current solution.

pmauduit:
Also (not related to the issue here but since we are on it):
https://github.com/georchestra/georchestra/blob/master/console/src/main/webapp/WEB-INF/spring/webmvc-config.xml#L331-L333

I'm concerned with both imports. It should work in a jetty context as PWD will be jetty home were webapps/console will exist, but not necessarily if deployed into another servlet container like tomcat on a more classical deployment.

f-necas (approved):
Thanks @pmauduit, everything LGTM !

pmauduit:
> Thanks @pmauduit, everything LGTM !

thanks for reviewing !

PR #4045, opened 2023-09-27
Fix email sending issue for multiple recipients

marwanehcine:
On  this PR, mail sending issue for multiple recipients is fixed.

### DIFF
--- a/console/src/main/java/org/georchestra/console/mailservice/Email.java
+++ b/console/src/main/java/org/georchestra/console/mailservice/Email.java
@@ -151,11 +151,15 @@ public class Email {
         if (isValidEmailAddress(from)) {
             message.setFrom(new InternetAddress(from));
         }
+        InternetAddress[] recipientAddress = new InternetAddress[recipients.size()];
+        int counter = 0;
         for (String recipient : recipients) {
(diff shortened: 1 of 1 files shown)


fvanderbiest:
Thanks Marwane. 
It used to work on previous geOrchestra versions - is it a regression ?

marwanehcine:
> Thanks Marwane. It used to work on previous geOrchestra versions - is it a regression ? Shall we backport it ?

I'm not sure if it is a regression, but It was not working when there is more than one receiver, for only one receiver it is working, problem appears when more than one receivers.

pmauduit (approved):
LGTM but I was wondering if there was not a way to write the loop using streams instead, in a more java8-ish approach.

pmauduit:
> I'm not sure if it is a regression, but It was not working when there is more than one receiver, for only one receiver it is working, problem appears when more than one receivers.

This is strange, I am pretty sure we were able to send mails to several recipients in the past. Looking at the `javax.mail.internet.MimeMessage` implementation, several `To: ` mail headers should have been appended.

fvanderbiest:
> This is strange, I am pretty sure we were able to send mails to several recipients in the past

100% agree

PR #4046, opened 2023-09-28
making Rabbitmq dependency optional

marwanehcine:
making Rabbitmq dependency optional on console

### DIFF
--- a/console/pom.xml
+++ b/console/pom.xml
@@ -176,6 +176,10 @@
     </dependency>
     <!-- Spring dependencies -->
     <dependency>
+      <groupId>org.springframework.boot</groupId>
+      <artifactId>spring-boot-autoconfigure</artifactId>
+    </dependency>
+    <dependency>
       <groupId>org.springframework</groupId>
       <artifactId>spring-core</artifactId>
     </dependency>
--- a/console/src/main/java/org/georchestra/console/events/RabbitmqEventsAutoConfiguration.java
+++ b/console/src/main/java/org/georchestra/console/events/RabbitmqEventsAutoConfiguration.java
@@ -1,0 +1,20 @@
+package org.georchestra.console.events;
+
+import org.springframework.boot.autoconfigure.condition.ConditionalOnBean;
+import org.springframework.boot.autoconfigure.condition.ConditionalOnExpression;
+import org.springframework.boot.autoconfigure.condition.ConditionalOnProperty;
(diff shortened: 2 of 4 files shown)


fvanderbiest:
Can you provide a PR for https://github.com/georchestra/datadir/blob/master/default.properties too ? Thx !

console/pom.xml
```diff
@@ -175,6 +175,10 @@
       <artifactId>commons-validator</artifactId>
     </dependency>
     <!-- Spring dependencies -->
+    <dependency>
+      <groupId>org.springframework.boot</groupId>
+      <artifactId>spring-boot-autoconfigure</artifactId>
+    </dependency>
```
pmauduit, line 181: I'm concerned by mixing spring boot dependencies on a classic spring project

console/src/main/java/org/georchestra/console/events/RabbitmqEventsAutoConfiguration.java
```diff
@@ -0,0 +1,20 @@
...
+import org.springframework.boot.autoconfigure.condition.ConditionalOnProperty;
+import org.springframework.context.ApplicationContext;
+import org.springframework.context.annotation.*;
+import org.springframework.context.support.ClassPathXmlApplicationContext;
+
+import javax.annotation.PostConstruct;
+
+@Configuration
+@ImportResource({ "file:webapps/console/WEB-INF/spring/rabbit-listener-context.xml",
+        "file:webapps/console/WEB-INF/spring/rabbit-sender-context.xml" })
```
pmauduit, line 14: Same remark as https://github.com/georchestra/georchestra/pull/4037#issuecomment-1728247260 : I would move the xml files into the classpath and resolve them using the classpath:// scheme, because there is no guarantee that the servlet container will run in a specific directory where a `webapps/` subdirectory will be available.

pmauduit (changes requested):

console/pom.xml
pmauduit: > I'm concerned by mixing spring boot dependencies on a classic spring project

@groldan is it safe to do so ?

pmauduit: confirmed by @groldan , we can't mix spring boot dependencies into regular spring applications, so we will have to rework this.

marwanehcine: OK, but why it is not possible ?

PR #4047, opened 2023-10-04
Making rabbitmq dependency optional on console

marwanehcine:
On this PR, we make  rabbitmq dependency optional on console


pmauduit (commented):
Thanks Marwanne, it indeed looks more relevant than the previous one without mixing spring-boot components.

I am wondering what a ApplicationContextInitializer provides though, why not just sourcing the /etc/georchestra/default.properties directly from the xml file ?

marwanehcine:
> Thanks Marwanne, it indeed looks more relevant than the previous one without mixing spring-boot components.
> 
> I am wondering what a ApplicationContextInitializer provides though, why not just sourcing the /etc/georchestra/default.properties directly from the xml file ?

The problem here is that <import>  in xml conf file for spring 3 are called before reading props file, so the only solution is to define listerner on web.xml to source default.props before <import>.  this is the only way it works!

console/src/main/webapp/WEB-INF/web.xml
```diff
@@ -14,6 +14,10 @@
         <param-name>org.eclipse.jetty.servlet.Default.dirAllowed</param-name>
         <param-value>false</param-value>
     </context-param>
+    <context-param>
+        <param-name>contextInitializerClasses</param-name>
+        <param-value>org.georchestra.console.events.DefaultPropsInitializer</param-value>
+    </context-param>
```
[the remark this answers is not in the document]
marwanehcine: Hello, good question, issue here is that "<import>" inside xml conf file (beans definition) is called too early so props files a not accessible yet, however using context-param, we load default.properties file even before call to <import> on this case.
I have checked other cases and it did not work only after user of context-param inside web.xml file

console/src/main/webapp/WEB-INF/spring/webmvc-config.xml
```diff
@@ -324,12 +324,5 @@
...
 
-  <bean id="rabbitmqEventsSender" class="org.georchestra.console.events.RabbitmqEventsSender" depends-on="eventTemplate">
-    <property name="eventTemplate" ref="eventTemplate"/>
-  </bean>
-
-  <import resource="file:webapps/console/WEB-INF/spring/rabbit-listener-context.xml"/>
-
-  <import resource="file:webapps/console/WEB-INF/spring/rabbit-sender-context.xml"/>
-
+  <import resource="classpath:spring/rabbit-context-${enableRabbitmqEvents}.xml"/>
```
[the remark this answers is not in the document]
marwanehcine: code updated

console/src/main/java/org/georchestra/console/events/DefaultPropsInitializer.java
```diff
@@ -0,0 +1,22 @@
...
+
+import java.io.IOException;
+
+public class DefaultPropsInitializer implements ApplicationContextInitializer<ConfigurableApplicationContext> {
+
+    @Override
+    public void initialize(ConfigurableApplicationContext applicationContext) {
+        try {
+            applicationContext.getEnvironment().getPropertySources().addFirst(
+                    new ResourcePropertySource(new FileSystemResource("/etc/georchestra/default.properties")));
```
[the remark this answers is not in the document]
marwanehcine: code updated!

console/src/main/java/org/georchestra/console/events/DefaultPropsInitializer.java
```diff
@@ -0,0 +1,27 @@
...
+
+    @Override
+    public void initialize(ConfigurableApplicationContext applicationContext) {
+        try {
+            applicationContext.getEnvironment().getPropertySources()
+                    .addFirst(new ResourcePropertySource(new FileSystemResource(
+                            System.getProperty("georchestra.datadir", "etc/georchestra") + "/default.properties")));
+        } catch (IOException e) {
+            LOGGER.debug("Cannot find default props file");
+            LOGGER.error(e.getStackTrace());
```
pmauduit, line 23: maybe LOGGER.error for the first one as well ?

pmauduit: or even `LOGGER.error("Cannot find ...", e);`

console/src/main/webapp/WEB-INF/spring/webmvc-config.xml
```diff
@@ -324,12 +324,5 @@
...
 
-  <bean id="rabbitmqEventsSender" class="org.georchestra.console.events.RabbitmqEventsSender" depends-on="eventTemplate">
-    <property name="eventTemplate" ref="eventTemplate"/>
-  </bean>
-
-  <import resource="file:webapps/console/WEB-INF/spring/rabbit-listener-context.xml"/>
-
-  <import resource="file:webapps/console/WEB-INF/spring/rabbit-sender-context.xml"/>
-
+  <import resource="classpath:spring/rabbit-context-${enableRabbitmqEvents:true}.xml"/>
```
pmauduit, line 327: false by default ?

pmauduit (changes requested):

console/src/main/webapp/WEB-INF/spring/webmvc-config.xml
marwanehcine: done.

console/src/main/java/org/georchestra/console/events/DefaultPropsInitializer.java
pmauduit: you can still put the exception as argument from the Logger.error call I'd expect.

Issue #4030, opened 2023-09-07
Add ability to launch custom scripts before the start of the app


edevosc2c:
# Why

It is becoming a major thing that a lot of clients will ask us to have small modifications over the existing georchestra components.

For instance, Datagrandest mostly have only customized images for their needs.

The downside of maintaining custom images is that it adds maintenance cost, (sometimes confusion) and a big lag with the upstream/official image.

Creating a whole new docker image is not always needed, especially for small modifications.

# Proposed solution

At the geonetwork-ui project, they introduced the ability to execute custom scripts: https://github.com/geonetwork/geonetwork-ui/pull/507

So I propose to have a similar system in place for every docker image that we manage. I'm not entirely sure how to do it, but the ability to specify the directory where all the custom scripts will be executed is a good idea.
This would allow specifying these scripts inside the datadir for example.

Having this functionality will avoid doing "hacky" stuff, especially in the helm chart of georchestra for customizing the container, example:
- https://github.com/georchestra/helm-georchestra/pull/53
- https://github.com/georchestra/helm-georchestra/blob/main/templates/cas/cas-deployment.yaml#L30
- https://github.com/georchestra/helm-georchestra/blob/7bf2c518b64791f44bd9e22bde8848e843081b4c/templates/geonetwork/geonetwork-deployment.yaml#L41
- https://github.com/georchestra/helm-georchestra/blob/7bf2c518b64791f44bd9e22bde8848e843081b4c/templates/geoserver/geoserver-deployment.yaml#L44

# Additional comment

geoserver docker image seems to already support this "script" thing: https://github.com/georchestra/georchestra/tree/master/geoserver/webapp/src/docker/docker-entrypoint.d

On top of that this would fit the need explained by @jeanpommier here: https://github.com/georchestra/helm-georchestra/issues/46#issuecomment-1695733879

--------------

Ping @jeanmi151 @jeanpommier @pmauduit @fvanderbiest

jeanmi151:
Okay after discussion with @edevosc2c , We think something like it is done for geonetwork-ui (datahub) would be a great thing : 
https://github.com/geonetwork/geonetwork-ui/pull/507/files#diff-964be1ec80820060f908f3aec54f46def8145c723a5b8a5bb44b9ec69c76d0c9

Whats we are thinking is : 
1. have a script with a env var (link to a path) that would activate (copy or sylmbolic link) customize webapp  before running the service
2. have a script with a env var (link to a path) that would execute all insider script contained in it
[...]

PR #4065, opened 2023-10-27
Implemention of new WebComponent header

tkohr:
# New geOrchestra header

The goal of this PR is to integrate the [new web component header](https://github.com/georchestra/header/) into georchestra.

This new header have the capability to display the original `iframe` with the `legacy-header` attribute (See How-to section below).

## Implementations

Following apps/repos have to be modified to be able to use it :
- [x] analytics: wc header shows up, needs white background,
- [x] console: wc does not show up yet => why?
- [x] security-proxy: where should this one show up?
- [x] datahub: todo: pass `headerUrl` from default.properties to add-header-webcomponent.sh script
- [x] datafeeder: pass `headerUrl` from default.properties to add-header-webcomponent.sh script

Submodules/other repos:
- [x] geonetwork: js error from other dep `lib.js` => moving wc script behind it does not help
- [x] geoserver
- [x] mapstore
- [x] cas
- [x] gateway (also do not forget password rotation message implemented in legacy header for MEL, but this is project specific)
- [x] home header

### Validate PRs :

- New Header https://github.com/georchestra/header/pull/11
- Geoserver https://github.com/georchestra/geoserver/pull/32
- Geonetwork https://github.com/georchestra/geonetwork/pull/257
- Mapstore https://github.com/georchestra/mapstore2-georchestra/pull/663
- CAS https://github.com/georchestra/georchestra-cas-server/pull/15
- Georchestra https://github.com/georchestra/georchestra/pull/4065
- Datafeeder https://github.com/georchestra/georchestra/issues/4065 | Needs validation of another PR (See #4030, PR https://github.com/georchestra/georchestra/pull/4083)

Docker : 
- HtDocs https://github.com/georchestra/htdocs/pull/3
- Datadir https://github.com/georchestra/datadir/pull/350
- Docker https://github.com/georchestra/docker/pull/242

### Other issues:
- [x] error in analytics js when using header.js from [mobile](https://github.com/georchestra/header/pull/11) branch (includes white background for analytics)
- [ ] Documentation

### Issues Done

- [x] new admin header is higher (108px) than non-admin (80px), mobile even shorter (64px) => unify height?
- [x] `headerUrl` needs to be defined in `default.properties`: empty for new header, `/header/` for legacy => acceptable?
- [x] handle highlighted active tabs to work on new and legacy
- [x] Handle i18n
- [x] Handle font-size (mostly avoid using rem unit) 
- [x] Implement custom script at startup in datafeeder to be able to add new header 

## How-to

Iframe can still be set with defining `legacy-url` attribute, style can also be set with `legacy-style` attribute :
  ```html
  <geor-header legacy-url="myheader.com" legacy-style="width: 100%"></geor-header>
   <script src="https://cdn.jsdelivr.net/gh/georchestra/header@dist/header.js"></script>
  ```

Attributes available :

| Attribute     | Description                                                                                          | Example                                                                     | For host | For legacy |
|---------------|------------------------------------------------------------------------------------------------------|-----------------------------------------------------------------------------|----------|------------|
| lang          | Used to force header language (default value : en)                                                   | `<geor-header lang='de'>`                                                   | v        |            |
| active-app    | Use this attribute to set the active class in menu                                                   | `<geor-header active-app='console'>`                                        | v        | v          |
| logo-url      | Use this attribute to set the logo for the new header (not legacy one).                              | `<geor-header logo-url='https://linktomylogo.com'>`                         | v        |            |
| legacy-header | Use this attribute to enable the legacy header `iframe` tag. Needs `legacy-url`.                     | `<geor-header legacy-header='true' legacy-url="/header/">`                  |          | v          |
| legacy-url    | Legacy URL: if set, activates iframe with src attribute pointing to this URL. Needs `legacy-header`. | `<geor-header legacy-header='true' legacy-url="/header/"></geor-header>`    |          | v          |
| style         | adds this style to iframe or host tag (if legacy url is not used)                                    | `<geor-header legacy-url="myheader.com" style="width: 100%"></geor-header>` | v        | v          |

### DIFF
--- a/console/src/main/webapp/manager/app/assets/index.html
+++ b/console/src/main/webapp/manager/app/assets/index.html
@@ -14,12 +14,12 @@
 
 <body ng-app="manager" ng-strict-di ng-controller="AppController">
 
-  <geor-header style="width:100%;height:90px" active-app="console"></geor-header>
-  <script src="http://localhost:5500/header.js"></script>
+  <geor-header legacy-header="{{ platformInfos.useLegacyHeader }}" legacy-url="{{ platformInfos.headerUrl }}" style="width:100%;height:{{ platformInfos.headerHeight }}px;border:none;" active-app="console"></geor-header>
+  <script ng-src="{{ trustSrc(platformInfos.headerScript) }}"></script>
 
   <div flash-message="2000"></div>
 
-  <div class="container">
+  <div class="container" style="padding-top:20px">
 
     <div class="profile label label-{{(profile === 'SUPERUSER') ? 'info' : 'warning'}} pull-right" title="{{'index.' + profile + 'tip' | translate }}">
       {{ 'index.' + profile | translate}}
--- a/console/src/main/webapp/manager/app/assets/index.html
+++ b/console/src/main/webapp/manager/app/assets/index.html
@@ -14,7 +14,8 @@
 
 <body ng-app="manager" ng-strict-di ng-controller="AppController">
 
-  <iframe src="/header/?active=console" scrolling="no" frameborder="0"></iframe>
+  <geor-header style="width:100%;height:90px" active-app="console"></geor-header>
+  <script src="http://localhost:5500/header.js"></script>
 
   <div flash-message="2000"></div>
 
--- a/analytics/src/main/webapp/WEB-INF/jsp/header.jsp
+++ b/analytics/src/main/webapp/WEB-INF/jsp/header.jsp
@@ -3,7 +3,7 @@
 <c:choose>
     <c:when test='<%= request.getParameter("noheader") == null %>'>
     <div id="go_head">
-        <geor-header legacy-url="${headerUrl}" legacy-style="width:100%;height:${headerHeight}px;border:none;overflow:hidden;"></geor-header>
+        <geor-header legacy-header="${useLegacyHeader}" legacy-url="${headerUrl}" style="width:100%;height:${headerHeight}px;border:none;" active-app="analytics"></geor-header>
         <script src="${headerScript}"></script>
     </div>
     </c:when>
--- a/analytics/src/main/webapp/WEB-INF/jsp/header.jsp
+++ b/analytics/src/main/webapp/WEB-INF/jsp/header.jsp
@@ -3,7 +3,8 @@
 <c:choose>
     <c:when test='<%= request.getParameter("noheader") == null %>'>
     <div id="go_head">
-        <iframe src="${headerUrl}?lang=<%= lang %>" style="width:100%;height:${headerHeight}px;border:none;overflow:hidden;" scrolling="no" frameborder="0"></iframe>
+        <geor-header legacy_url="${headerUrl}" legacy_style="width:100%;height:${headerHeight}px;border:none;overflow:hidden;"></geor-header>
+        <script src="${headerScript}"></script>
     </div>
     </c:when>
 </c:choose>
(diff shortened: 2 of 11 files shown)


fvanderbiest:
> `headerUrl` needs to be defined in `default.properties`: empty for new header, `/header/` for legacy => acceptable?

Sounds good

fvanderbiest:
> * security-proxy: where should this one show up?

It seems we have a header.jsp defined in the SP here: https://github.com/georchestra/georchestra/blob/master/security-proxy/src/main/webapp/header.jsp

It is integrated in error pages like https://github.com/georchestra/georchestra/blob/4609670eb5a615f842c0deb04194eac3d403dcd6/security-proxy/src/main/webapp/403.jsp#L17

jahow:
> I've hit a problem with the console.
> 
> Font-size is set in `rem`, and the new header doesn't shows up the same way in console and in analytics. (font-size set to 10px in console's html tag)
> 
> Do we set a fixed font-size in the new header ? @tkohr @fvanderbiest

Definitely. I struggle to see any benefit in `rem`.

edevosc2c:
> > `headerUrl` needs to be defined in `default.properties`: empty for new header, `/header/` for legacy => acceptable?
> 
> Sounds good

After discussion with florian. I was proposing for activating the new header either:
- `headerUrl` to `false`
- a new variable like `headerWebComponent` to `true`

Because `headerUrl` to empty doesn't actual mean anything, on the DevOps side (user configuring) it sounds like the variable was mistakenly left empty.

jahow:
> a new variable like headerWebComponent to true

I would suggest not mentioning a technical concern inside a configuration parameter. Also I would suggest doing it the other way around, and introduce a parameter for enabling the "legacy" header (iframe) like `useLegacyHeader = true`; then once a platform moves to the new header the parameter will disappear.

My 2 cents while passing by :slightly_smiling_face:

edevosc2c:
> > a new variable like headerWebComponent to true
> 
> I would suggest not mentioning a technical concern inside a configuration parameter. Also I would suggest doing it the other way around, and introduce a parameter for enabling the "legacy" header (iframe) like `useLegacyHeader = true`; then once a platform moves to the new header the parameter will disappear.
> 
> My 2 cents while passing by 🙂

Good idea but we can also do the same for the proposed parameter like `headerWebComponent`, once the web component is the default and the legacy header is removed then we can remove the parameter `headerWebComponent` too.

Still in both case, what is the behavior of `headerUrl` when either `headerWebComponent = true` or `useLegacyHeader = true` is used. Should it be ignored, then we can comment it?

jahow:
> Good idea but we can also do the same for the proposed parameter like `headerWebComponent`, once the web component is the default and the legacy header is removed then we can remove the parameter `headerWebComponent` too.
> 
> Still in both case, what is the behavior of `headerUrl` when either `headerWebComponent = true` or `useLegacyHeader = true` is used. Should it be ignored, then we can comment it?

I would say that if `useLegacyHeader` is true then the setting `headerUrl` is used; otherwise it is ignored.

console/src/main/webapp/manager/app/assets/index.html
```diff
@@ -14,7 +14,8 @@
 
 <body ng-app="manager" ng-strict-di ng-controller="AppController">
 
-  <iframe src="/header/?active=console" scrolling="no" frameborder="0"></iframe>
+  <geor-header style="width:100%;height:90px" active-app="console"></geor-header>
+  <script src="http://localhost:5500/header.js"></script>
```
tkohr, line 18: There's still a forgotten dev value here. However, why is this value used and not the ones above in the `header.jsp`?

f-necas: Fixed :)

tkohr (commented):
Thanks for the huge effort @f-necas ! I've tested all components today (apart from datafeeder and gateway) and everything what I see works fine with the latest fixes :tada: . I just can't officially approve the PR since I was the one who opened it.

Issue #4050, opened 2023-10-13
console - error when email recovery token already used


fvanderbiest:
When accessing for the second time an email recovery link (like https://www.datagrandest.fr/console/account/newPassword?token=c90b6952-2d1f-446d-819d-c5aa6d9ae14d) an unfriendly error is displayed:
![image](https://github.com/georchestra/georchestra/assets/265319/58eaa6fb-46c7-4e0e-9875-da6b26f29ce5)

This can happen with email protection systems validating all incoming links, thus invalidating the token.

Task: providing a nicer, friendly error page

PR #4067, opened 2023-10-31
console - improve password recovery path (#4050)

pmauduit:
Fixing behaviour, if the token is not found then the model is incompletely prepared, and Java won't be able to provide the expected form.

Preferring a redirect to passwordRecovery. This induces an extra 302 instead of getting the page directly, but sounds more logical, to get back the whole form from the begining.

Also storing the "bad.token" error message in session to avoid having to change the default configuration in the datadir (passwordRecovery does not expect parameters), and keeping compatible with the current default geOrchestra configuration.


console/src/main/webapp/WEB-INF/i18n/application_fr.properties
```diff
@@ -46,6 +46,7 @@ passwordRecoveryForm.subtitle=Vous avez perdu votre mot de passe ou votre compte
 passwordRecoveryForm.description=Vous recevrez par courriel un lien vous permettant de définir votre mot de passe.
 passwordRecoveryForm.fieldset.email=Obtenir un nouveau mot de passe
 passwordRecoveryForm.externalPasswordWarningMessage=Vous êtes sur le point d'envoyer un message de remise à 0 du mot de passe utilisateur. Le mot de passe de cet utilisateur est géré par un système externe. Le mot de passe de cet utilisateur sera dès lors géré par geOrchestra.
+passwordRecoveryForm.badtoken=Jeton inconnu ou expirè
```
fvanderbiest, line 49: expiré

fvanderbiest:
Asking @marwanehcine for a review here. Thanks !

console/src/main/java/org/georchestra/console/ws/passwordrecovery/NewPasswordFormController.java
```diff
@@ -106,9 +101,9 @@ public String setupForm(@RequestParam("token") String token, Model model) throws
             return "newPasswordForm";
 
         } catch (NameNotFoundException e) {
-
-            return "passwordRecoveryForm";
-
+            model.asMap().clear();
+            request.getSession().setAttribute("errmsg", "bad.token");
+            return "redirect:/account/passwordRecovery";
```
f-necas, line 106: As only session is used, we could use HttpSession instead.
`public String setupForm(@RequestParam(name = "token", required = false) String token, Model model,
            HttpSession session) throws IOException `
 and 
 `session.setAttribute("errmsg", "bad.token")`
 but it's definitely not a mandatory thing.

f-necas (approved):
Didn't test it but everything seems ok 👍🏼

console/src/main/java/org/georchestra/console/ws/passwordrecovery/NewPasswordFormController.java
pmauduit: you mean, passing the session object directly instead of the whole request ?

pmauduit:
> Didn't test

https://github.com/camptocamp/georchestra-functional-tests/blob/master/src/test/groovy/ConsoleSpec.groovy#L147-L157 ;-)

fvanderbiest:
to be backported to 23.0.x for 23.0.3 too
Thanks !

console/src/main/java/org/georchestra/console/ws/passwordrecovery/NewPasswordFormController.java
pmauduit: after having discussed, it is ok to leave it as it is

pmauduit:
https://github.com/georchestra/georchestra/pull/4077

PR #4073, opened 2023-11-09
geoserver - making sure gs-gwc-rest dependencies are included in the webapp

pmauduit:
Without the REST API, we can still have basic GWC UI integrated to Geoserver, but it won't be usable, as the REST calls will fail.


fvanderbiest:
Thanks Pierre !
Does it need a backport to 23.0 ?

pmauduit:
dunno.

PR #4091, opened 2023-11-17
datafeeder - remove useless WWW:DOWNLOAD link

fvanderbiest:
Given the fact that both GeoNetwork and datahub are able to provide such a download link via the WFS API, it does not make sense to favor the old SHP format (which appears as "unknown format" at the bottom of the list).

image


fvanderbiest:
Tests are failing, but does not seem related to this change.

PR #4094, opened 2023-11-23
datafeeder store in custom non user orgs in datastore

jeanmi151:
Actually the customization of the  datafeeder target database is not possible unlike as it is explicited in the configuration

https://github.com/georchestra/georchestra/blob/master/datafeeder/src/test/resources/datadir/datafeeder/datafeeder.properties#L97C48-L97C56

this PR make the configuration possible


datafeeder/src/main/java/org/georchestra/datafeeder/service/publish/impl/GeorchestraDataBackendService.java
```diff
@@ -181,7 +181,14 @@ private String resolveTargetTypeName(@NonNull DatasetUploadState dataset, Map<St
...
         }
         String schema = nameResolver.resolveDatabaseSchemaName(orgName);
-        connectionParams.put(PostgisNGDataStoreFactory.SCHEMA.key, schema);
+
+        for (String k : connectionParams.keySet()) {
+            String v = connectionParams.get(k);
+            if ("<schema>".equals(v)) {
+                connectionParams.put(PostgisNGDataStoreFactory.SCHEMA.key, schema);
+            }
+        }
```
pmauduit, line 190: so it means that `connectionParams` already contains a `schema` entry ? I don't think it is necessary to iterate over every parameters though. Maybe something like:
```
def schema = connectionParams.get(PostgisNGDataStoreFactory.SCHEMA.key);
if (schema.equals("<schema>") {
  connectionParams.put(PostgisNGDataStoreFactory.SCHEMA.key, schema);
} else {
 // maybe still the need to normalize the field ? https://github.com/georchestra/georchestra/blob/24cd6aa3b245860a8f5a133e2802577af377a68c/datafeeder/src/main/java/org/georchestra/datafeeder/autoconf/GeorchestraNameNormalizer.java#L73-L95
}
```

pmauduit (commented):
I don't know the codebase much than that but some remarks reading the modification:
* no need to iterate over the whole set of parameters from the map
* maybe the field still needs some normalizing if it comes directly from the configuration

datafeeder/src/main/java/org/georchestra/datafeeder/service/publish/impl/GeorchestraDataBackendService.java
jeanmi151: I did it like that to be aligned with this that do the same : https://github.com/georchestra/georchestra/blob/master/datafeeder/src/main/java/org/georchestra/datafeeder/service/publish/impl/GeorchestraOwsPublicationService.java#L213

f-necas (approved):
Ok for me. Should maybe crash if the `datafeeder.publishing.backend.local.schema` key is not provided but it can't work without it anyway.  

Thanks @jeanmi151 for this nice catch !

jeanmi151:
just finished testing, everything works for me : 
- is keys, doesn't exist won't work/push the data somewhere (because no schema provided)
- if value = <schema> works as before, storing into the user orgs schema name
- if value = public (for exemple) works and store it into public schema (as wanted)

Can I merge @pmauduit  ?

I need this to be ported into 23.0.x too

also this is directly connected to https://github.com/georchestra/datadir/pull/360
that need to be merge

jeanmi151:
@groldan is it okay for you ?

jeanmi151:
all good for me... merging !

PR #4102, opened 2023-11-30
console - ensures we remove empty strings from the SUPERUSER recipients

pmauduit:
tests: Added a UT, `mvn clean verify -Pit` OK.


pmauduit:
As it should be considered as an integrity error to have superusers without e-mail in the LDAP, I am wondering if we should not log something when it arises, but who cares of reading logs anyway ? :-) Also no idea how to do this using Java streams.

PR #4103, opened 2023-11-30
console - ensures we remove empty strings from the SUPERUSER recipients

pmauduit:
if some members of the `SUPERUSER` role don't have an e-mail address (or an empty string) defined in their LDAP entry, then the account creation will fail at notifying them by mail, returning a JSON error to the user trying to create an account. 

tests: Added a UT, `mvn clean verify -Pit` OK.

Same as https://github.com/georchestra/georchestra/pull/4102 for `23.0.x`.


f-necas (approved):
Tested and it works well. 

We may create a new issue for delegatAdmin or encapsulate `recipients.add(delegatedAdmin.getEmail());` with a null/empty string condition.

PR #4118, opened 2023-12-13
datafeeder - fix integration tests

f-necas:
Fixing ITs tests by setting the default schema to `schema`


jeanmi151:
still not quite running ;) :p

f-necas:
> still not quite running ;) :p

Yes that's why I've put it in draft juste before your comment. I'm fixing something else

jeanmi151 (approved):
the test IT pass to we can merge

Issue #3975, opened 2023-05-23
console - too much information


fvanderbiest:
`/console/account/passwordRecovery` form:
![image](https://github.com/georchestra/georchestra/assets/265319/985dc561-f58a-4b1a-a694-9770e34a5708)

fvanderbiest:
#214 also leaks emails.

pmauduit:
The current behaviour as of today is to redirect onto the "a mail has been sent with a link to recover / change your password" (as long as the email is "syntaxically correct", regardless if the e-mail is in the LDAP or not. I don't know if this is the intended behaviour. But at least we don't leak the existence anymore.

fvanderbiest:
This looks good to me !
Thanks

PR #4126, opened 2023-12-19
Console : Update recovery message

f-necas:
Solves issue 
- https://github.com/georchestra/georchestra/issues/3975


fvanderbiest:
In case of `NameNotFoundException` I think we should not even display a specific message.
In fact, there should be no difference whether the email is present in the LDAP or not.

fvanderbiest:
Awesome. To be ported to 23.0.x for the upcoming 23.0.5 release (it's a bugfix). Thanks !

edevosc2c:
23.0.6 was released yesterday :D

PR #4134, opened 2024-01-02
Adding i18n to email template for new account creation

marwanehcine:
On this PR, we  make use of I18n messages  while building email text template to be sent when a new account is created (OAuth2 or not)


fvanderbiest:
Thanks Marwane.

If we introduce i18n for one console message, we should translate all of them.
But most platforms customize these templates. So I think i18n is useless here - we should let platform administrators manage them.

fvanderbiest (changes requested):
I'd be in favor of reverting the changes relative to i18n

PR #4136, opened 2024-01-05
console - block org description to 255 chars

f-necas:
Used to avoid `Caused by: java.sql.BatchUpdateException: Batch entry 0 insert into Groups`in Geonetwork (GN limited to 255 chars)


fvanderbiest (approved):
LGTM, please backport to 23.0.x too

PR #4137, opened 2024-01-05
Datafeeder : Allow import of csv files

f-necas:
# CSV in datafeeder 

Allows to import CSV geo and non geo data in datafeeder. 

For the front part, see PR : https://github.com/georchestra/geonetwork-ui/pull/9

Upsteam PR : https://github.com/geonetwork/geonetwork-ui/pull/773

## Screenshot
image


datafeeder/pom.xml
```diff
@@ -530,8 +539,8 @@
           </dependency>
         </dependencies>
         <configuration>
-          <source>${java.version}</source>
-          <target>${java.version}</target>
+          <source>11</source>
+          <target>11</target>
```
pmauduit, line 543: why ?

datafeeder/src/main/java/org/georchestra/datafeeder/service/DatasetsService.java
```diff
@@ -445,6 +459,10 @@ private boolean isShapefile(@NonNull Path path) {
         return path.getFileName().toString().toLowerCase().endsWith(".shp");
     }
 
+    private boolean isCsv(@NonNull Path path) {
+        return path.getFileName().toString().toLowerCase().endsWith(".csv");
```
pmauduit, line 463: there is a `FilenameUtils.getExtension(String path)` in apache commons-io which seems to handle specific corner-cases of (Windows) filenames, in case the current approach is judged too fragile (this is just a suggestion, I don't mind keeping it as it is, it seems that it is already what was done with shapefiles anyway).

datafeeder/pom.xml
f-necas: Just my IDE shouting at me...

datafeeder/src/main/java/org/georchestra/datafeeder/batch/service/DataUploadAnalysisService.java
```diff
@@ -175,6 +166,43 @@ public DatasetUploadState analyze(DatasetUploadState item) throws Exception {
...
+
+    @VisibleForTesting
+    public String csvSixLinesAsBase64(String path) throws Exception {
+        ByteArrayOutputStream os = new ByteArrayOutputStream();
+        try (FileInputStream fis = new FileInputStream(path)) {
+            int numLines = 0;
+            while (true) {
+                int b = fis.read();
+                if (b == '\n' || b == '\r') {
+                    numLines++;
```
pmauduit, line 194: thinking of this, I'm wondering if the CSV being provided is a Windows-like text file ('\r\n'), then maybe we will always return 3 lines instead of 6.

datafeeder/pom.xml
```diff
@@ -341,6 +350,11 @@
       <artifactId>jersey-hk2</artifactId>
       <version>${jersey-version}</version>
     </dependency>
+    <dependency>
+      <groupId>com.univocity</groupId>
+      <artifactId>univocity-parsers</artifactId>
+      <version>2.9.1</version>
+    </dependency>
```
pmauduit, line 357: looks like it is not in use anymore (checking `com.univocity.*` imports in the current diff)

datafeeder/src/main/java/org/georchestra/datafeeder/batch/service/DataUploadAnalysisService.java
f-necas: I've pushed a sort of solution to it. With a continue on `\r` char, it should handle CRLF and LF correctly.

pmauduit: ok, sounds legit

f-necas:
Force pushed squashed commits. See branch `old-csv-support` for full commits history

groldan (changes requested):
@f-necas this looks good overall.
Please address the pending change request (curl example and pom), and squash into a single commit with a good title and description.
Also there's a test failure, we'd need to address that.

datafeeder/pom.xml
```diff
@@ -19,7 +19,7 @@
     <java.version>1.8</java.version>
```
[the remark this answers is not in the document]
f-necas: If I remove it, I need to explicitly add 1.8 to maven-compiler-plugin :
```
//pom.xml:540
<configuration>
          <source>1.8</source>
          <target>1.8</target>
          <encoding>UTF-8</encoding>
          <annotationProcessorPaths>
            <path>
              <groupId>org.projectlombok</groupId>
```
Don't know what's the best.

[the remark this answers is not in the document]
pmauduit: > If I remove it, I need to explicitly add 1.8 to maven-compiler-plugin

Is it something related to your IDE ? If so, I'll leave the pom as it is originally, and trust maven instead.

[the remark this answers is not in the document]
pmauduit: I'd suggest to set the java.version to 11 in the DF pom.

fvanderbiest:
The PR should include a line in the release notes to explain that the dataset name is now generated from the dataset title rather than the original file name. This is an important change.

pmauduit:
> The PR should include a line in the release notes to explain that the dataset name is now generated from the dataset title rather than the original file name. This is an important change.

This has been done: https://github.com/georchestra/georchestra/pull/4188/files#diff-fa9ce18bbd795fbe01c3be557f7253cbb15ef64db8e5e94e97ac8bdcdd38a7d4R102

pmauduit:
IT testsuite is a bit flacky, because it seems to time-out from time to time. No issue launching it locally, the GHA actually went green at the 4th try, taking 12+ minutes to run, I had run it locally in less than 2 minutes, so probably a matter of available resources ? Considering a large amount of resources is needed to run the suite.

PR #4142, opened 2024-01-11
geoserver - updating submodule

pmauduit:
* backport GEOS-11235 from upstream (session not closed on GS when using the georchestra gateway), see https://osgeo-org.atlassian.net/browse/GEOS-11235
* themable header, see https://github.com/georchestra/geoserver/pull/35


pmauduit:
Note: I don't know why dependabot did not work, I'd have expected a PR for the themable header which has been merged 3 weeks ago.

pmauduit:
force-pushed because my local copy was not up-to-date.

landryb:
candidate for a backport in 23.0 ?

pmauduit:
the first one (the one I worked on) is specific to the georchestra gateway, when using GS behind a SP, we are not affected by the issue encountered.
The second one is about the new header, which is not part of 23.0.x yet as far as I am aware of (@f-necas ?)

landryb:
ah sorry i hadnt seen it was in gateway situation. phew. yes for the second the new header isnt in 23.0.

PR #4143, opened 2024-01-11
Copy custom jetty env if file is found

edevosc2c:
This custom file is necessary for having geoserver JNDI working properly.

Instead of built-in the copy of the file through the docker composition or the helm chart.

I propose to have this copy directly in the container.

Probably to be backported in 23. Won't hurt the actual fonctionality of geoserver.

### DIFF
--- a/geoserver/webapp/src/docker/docker-entrypoint.d/01-copy-jetty-env-if-found
+++ b/geoserver/webapp/src/docker/docker-entrypoint.d/01-copy-jetty-env-if-found
@@ -1,0 +1,8 @@
+#!/bin/sh
+FILE=/etc/georchestra/geoserver/jetty-env.xml
+if [ -f "$FILE" ]; then
+    echo "Copying jetty-env.xml"
+    cp $FILE /var/lib/jetty/webapps/geoserver/WEB-INF/jetty-env.xml
+else
+    echo "jetty-env.xml not found, ignoring."
+fi


geoserver/webapp/src/docker/docker-entrypoint.d/01-copy-jetty-env-if-found
```diff
@@ -0,0 +1,8 @@
+#!/bin/sh
+FILE=/etc/georchestra/geoserver/jetty-env.xml
```
pmauduit, line 2: `/etc/georchestra` is purely conventional, don't we have an env variable which could be used instead of the hardcoded path ?

pmauduit: actually it is also set / hardcoded into the Dockerfile, so I'm ok with the change:
https://github.com/georchestra/georchestra/blob/master/geoserver/webapp/src/docker/Dockerfile#L42

jeanpommier (commented):
Really ? Naming the env. var `FILE` ? What about giving it an explicit name, wouldn't it be better ?

Apart from that, sounds like a good idea.

jeanpommier:
Oh, no, sorry, I hadn't looked close enough, it's just local variable, doesn't matter. My bad

PR #4145, opened 2024-01-12
console - delegation - uid containing dots fix

pmauduit:
When the user identifier contains a dot, spring will truncate the path variable, leading the code not being able to resolve the concerned user.

This is basically the same fix as 466244b1, but on the `DelegationController`.

Tests: added an IT, `mvn clean verify -Pit` OK locally


pmauduit:
Codacy fails because:

* My test class is named IT and does not match the expected regex for test file naming convention, I think we can safely ignore it.
* There are no assertion in my test method (but we have `andExpect()`-like methods instead), I think it can also be ignored.

fvanderbiest:
Thanks Pierre. LGTM.

fvanderbiest:
To be backported to 23.0.x when merged.

f-necas (approved):

PR #4152, opened 2024-01-15
[Console] Avoid searching on B64 image in orgs

f-necas:
Set a simplified org object to avoid search inside b64 image


fvanderbiest (approved):
Awesome ! Makes sense indeed.
To be backported to 23.0.x too (it's a nice bug). Thanks !

PR #4170, opened 2024-01-25
[Console] New header and long logins

f-necas:
## New header

Adds missing new header on `account/` URLs

## Long logins 

Supports long logins (often coming from external sources) and truncate or break-word them.


fvanderbiest (approved):
LGTM

PR #4181, opened 2024-02-01
console - allow bigger organization thumbnails

jeanmi151:
this is to increase by the default (200000) value of maxFormContentSize in console


fvanderbiest:
Why did we close it ?
I think 200kb is too small, and 5Mb too big for the organization thumbnail

fvanderbiest:
I think we should go on with this, and set the limit to 1Mb

jeanmi151:
because this is only for docker/containers specific

fvanderbiest:
At least it will work in this context !

fvanderbiest:
I think the value of 5Mb is too big, we should set it at 1Mb

f-necas (approved):

PR #4183, opened 2024-02-04
commons, console - add a isExternal flag & hide password update forms for users authenticated externally

marwanehcine:
In preauth mode, no need to manage user password, a code refactor is done to handle this case.

### DIFF
--- a/commons/src/main/java/org/georchestra/security/model/GeorchestraUser.java
+++ b/commons/src/main/java/org/georchestra/security/model/GeorchestraUser.java
@@ -92,6 +92,9 @@ public class GeorchestraUser implements Serializable {
     /** Maps to Account.getOAuth2Uid */
     private String oAuth2Uid;
 
+    /** Provided by request header {@code sec-external-authentication} */
+    private Boolean isExternalAuth = false;
+
     public void setRoles(List<String> roles) {
         this.roles = roles == null ? new ArrayList<>() : roles;
     }
--- a/console/src/main/java/org/georchestra/console/ws/edituserdetails/EditUserDetailsFormBean.java
+++ b/console/src/main/java/org/georchestra/console/ws/edituserdetails/EditUserDetailsFormBean.java
@@ -36,6 +36,7 @@ public class EditUserDetailsFormBean implements java.io.Serializable {
     private String description;
     private String postalAddress;
     private boolean isOAuth2;
+    private boolean isExternalAuth;
 
     @Override
     public String toString() {
@@ -131,4 +132,13 @@ public class EditUserDetailsFormBean implements java.io.Serializable {
     public void setIsOAuth2(boolean isOAuth2) {
         this.isOAuth2 = isOAuth2;
     }
+
+    public boolean getIsExternalAuth() {
+        return isExternalAuth;
+    }
+
+    public void setIsExternalAuth(boolean isExternalAuth) {
+        this.isExternalAuth = isExternalAuth;
+    }
+
 }
--- a/console/src/main/java/org/georchestra/console/ws/edituserdetails/EditUserDetailsFormController.java
+++ b/console/src/main/java/org/georchestra/console/ws/edituserdetails/EditUserDetailsFormController.java
@@ -19,6 +19,7 @@
 
 package org.georchestra.console.ws.edituserdetails;
 
+import static org.georchestra.commons.security.SecurityHeaders.SEC_EXTERNAL_AUTHENTICATION;
 import static org.georchestra.commons.security.SecurityHeaders.SEC_USERNAME;
 
 import java.io.IOException;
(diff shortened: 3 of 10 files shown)


console/src/main/java/org/georchestra/console/ws/edituserdetails/EditUserDetailsFormController.java
```diff
@@ -67,6 +67,11 @@
 @SessionAttributes(types = EditUserDetailsFormBean.class)
 public class EditUserDetailsFormController {
 
+    private static final String PREAUTH_IDP = "preauth-idp";
+    private static final String PREAUTH_USER_ID = "preauth-userid";
+
```
pmauduit, line 72: `preauth` prefixed headers are meant to be consumed by the georchestra gateway. Once behind the gateway, the proxified webapps should not receive them (see documentation here: https://github.com/georchestra/georchestra-gateway/blob/main/docs/pre-authentication.adoc), but receive the usual `sec-*` prefixed headers.

console/src/main/webapp/WEB-INF/i18n/application.properties
```diff
@@ -153,3 +153,5 @@ shortName.label=Short name
 name.label=Name
 address.label=Address
 orgType.label=Organization Type
+preauth.name.label=Identity Provider
+preauth.uid.label=User ID
```
pmauduit, line 157: I think we are mixing 2 concepts:

* preauthentication: which is the mechanism where the authentication is performed by another component before hitting georchestra
* external identity providers

if we are using the same labels, maybe we should find something more "generic" for both concepts.

pmauduit (changes requested):
I think there is a misconception here (see comments). We don't want to globally disable password modifications, but disable it when it is not relevant, eg. when the user is connected via:

* cascaded SASL LDAP authentication (already implemented)
* third-party identity provider
* preauthentication via headers

As far as I understand the current PR, the intent is to globally disable it. Maybe it will require to introduce a header sent by the gateway or the SP to indicate that the authentication is "external" / not managed from georchestra.

commons/src/main/java/org/georchestra/security/model/GeorchestraUser.java
```diff
@@ -92,6 +92,9 @@ public class GeorchestraUser implements Serializable {
     /** Maps to Account.getOAuth2Uid */
     private String oAuth2Uid;
 
+    /** Provided by request header {@code sec-external-authentication} */
+    private String isExternalAuth = "false";
```
pmauduit, line 96: Why not a boolean instead of a String ?

console/src/main/java/org/georchestra/console/ws/edituserdetails/EditUserDetailsFormBean.java
```diff
@@ -35,7 +35,7 @@ public class EditUserDetailsFormBean implements java.io.Serializable {
     private String org;
     private String description;
     private String postalAddress;
-    private boolean isOAuth2;
+    private String isOAuth2;
```
pmauduit, line 38: same remark

console/src/main/webapp/WEB-INF/views/changePasswordForm.jsp
```diff
@@ -41,91 +41,105 @@
 
 <body>
     <%@ include file="header.jsp" %>
+    <% String externalAuthentication=request.getHeader("sec-external-authentication");%>
```
pmauduit, line 44: Don't we have access to the GeorchestraUser here, instead of having to work with the raw http headers ?

pmauduit (commented):
Gobally the same remark as the one I did on the georchestra-gateway pull-request

console/src/main/webapp/WEB-INF/views/changePasswordForm.jsp
marwanehcine: We have access only to 'changePasswordForm'.
We need to handle access to this page globally (for OAuth2 and PreAuth cases). It would be better to use header to handle both cases, I think.

pmauduit: grepping the codebase, I can't see anywhere else where we call `request.getHeader()` directly from a JSP file.

Taking what is done in `EditUserDetailsFormController` as an example, I'd have opted for putting the flag into the model ; also maybe introducing a constant into the SecurityHeaders class from the georchestra-commons would have been relevant, at least it would have been closer to how things are already done here.

* https://github.com/georchestra/georchestra/blob/master/console/src/main/java/org/georchestra/console/ws/edituserdetails/EditUserDetailsFormController.java#L115
* https://github.com/georchestra/georchestra/blob/master/commons/src/main/java/org/georchestra/commons/security/SecurityHeaders.java#L53-L65

marwanehcine: Hello,
  - request.getHeader() is not used in console but it has been used on header module
  - in EditUserDetailsFormController, we get user details from LDAP (EditUserDetailsFormController.java, line 116). We have already made the decision to not put external auth boolean into LDAP. So it is not possible to put this information into model object.
  - regarding SecurityHeaders constants, i have updated code in gateway and georchestra to use SecurityHeaders static constants.

console/src/test/java/org/georchestra/console/ws/changepassword/ChangePasswordControllerTest.java
```diff
@@ -86,7 +91,9 @@ public void initForm() {
     @Test
     public void setupForm() throws Exception {
         userIsSpringSecurityAuthenticatedAndExistInLdap("me");
-        String ret = ctrlToTest.setupForm(model);
+        request.addHeader(SEC_EXTERNAL_AUTHENTICATION, "false");
+
+        String ret = ctrlToTest.setupForm(request, response, model);
         assertEquals("changePasswordForm", ret);
     }
```
pmauduit, line 99: Maybe it would deserve the opposite test ? (when the header is set to `true`)

pmauduit (approved):
LGTM

PR #4193, opened 2024-02-20
[CONSOLE] Force confirmation on org deletion

f-necas:
Actual behavior : 

Org are directly deleted when button is click.

New behavior :

A confirm popup is displayed to allow org deletion.

### DIFF
--- a/console/src/main/webapp/manager/app/app.es6
+++ b/console/src/main/webapp/manager/app/app.es6
@@ -177,7 +177,20 @@ angular.module('manager', [
       const alphanum = v => v && v.match(regexp)
       ctrl.$validators.shortname = alphanum
     }
-  }))
+  })).directive('ngConfirmClick', [
+    function () {
+      return {
+        link: function (scope, element, attr) {
+          var msg = attr.ngConfirmClick || 'Are you sure?'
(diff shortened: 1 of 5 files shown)


console/src/main/webapp/manager/app/app.es6
```diff
@@ -177,7 +177,20 @@ angular.module('manager', [
       const alphanum = v => v && v.match(regexp)
       ctrl.$validators.shortname = alphanum
     }
-  }))
+  })).directive('ngConfirmClick', [
+    function () {
+      return {
+        link: function (scope, element, attr) {
+          var msg = attr.ngConfirmClick || 'Are you sure?'
```
fvanderbiest, line 184: does it need a translation ? or is it already ?

f-necas: The translation is inside the attribute : `ng-confirm-click="{{ 'org.delete' | translate}} ?"`

In this case it displays : `Delete organization ?`

fvanderbiest: Yes, I saw it, but my question was about the availability of translation strings for 'Are you sure?' and 'Delete organization ?' since they are not provided by the PR.

PR #4200, opened 2024-02-29
Fix expired role route and add tests

f-necas:
This PR fixes the route `console/private/roles/EXPIRED` which hit a 404.

In console, it now displays the role correctly : 
image


jeanmi151 (approved):
Tested and working on local compile docker

PR #4201, opened 2024-02-29
Invalidate session on applications when user content changes

emmdurin:
When a change is made on a user, we want geOrchestra applications to invalidate and reload their session to take changes in account. Changes are detected by matching application session data to `sec-*` headers sent by gateway, assuming theses headers are always kept up to date by gateway.


console/src/main/java/org/georchestra/console/filters/UserChangeFilter.java
```diff
@@ -0,0 +1,39 @@
...
+import org.springframework.security.core.context.SecurityContextHolder;
+
+import javax.servlet.*;
+import javax.servlet.http.HttpServletRequest;
+import javax.servlet.http.HttpServletResponse;
+import java.io.IOException;
+import java.util.Set;
+import java.util.stream.Collectors;
+
+public class UserChangeFilter implements Filter {
```
pmauduit, line 14: 2 remarks on the current implementation choices:

* you are using plain `javax.servlet` Filter, as we are using spring / spring security, why not integrating a spring filter to the spring security filter chain instead ? (https://docs.spring.io/spring-security/reference/servlet/architecture.html#adding-custom-filter). Also considering the fact that you are actually using spring security objects in the code (SecurityContextHolder on line 22), it would make more sense to me to stay in a "pure spring" approach. It will also make any move to jakartaEE easier, as we are only relying on the framework which should abstract the `javax.*` / `jakarta.*` upgrade / renaming.
* If the idea is to be able to reuse the same filter across different georchestra webapps, I'd nest the code into another module, like commons or spring-security-proxy-integration

f-necas: I moved it to commons but didn't succeed to make it working in `applicationContext-security.xml`. 

App didn't react as expected..

f-necas (approved):

pmauduit (approved):
So as discussed privately: it is not possible to use the spring security filter chain instead of plain servlet config, because we have no way of filtering against the path, when using spring security mechanisms. It could be done codewise, by inspecting the ServletRequest object, but then will not be portable / console-specific.

Also even if it would make sense to me to react the same way on each URLs on the session invalidation, this does not seem to work at runtime and provides a poor user experience.

I am not competent enough on the browser/js interactions to judge, sorry

PR #4203, opened 2024-03-08
console - handle case when no `sec-external-authentication` http header is sent

marwanehcine:
Following PR remark on the georchestra-gateway here:
https://github.com/georchestra/georchestra-gateway/pull/101#discussion_r1517732378

There is a possibility that the header is not sent to the proxified webapps when this is not relevant, hence the modifications introduced by PR: https://github.com/georchestra/georchestra/pull/4183 could lead to an NPE.

Adding an extra check to avoid it.

### DIFF
--- a/console/src/main/java/org/georchestra/console/ws/changepassword/ChangePasswordFormController.java
+++ b/console/src/main/java/org/georchestra/console/ws/changepassword/ChangePasswordFormController.java
@@ -42,6 +42,7 @@ import org.springframework.web.bind.annotation.SessionAttributes;
 
 import javax.servlet.http.HttpServletRequest;
 import javax.servlet.http.HttpServletResponse;
+import java.util.Objects;
 import java.util.Optional;
 
 import static org.georchestra.commons.security.SecurityHeaders.SEC_EXTERNAL_AUTHENTICATION;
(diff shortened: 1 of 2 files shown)


console/src/main/java/org/georchestra/console/ws/changepassword/ChangePasswordFormController.java
```diff
@@ -93,8 +94,8 @@ public String setupForm(HttpServletRequest request, HttpServletResponse response
             throws DataServiceException {
         Optional<String> uid = getUsername();
         if (uid.isPresent()) {
-            boolean isExternalAuth = Boolean
-                    .parseBoolean(SecurityHeaders.decode(request.getHeader(SEC_EXTERNAL_AUTHENTICATION)));
+            boolean isExternalAuth = !Objects.isNull(request.getHeader(SEC_EXTERNAL_AUTHENTICATION))
```
pmauduit, line 97: you can use `Objects.nonNull` : https://docs.oracle.com/javase/8/docs/api/java/util/Objects.html#nonNull-java.lang.Object-

pmauduit (changes requested):
I allowed myself to edit the description, so that other reviewers can get a bit more of context of what this PR provides.

Also, I'd have expected a modification into the following test from the previous mentioned PR:
https://github.com/georchestra/georchestra/blob/master/console/src/test/java/org/georchestra/console/ws/changepassword/ChangePasswordControllerTest.java#L92-L98

As we don't send the header at all when it is not relevant, then we should not receive a header set to false.

PR #4208, opened 2024-03-14
console - email address change workflow

emmdurin:
Added a workflow to change user e-mail address.

Still missing things :
* ~~check e-mail address format before sending validation e-mail~~
* ~~e-mail conflict checking~~
* ~~e-mail template to fine tune~~
* ~~use a separate column in token DB to store e-mail address instead of reusing UID column~~
* ~~translations~~


console/src/test/resources/templates/changeemail-email-template.txt
```diff
@@ -0,0 +1,12 @@
+Dear {name},
+
+You asked to change your e-mail address on {publicUrl}/.
+If you did not request any password update, just ignore this e-mail, you're safe (I DON'T THINK SO).
```
landryb, line 4: e-mail update ?

landryb:
does this make sure that the new e-mail isn't already in use by another existing user ? iirc an email can/should be linked with only one account..

emmdurin:
> does this make sure that the new e-mail isn't already in use by another existing user ? iirc an email can/should be linked with only one account..

If it is already in use, it will silently fail with an exception when the user loads validation link, and will return to user details pages without have the e-mail changed.

We may check for conflicts before, but what to do next ? If we inform the user that new e-mail already exists, doesn't this have a security concern so that he could know if a specific e-mail from anyone is registered on the platform ? I ask this because I see that in password recovery procedure, we do not allow users to know if the e-mail entered is registered.

console/src/test/resources/templates/changeemail-email-template.txt
emmdurin: If you are reacting about the "I DON'T THINK SO", I had a doubt when I wrote it about a possible security problem, but now I think I misunderstood the workflow and there is no longer a problem if a user receive such an e-mail without having asked for it. It just means that someone else entered his e-mail in change form, but without validating it won't do anything more.

landryb:
> If we inform the user that new e-mail already exists, doesn't this have a security concern so that he could know if a specific e-mail from anyone is registered on the platform ?

that's already the case in the account creation form, if you try to register with an email that is already linked with an existing account you get rejected/pointed at the password reset form

image

so yes, maybe you can see that as a security concern/user list harvesting, but if you look at it from the pov of the end-user or an user admin that has confused users on the phone, making sure that it is clearly reported to the user that this new e-mail can't be used (before sending anything to anyone) is important.

emmdurin:
> so yes, maybe you can see that as a security concern/user list harvesting, but if you look at it from the pov of the end-user or an user admin that has confused users on the phone, making sure that it is clearly reported to the user that this new e-mail can't be used (before sending anything to anyone) is important.

OK, so we should go this way and report it to the user

landryb:
i welcome the feature but i don't think i qualify to review this java/jsp madness ;)

f-necas:
I pushed a commit to :

- Format UserTokenDao file
- Update email template and some translations
- Move a console admin log to a normal dev log
- Add a log translations for console admin page 

image
`logs.emailchangeemailsent` is no longer needed as it won't be logged here

f-necas (approved):
LGTM :) Thanks for this nice feature !

As discussed we can improve IT tests on it but I hope we could find time to re-develop the console without jsp and angularJS behind it.

We can also use hibernate in future to get rid on those queries in every file.

*I also have to fix this geonetwork CI ...*

Issue #4210, opened 2024-03-19
non-gateway header shows undefined undefined instead of givenname/firstname


landryb:
seen on demo.georchestra.org which uses the s-p and not the gateway, the s-p returns this for `/whoami`:
```
{
  "GeorchestraUser": {
    "roles":
       "ROLE_SUPERUSER",
       "ROLE_MAPSTORE_ADMIN",
       "ROLE_USER",
       "ROLE_ADMINISTRATOR",
       "ROLE_GN_ADMIN",
       "ROLE_EMAILPROXY"
    ],
    "username":"testadmin"
  }
}
```
this comes from https://github.com/georchestra/georchestra/blob/master/security-proxy/src/main/java/org/georchestra/security/Proxy.java#L371 and doesn't contain firstname/lastname which are consumed by the new webcomponent header

the gateway apparently returns much more things for `whoami`:
https://github.com/georchestra/georchestra-gateway/blob/master/gateway/src/main/java/org/georchestra/gateway/app/GeorchestraGatewayApplication.java#L81

there's no such thing as userMapper in the s-p/georchestra code ?

landryb:
![image](https://github.com/georchestra/georchestra/assets/3421760/b0882740-53f8-4fc9-8914-8fb828751c41)

is the user experience

pmauduit:
Here is the layout of a /whoami response from a gateway:

```
GeorchestraUser: {
  username: "pmauduit",
  roles: [
    "ROLE_SUPERUSER",
    "ROLE_GN_ADMIN",
    "ROLE_USER",
    "ROLE_ADMINISTRATOR"
  ],
  organization: "C2C",
  id: "d5f1491a-f5f8-4998-ae51-76cff38d500a",
  lastUpdated: "7c81fd7e1d81b233fc6ce3550f908a82cddc02dc6a725b6aa1e6c533f93b5c3d",
  firstName: "Pierre",
  lastName: "Mauduit",
  email: "pierre.mauduit@xxx",
  postalAddress: null,
  telephoneNumber: "+336123456",
  title: null,
  notes: null,
  ldapWarn: false,
  ldapRemainingDays: null,
  oauth2Provider: null,
  oauth2Uid: null
}
```

PR #4212, opened 2024-03-21
sp - harmonize /whoami endpoint output with the gateway (#4210)

pmauduit:
Autowire the `usersApi` bean to the Proxy class so that we can resolve a `GeorchestraUser` object from the connected user. This allows to get the same java object as the one being returned by the same endpoint from the Gateway.

Tests:
* utests adapted
* runtime tested with testadmin:testadmin + anonymous user (e.g. blank call to /whoami)

```
% curl -u testadmin:testadmin http://localhost:8080/whoami
{"GeorchestraUser":{"lastUpdated":"035cabcdc6fdd76c5d8670dbc534e6123a5d04f2a7a80dd5cad138c6eb40a895","firstName":"Test","lastName":"ADMIN","notes":"Internal CRM notes on testadmin","roles":["ADMINISTRATOR","SUPERUSER","GN_ADMIN","USER","MAPSTORE_ADMIN","EMAILPROXY"],"organization":"PSC","id":"0c6bb556-4ee8-46f2-892d-6116e262b489","email":"psc+testadmin@georchestra.org","username":"testadmin"}}

% curl http://localhost:8080/whoami
{"GeorchestraUser":null}
```

Note: I'd have liked to leave the JSON serialization to spring, as it is done into the Gateway, but could not make it. This would have avoided the 10+ calls to `JSONObject.put()`, but sounds acceptable to me for the time both SP and Gateway will coexist.

### DIFF
--- a/security-proxy/src/main/java/org/georchestra/security/Proxy.java
+++ b/security-proxy/src/main/java/org/georchestra/security/Proxy.java
@@ -22,6 +22,7 @@ package org.georchestra.security;
 import com.google.common.annotations.VisibleForTesting;
 import com.google.common.collect.ImmutableMap;
 import com.google.common.collect.Maps;
+import lombok.Setter;
 import org.apache.commons.lang3.StringUtils;
 import org.apache.commons.logging.Log;
 import org.apache.commons.logging.LogFactory;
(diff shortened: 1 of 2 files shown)


jeanmi151:
working for me too

PR #4213, opened 2024-03-21
Documentation 

pierrejego:
Move documentation to standard geOrchestra documentation module

Initialize new method
Create a first entry page ( Installation and requirements ) 

Move the current docs folder to docsv1 ( link still in use in the documentation ) 

TODO continue to add documentation / migrate interresting point from current documentation 

Documentation will be accessible from here : https://www.georchestra.org/documentation.html


landryb:
@jeanpommier i suppose that's related but now the link to 'installation guide' on https://www.georchestra.org/documentation.html leads to a 404 on https://github.com/georchestra/georchestra/blob/master/docs/index.md

jeanpommier:
Ah, yes, for now we should point it to docv1. I'll look if other links are broken too.

For sure, this is a potentially disruptive change...

jeanpommier:
> https://www.georchestra.org/documentation.html leads to a 404

fixed

Issue #4227, opened 2024-04-09
[geowebcache] still has the iframe header


landryb:
seen in georchestra/datadir#397, cf https://github.com/georchestra/georchestra/blob/master/geowebcache-webapp/src/main/java/org/georchestra/geowebcache/GeorchestraGeoWebCacheDispatcher.java#L58

PR #4228, opened 2024-04-10
gwc - manages new georchestra header (#4227)

pmauduit:
Supports the new webcomponent based georchestra header in the geowebcache app.

Tests: 
* verify OK
* mainly runtime, switching the flag in the default.properties datadir file to check the behaviour.

### DIFF
--- a/geowebcache-webapp/src/main/java/org/georchestra/geowebcache/GeorchestraGeoWebCacheDispatcher.java
+++ b/geowebcache-webapp/src/main/java/org/georchestra/geowebcache/GeorchestraGeoWebCacheDispatcher.java
@@ -60,16 +60,6 @@ public class GeorchestraGeoWebCacheDispatcher extends GeoWebCacheDispatcher impl
      * */
     private String georchestraStylesheet = "";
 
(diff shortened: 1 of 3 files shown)


f-necas:
I think we may include stylesheets and other thing by default ?  
Like `<geor-header active-app="geowebcache" legacy-header="@useLegacyHeader@" legacy-url="@headerUrl@" style="width:100%;height:@headerHeight@px;border:none;" logo-url="@logoUrl@" stylesheet="@georchestraStylesheet@"></geor-header>`

pmauduit:
> I think we may include stylesheets and other thing by default ?

Then we'd need to document the properties into the datadir's default.properties:
https://github.com/georchestra/datadir/blob/master/default.properties#L20-L47

what are the default values for the different properties ?

pmauduit:
> what are the default values for the different properties ?

https://github.com/georchestra/datadir/blob/master/default.properties#L20-L47

f-necas (approved):
Thanks @pmauduit. I'll have to find a way to implement geowebcache inside the new header 😄

Issue #4196, opened 2024-02-21
console - properties from datadir are not overrideable anymore


pmauduit:
Following https://github.com/georchestra/georchestra/pull/4047, it is not possible anymore to override properties from `default.properties` with the `console/console.properties` file, because the `DefaultPropsInitializer` class forces a precedence on `default.properties` over the other properties files.

So if we have a general LDAP configuration for geOrchestra into the default.properties file, and a custom LDAP configuration set into console/console.properties, the properties from the latter are never taken into account.

pmauduit:
Tested by removing the DefaultPropsInitializer class (from the classpath + reference in web.xml), I could get back the expected behaviour.

pmauduit:
I have got something almost working, taking back what #4047 intended to do, and fixing this issue (by removing the DefaultPropsInitializer).

PR #4229, opened 2024-04-10
console - rework "making rabbitMQ optional" (#4047,#4196)

pmauduit:
Rework of PR #4047 

* Removing the DefaultPropsInitializer class (see #4196 for the motivation)
* Adding conditional configuration class to load (or not) the 2 added xml files in the classpath which define beans related to rabbitMq.
* loading the properties files from the datadir into the rabbitMq xml files so that we can resolve the variables being used.
* some (related and unrelated) cleanup (e.g. autowired objects which were actually unused)

Tests:
* clean verify OK
* IT added


pmauduit:
codacy complaining because added test classes are ending with "IT" instead of expected "Test".

f-necas (approved):
Thanks @pmauduit for this nicer way to enable/disable rbmq !
Just a small PR on datadir to explain/document `enableRabbitmqEvents` and it's good for me.

pmauduit:
> Just a small PR on datadir to explain/document `enableRabbitmqEvents` and it's good for me.

There already is a property / comment in the default.properties here:
https://github.com/georchestra/datadir/blob/master/default.properties#L80-L81

(not that meaningful though)

PR #4238, opened 2024-04-25
console - make sure properties file from the datadir are sourced when configuring rabbitMq

pmauduit:
This is following PR #4229, which was not completely loading the full context when checking if rabbitMQ was enabled, the intended behaviour was true only if rabbitmq related properties were also defined as environment variables (or Java properties), and did not take into account properties file from the datadir.

Tests: mainly runtime, as it is transparent from the integration test (we define the expected properties via a TestPropertySource annotation).


f-necas (approved):
Awesome, thanks @pmauduit, Gateway is back on track with it :)

PR #4243, opened 2024-05-23
Fixed GDPR account data download when OGC statistics are disabled

emmdurin:
When OGC statistics is disabled, related tables are not present in database so trying to download complete account data using dedicated button in Data protection panel on user profile fails when trying to retrieve data from OGC statistics table with this error :

```
java.lang.IllegalStateException: org.georchestra.ds.DataServiceException: org.postgresql.util.PSQLException: ERROR: relation "ogcstatistics.ogc_services_log" does not exist
  Position: 59
	at org.georchestra.console.ds.AccountGDPRDaoImpl.visitOgcStatsRecords(AccountGDPRDaoImpl.java:175)
```

### DIFF
--- a/console/src/main/java/org/georchestra/console/ds/AccountGDPRDaoImpl.java
+++ b/console/src/main/java/org/georchestra/console/ds/AccountGDPRDaoImpl.java
@@ -172,7 +172,7 @@ public class AccountGDPRDaoImpl implements AccountGDPRDao {
                     AccountGDPRDaoImpl::createOgcStatisticsRecord, consumer);
             log.info("Extracted {} OGC statistics records for user {}", reccount, userName);
         } catch (DataServiceException e) {
-            throw new IllegalStateException(e);
+            log.info("No OGC statistics recorded for user {}", userName);
         }
     }
 


f-necas:
I'd prefer a minimal explanation on the PR. What is this PR for ? How does it handle gdpr things ? What's displayed to the end user ? A blank page ?

emmdurin:
I don't understand your point. The purpose of the feature triggering the error is to retrieve all data linked to a user account and give it to him as a ZIP file. OGC statistics are a part of theses data. If it is disabled, there is no statistics data linked to the account, so there is no statistics data to retrieve and give to the user. We just have to return all other data and ignore the missing statistics.

f-necas:
Oh then I misread your PR description. 
Can the end-user download something now ?  

Then, why not conditionnaly execute `visitOgcStatsRecords` function ? 
```
if (<ogcEnabled>)visitOgcStatsRecords()
```

emmdurin:
Maybe we can do that, but I'm afraid that there may be some glitches just after a parameter is changed. If we can assume that table must be manually created when enabling statistics and manually deleted when disabling statistics, I think it is OK. If it is not done, in first case user will get an error, in second case user won't get all stored data.

f-necas:
If it's a NPE we can check the nullabillity of ogcEnabled. And if someone enable it (we can set it to false by default), it's his concern to check if the table in the database.

In any case the `ogcEnabled` var should be documented with something like this.

- ogcEnabled works only with security proxy 
- if you enable it, check that the table exists

f-necas (approved):

fvanderbiest:
@f-necas do you think it would it be interesting to have this in 24.0.x too ?

Issue #3808, opened 2022-09-22
TLS versions support


fvanderbiest:
For compatibility we had to explicitly authorize TLS V1 for geoserver:
https://github.com/georchestra/georchestra/blob/f8e67f5fde57f8f089263af72664f081773f8f9e/geoserver/webapp/src/docker/Dockerfile#L48

I'm not sure this is relevant anymore
Also, TLS v1.3 is to be considered I guess.

A bit of history: this was introduced by https://github.com/georchestra/georchestra/commit/18ecf5a511972624a5a2746de543c97dc2de5506 to prevent `Could not list layers for this store, an error occurred retrieving them: Remote host closed connection during handshake` when hitting remote OGC services.

edevosc2c:
If the application support TLS1.3 then only `-Dhttps.protocols=TLSv1.2,TLSv1.3` should be used.

fvanderbiest:
This was my idea too :-)

fvanderbiest:
TLS 1.0 and 1.1 are currently deprecated on all major browsers - https://caniuse.com/tls1-1

PR #4245, opened 2024-06-06
Force TLS 1.2 and TLS 1.3 only

edevosc2c:
Fixes https://github.com/georchestra/georchestra/issues/3808


jeanpommier:
Seems fine to me. Did you check the geoserver issue mentioned in the ticket, to make sure it is not relevant anymore ?

edevosc2c:
No, but I was also thinking whenever we still want to force TLS version or just rely on java default TLS versions.

Maybe the latter one is better?

pmauduit:
> Did you check the geoserver issue mentioned in the ticket, to make sure it is not relevant anymore ?

In my opinion, the geoserver issue related to this ticket was about having GS hitting remote services not supporting recent versions of TLS. Considering it was almost 8 years ago, we can assume that remote services are now compliant with more recent versions of TLS ?

pmauduit (approved):

Issue #3930, opened 2023-04-04
datafeeder - when a job fails, the credentials being used under the hood are revealed by mail & into the logs


pmauduit:
Here is the mail template received after a failure:

```
Dear Pierre,

We're sorry to inform you that the publication process for your
armoires dataset submitted on 2023-04-04 15:09:20.998
has encountered an error and could not finish correctly.

The platform administrator has been notified too.

Job identifier: a26d1277-e1ef-4cac-9528-347222f7c0e3
Reported error: Error publishing the following datasets:
armoires: Status 500:
Request: POST https://host/geoserver/rest/workspaces/c2c/datastores/datafeeder_c2c/featuretypes HTTP/1.1
Accept: application/json

Authorization: Basic <XXXXXX>

Content-Length: 468
Content-Type: application/json
```

Here are the logs of the datafeeder service:
![image](https://user-images.githubusercontent.com/594335/229838319-b17bd5cb-05e3-4e3a-a58a-fc295af9d640.png)

pmauduit:
The default configuration / datadir suggests to use sec-* headers behind the SP before the following commit: https://github.com/georchestra/datadir/commit/92ec9add8e51429459ec5ba7f5550ab0c9eb23dd, so keeping configured using headers should be safe, but this has to be fixed in case of using a basic-authentication instead

pmauduit:
I have a (partial ?) fix for this one, at least when the error is related to GeoServer, see PR above, feedback welcome.

PR #4250, opened 2024-06-12
datafeeder - avoid providing sensitive informations during a GS failure (#3930)

pmauduit:
This partially addresses the mentioned issue by:

* limiting the output by stripping the `ServerException.getMessage()`
* avoid printing the full exception in the logger

Some concerns:

* This might not fully address the mentioned issue (what about GN errors ? other ones ?)
* Maybe this should be fixed into the gs-restconfig client library instead of into the DF / application code

Considering the deadlines for the geOrchestra release, I did not want to investigate further upstream.

Tests:
* utests added to make sure the message parsing is safe
* runtime tested (setting a wrong password for the geoserver_privileged_user user)


datafeeder/src/main/java/org/georchestra/datafeeder/batch/service/PublishingBatchService.java
```diff
@@ -239,6 +256,19 @@ private void doOnEachRunningDataset(DataUploadJob job, Consumer<DatasetUploadSta
         }
     }
 
+    @VisibleForTesting
+    String formatGsClientServerExceptionMessage(String message) {
+        String error;
+        try {
+            error = IntStream.range(0, 2)//
+                    .mapToObj(index -> message.split("\n")[index])//
+                    .collect(Collectors.joining());
```
pmauduit, line 265: Suggested by @f-necas:

```
message.lines().limit(2).collect(Collectors.joining("\n"))
```

datafeeder/src/main/java/org/georchestra/datafeeder/batch/service/PublishingBatchService.java
```diff
@@ -239,6 +253,11 @@ private void doOnEachRunningDataset(DataUploadJob job, Consumer<DatasetUploadSta
         }
     }
 
+    @VisibleForTesting
+    String formatGsClientServerExceptionMessage(String message) {
+        return message.lines().limit(2).collect(Collectors.joining("\n"));
```
f-necas, line 258: And if we want to be a little bit more performant we can use `reduce` instead. 😄 
```suggestion
        return message.lines().limit(2).reduce("", String::concat));
```

f-necas (approved):
LGTM, it's not ideal but at least basic auth isn't shared.

datafeeder/src/main/java/org/georchestra/datafeeder/batch/service/PublishingBatchService.java
pmauduit: what does it change ? I feel a bit lost with Java streams to be honest

f-necas: As I understand `reduce` apply modification on stream elements themselves instead of `collect` which execute a function on mutable objects.

I just tried to understand a little bit more and eneded up with this : 
```
		long startTime;
        long stopTime;

        startTime = System.nanoTime();
        System.out.println(error.lines().limit(2).collect(Collectors.joining()));
        stopTime = System.nanoTime();
        System.out.println(stopTime - startTime);

        startTime = System.nanoTime();
        System.out.println(error.lines().limit(2).reduce("", String::concat));
        stopTime = System.nanoTime();
        System.out.println(stopTime - startTime);
```
```
 Error publishing the following datasets:  armoires: Status 500:
2473545
 Error publishing the following datasets:  armoires: Status 500:
727366
```

pmauduit: it is unclear to me, but reading https://stackoverflow.com/questions/22577197/java-8-streams-collect-vs-reduce, it looks like collect performs better than reduce. I guess for the needed computing time here to concatenate 2 strings, performance is not an issue, but still.

pmauduit: with the suggested reduce() version, we lose the line feeds, but I think it is acceptable.

PR #4251, opened 2024-06-13
Add isExternal to full account and conditionnal display in console

f-necas:
This PR allows to hide login and change password blocks in console when user comes from an external auth.
image


landryb:
maybe there should be a visual clue that the fields are not there/hidden because the account is external ? otherwise user admins might be puzzled. unless there's already a clear clue from the values carried by the external account...

f-necas:
@landryb Thanks for your review ! I've added an `External account` label in order to keep uid (which could be useful to distinguish users) :
image

landryb:
> @landryb Thanks for your review ! I've added an `External account` label in order to keep uid (which could be useful to distinguish users) : image

looks purrrfect :)

PR #4254, opened 2024-06-19
Forgot to display oAuth2 provider name

f-necas:
Adds also external provider name


landryb (approved):
i suppose it makes sense, what is displayed ? something a user admin can understand ?

Issue #4276, opened 2024-07-03
gwc - Upgrade GeoWebCache


fvanderbiest:
geOrchestra 24 ships GWC 1.16.x when upstream is 1.25.x 
It's high time to update.

pmauduit:
https://mvnrepository.com/artifact/org.geowebcache/gwc-core/1.25.2

edevosc2c:
Will need to be backported in geOrchestra 23.1.x too

PR #4282, opened 2024-07-11
gwc: update to 1.25.2

f-necas:
https://github.com/georchestra/georchestra/issues/4276


pmauduit:
looking at the failing tests, it looks like you have an issue with the spring version(s) being used, `mvn dependency:tree` might help here.

landryb (approved):

pmauduit (approved):
LGTM, but we might want to disambiguate the GEOWEBCACHE_CONFIG_DIR vs georchestra datadir's  `geowebcache/geowebcache.xml`, I suspect that geowebcache will require a place where it can write its own configuration, but I am not sure this is the reason why we have 2 different directories.

PR #4273, opened 2024-06-27
switch to GN 4.4.5

landryb:
with all the hard work from @f-necas in georchestra/geonetwork#297 .. ready for testing for the ones using master.

i'm not sure the GN PR should be merged to 4.2.x branch there, rather `[georchestra-gn4.4.5-rebase](https://github.com/georchestra/geonetwork/tree/georchestra-gn4.4.5-rebase)` branch should be renamed to `georchestra-gn4.4.x` and the submodule point to the tip of it as this PR does.

PR #4283, opened 2024-07-12
update maven version use by the mvnw wrapper to 3.8.3

landryb:
required for gn 4.4.5, cf #4273


landryb:
bah, no need for the pr since it doesnt trigger the GN workflow, i've cherrypicked the commit and pushed it to #4273

PR #4308, opened 2024-08-12
Set users order in console alphabetically on surname field

f-necas:
Easier for finding users across pages in console app.

### DIFF
--- a/console/src/main/java/org/georchestra/console/ws/backoffice/users/UsersController.java
+++ b/console/src/main/java/org/georchestra/console/ws/backoffice/users/UsersController.java
@@ -25,13 +25,7 @@ import java.text.ParseException;
 import java.text.SimpleDateFormat;
 import java.time.LocalDate;
 import java.time.format.DateTimeParseException;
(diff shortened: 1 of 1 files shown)


pmauduit:
I wonder if we want to force such a behaviour or if it should not be a parameter to the endpoint, like `?sort=surname` or so, and if the surname could be null, but LGTM.

f-necas:
@pmauduit updated to allow null values :
```java
List<User> users = new ArrayList<>();
users.add(new User("John", "Bsurname"));
users.add(new User("Jane", ""));
users.add(new User("John", "Vsurname"));
users.add(new User("Jane", null));
users.add(new User("John", "csurname"));
users.sort(Comparator.comparing(User::getSurname, Comparator.nullsLast(String::compareToIgnoreCase)));
System.out.println(users);
//[User{name='Jane', surname=''}, User{name='John', surname='Bsurname'}, User{name='John', surname='csurname'}, User{name='John', surname='Vsurname'}, User{name='Jane', surname='null'}]
```

PR #4310, opened 2024-08-14
console - use exec maven instead of mvn-scm-plugin for checkout

f-necas:
Can noticeably decrease building time on local pc by cloning with --depth 1

### DIFF
--- a/console/pom.xml
+++ b/console/pom.xml
@@ -594,25 +594,30 @@
         <finalName>${project.artifactId}-generic</finalName>
         <plugins>
           <plugin>
-            <groupId>org.apache.maven.plugins</groupId>
-            <artifactId>maven-scm-plugin</artifactId>
-            <version>1.11.2</version>
-            <configuration>
-              <checkoutDirectory>${project.build.directory}/deb/etc/georchestra</checkoutDirectory>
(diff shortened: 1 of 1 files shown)


landryb:
im not sure im qualified to review maven goo :)

- i suppose your initial comment meant *decrease* build time ?
- using plain exec instead of a dedicated plugin sounds like a loss
- isnt the dedicated plugin supposed to handle shallow clones via the `shallow` parameter ? cf https://github.com/apache/maven-scm/pull/56
- why only the console ? the same construct is found in other webapps:
```
analytics/pom.xml
console/pom.xml
datafeeder-ui/pom.xml
datafeeder/pom.xml
geoserver/webapp/pom.xml
geowebcache-webapp/pom.xml
header/pom.xml
security-proxy/pom.xml
```

f-necas:
I tried to use the shallow option multiple times with the maven scm plugin but without success. If it suits to everyone, i will edit other apps.

pmauduit:
> we close it ?

I think we can live with the scm plugin for now. Another topic which relates to this PR and does not bring to a consensus right now is "do we still want/need to provide a sample datadir in our deliverables ?".

landryb:
agree with @pmauduit .. exec feels wrong, even if mvn-scm-plugin doesnt provide the feature :/

PR #4315, opened 2024-08-20
gwc - introduces an environment-aware xml configuration reader

pmauduit:
This PR introduces a class which is basically a copy/paste from
the upstream XMLConfiguration class, which adds some converters
to the xstream configuration, so that `${...}` strings in the
XML are resolved against the environment variables.

Caveat: default implementation seems to also manage backup of the
configuration file. In this implementation, we forbid the possibility
to serialize back the configuration into the original XML file, because
it would dump the resolved env variables back again.

Caveat 2: This should be considered as a temporary solution, as
a PR on the XMLConfiguration class which would follow the GS / GWC
convention (e.g. honoring the `ALLOW_ENV_PARAMETRIZATION` java
property) should be issued upstream and hopefully merged, so that we
could stick to upstream / not custom-geOrchestra code.

Tests: 2 utests added, runtime tested by setting a description from an
env variable.

### DIFF
--- a/geowebcache-webapp/pom.xml
+++ b/geowebcache-webapp/pom.xml
@@ -160,6 +160,12 @@
       <scope>test</scope>
     </dependency>
     <dependency>
+      <groupId>com.github.stefanbirkner</groupId>
+      <artifactId>system-rules</artifactId>
+      <version>1.19.0</version>
+      <scope>test</scope>
+    </dependency>
+    <dependency>
       <groupId>com.hazelcast</groupId>
       <artifactId>hazelcast</artifactId>
       <version>${hazelcast.version}</version>
--- a/geowebcache-webapp/src/main/java/org/georchestra/geowebcache/config/EnvironmentAwareXMLConfiguration.java
+++ b/geowebcache-webapp/src/main/java/org/georchestra/geowebcache/config/EnvironmentAwareXMLConfiguration.java
@@ -1,0 +1,1498 @@
+/**
+ * This program is free software: you can redistribute it and/or modify it under the terms of the
+ * GNU Lesser General Public License as published by the Free Software Foundation, either version 3
+ * of the License, or (at your option) any later version.
+ *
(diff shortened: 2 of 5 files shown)


pmauduit:
Note: apache commons-text would have allowed to use some interpolators, e.g. using `${env:BLAH}` instead and several interesting things (xml xpaths, base64-encode/decode, ...) ; but it is made for plain-text and won't XML-escape the string automagically, so it would have left the responsability to the administrator that environment-passed values are XML-safe ...

landryb:
i know k8s and friends wants you to put everything in env vars, but .. passwords in env vars.. can be a huge security risk outside of the kloug ;)

pmauduit:
> passwords in env vars.. can be a huge security risk outside of the kloug ;)

sure, but this PR should not alter the former behaviour, just adding the possibility to do so.

Systemd also allows to pass some environment variables to a service, but in both setups, in case of a RCE vulnerability, a `cat /proc/self/environ` and we are doomed. so I don't get why the risk would be different between regular setups and docker/k8s ?

landryb:
> so I don't get why the risk would be different between regular setups and docker/k8s ?

because i suppose docker/k8s takes extra care of not providing access to env vars to random users logged in on the host.. or i hope so, given the amount of stuff that gets pushed there ;)

pmauduit:
> because i suppose docker/k8s takes extra care of not providing access to env vars to random users logged in on the host..

but if the prerequisite is to be logged in to the computer running the service, then either way, it's trivial to get the  environment variables back as cleartext ?

edevosc2c (approved):

groldan:
ok, so, as far as I can see, this allows to parameterize the http username/password for WMS layers, and for the configuration defaults.

That's a cool feature, do you mind if I try to come up with a pull request for upstream?

pmauduit:
> That's a cool feature, do you mind if I try to come up with a pull request for upstream?

sure, feel free :-)

pmauduit:
superseded by https://github.com/GeoWebCache/geowebcache/pull/1364 upstream.

PR #4324, opened 2024-09-09
User link for oauth users available in admin logs

f-necas:
When an external account is created, logs in console doesn't have links because the target is the email and not the user local name. 

In [logger.es6](https://github.com/georchestra/georchestra/blob/master/console/src/main/webapp/manager/app/components/logger/logger.es6#L173), it needs to be inside users to get the right match


fvanderbiest:
Thx !
to be backported to 24

PR #4327, opened 2024-09-10
Add missing oauth tpl

f-necas:
Not supposed to crash because template is loaded in a separate method which handles IOException


console/src/main/webapp/WEB-INF/templates/new-oauth2-account-notification-template.txt
```diff
@@ -0,0 +1,15 @@
+Dear administrator,
+
+A new OAuth2 user has signed up on {instanceName} :
```
fvanderbiest, line 3: no space before semicolon

fvanderbiest: A new user signed up using OAuth2 on  {instanceName} :

Issue #4330, opened 2024-09-11
console - rabbitMq - library versions mismatch


pmauduit:
https://github.com/georchestra/georchestra/blob/master/console/pom.xml#L273-L282

Discovered by working on https://github.com/georchestra/georchestra-gateway/issues/143 ; we should remove one of the spring-amqp explicit dependency, as spring-rabbit depends on it anyway.

Also it seems that migrating to v2. or later will require some extra work, as the v1 is explicitely being used in the xml configuration file:
https://github.com/georchestra/georchestra/blob/master/console/src/main/resources/spring/rabbit-listener-context.xml#L5

pmauduit:
it looks like it is fixed by mentioned PR above, closing.

PR #4332, opened 2024-09-11
console - rework rabbitMQ integration (#4330)

pmauduit:
### 1. Fixing version mismatch between dependencies in the pom

Some weeks ago, a version bump of `spring-amqp` broke the rabbitmq integration (see CICD of the gateway and
https://github.com/georchestra/georchestra-gateway/issues/143 for the details). While digging into the issue, it turned out that PR #4295 upgraded the dependency to a version non compatible with `spring-rabbitmq`, which generated a docker image for the console which won't be viable in regards to the rabbitmq integration.

spring-rabbitmq already depends on spring-amqp, which by the way should be forced into a specific version somewhere else. Sticking to 1.7.12.RELEASE of spring-rabbit, so that we keep a compatible version with amqp, and removing the explicit dependency to spring-amqp.

### 2. Rework console message consumption

When a user account from a third-party oauth2 identity provider is saved into the LDAP (at first connection through the gateway), a message is sent via rabbitMq and consumed by the console.

In order to avoid consumption of the same message several times, a `Set<String>` was created as a static member of the `RabbitmqEventsListener` class. But if an exception popped before reaching the call to add the message uid to the list, then the `onMessage` method was then called once again, in a kind of (aggressive) infinite loop.

We had some issues with the console service recently on one of our infrastructure, and we suspect that this behaviour generated a huge amount of unecessary logs because of this infinite loop.

This PR suggests to rewrite the `onMessage()` in the following way:

* removing the `Set<String>` (what will it look like in some centuries of console uptime ?)
* log the error to the logs
* discards the message (by simply returning from the `onMessage()` without throwing any exception)

### 3. tests

Added an integration test (along with a dependency against testcontainers).

Actually adding this new IT  was the reason leading me to the 2. above. And by running the testsuite with maven instead of directly the test into my IDE, I stumbled upon this: https://github.com/georchestra/georchestra/blob/master/console/pom.xml#L584-L585 but could not figure out how limiting forks to 2 was a way to avoid race conditions, as it is enough to already get IT stepping onto each other. Fact is the ports / hostnames being used by the testcontainers are passed using `system.setProperties()` calls before being consumed into the spring xml definition files, and having 2 tests running in parallel is already enough to trigger a race condition, targetting the wrong docker containers. Best way to workaround this would be to use a `DynamicPropertySource`, but it does not exist yet in the spring version being used here ... :(

### 4. Caveats

we are late on the spring-amqp/rabbit version (and spring version generally speaking), but at least we will have a test ready when we will go for a version bump.

### DIFF
--- a/console/pom.xml
+++ b/console/pom.xml
@@ -272,13 +272,8 @@
     </dependency>
     <dependency>
       <groupId>org.springframework.amqp</groupId>
-      <artifactId>spring-amqp</artifactId>
-      <version>2.3.11</version>
-    </dependency>
-    <dependency>
-      <groupId>org.springframework.amqp</groupId>
       <artifactId>spring-rabbit</artifactId>
-      <version>1.1.1.RELEASE</version>
+      <version>1.7.12.RELEASE</version>
     </dependency>
     <!-- commons -->
     <dependency>
@@ -418,6 +413,12 @@
     <dependency>
       <groupId>org.georchestra</groupId>
       <artifactId>georchestra-testcontainers</artifactId>
+      <scope>test</scope>
+    </dependency>
+    <dependency>
+      <groupId>org.testcontainers</groupId>
+      <artifactId>rabbitmq</artifactId>
+      <version>1.16.0</version>
       <scope>test</scope>
     </dependency>
   </dependencies>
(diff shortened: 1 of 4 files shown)


f-necas:
Related : error was coming from a missing template which ended up with an NPE and didn't send that the message was received.
- https://github.com/georchestra/georchestra/pull/4329

f-necas (approved):
The message seems to be definitely lost but in this case it's not very problematic. Is there a way to monitor lost messages ? We could invest this possibility in another PR.

Anyway, we should merge #4329 with this one.

pmauduit:
> Is there a way to monitor lost messages ? We could invest this possibility in another PR.

apart from checking the logs, I don't know.

PR #4336, opened 2024-09-13
Hide user list in edit user details conditionnaly.

f-necas:
Introduce a new configuration parameter in 24.0.x and upper :
`gdpr.displayMemberList=false/true` by default `false`

If users are empty, doesn't display the list in editUserDetailsForm.


fvanderbiest:
Thanks Florian.
I'd say `gdpr.displayMembersList` (with an additional `s`)

fvanderbiest (approved):

PR #4344, opened 2024-09-19
datafeeder - Change ogc api links at end of process

f-necas:
Allows to implement multiple schemas in datafeeder : 
- https://github.com/georchestra/data-api/pull/39

It modifies the link at the end of the process to implement prefix. Can be released only if a new release of data-api is done.

**double-check needed**
I don't know where my mistake is, I had to add an YamlFactory to read data-api's configuration file


f-necas:
No they were set only with `dataset.getPublishedName()`, but to avoid setting the prefix in data-api and datafeeder, I prefered parsing directly data-api's conf file. 

Where I think there's a mistake is mostly about the way I import the file. I don't really understand why I have to implement a `YamlPropertySourceFactory` to read yaml file. Isn't it implemented by default by spring ?

pmauduit:
What I am a bit concerned about is that it induces a dependency from the DF to the data-api, in term of configuration at least :-/
In the other cases (DF -> GN / GS) there is a direct communication with the component, but since the data-api is not even aware of publication details. So I am a bit clueless here, is the yaml file from the data-api a strong requirement ? Also maybe it is not necessary to get all the configuration details from the file, so I was wondering if an extra webservice into the data-api would make it ?

f-necas:
It won't fail if it doesn't exists, it create a `PostgisSchemasConfiguration` bean with default config (schemas *, prefix_tables: true). 
And if this value isn't set in datafeeder.properties `datafeeder.publishing.ogcfeatures.public-url: https://georchestra.mydomain.org/ogc`, it won't go through this condition, and won't create ogc link.

We could query data-api with a webservice, and would be a good idea. It could be done with a refactor of the DF, where we remove all those hard links to GN and GS ?

pmauduit:
> In the other cases (DF -> GN / GS) there is a direct communication with the component, but since the data-api is not even aware of publication details, what shall we do ? I am a bit clueless here, is the yaml file from the data-api a strong requirement ? (e.g. will it fail without ? is there a way to disable this data-api interaction ?) Also maybe it is not necessary to get all the configuration details from the file, so I was wondering if an extra webservice into the data-api would make it (even totally out of scope from the ogc api specs) ?

Thinking out loud, I am not actually shocked of accessing the configuration file from the data-api, just that for now it is not into the datadir officially, so maybe we'd need something like `-Ddata-api.configuration.filepath=...` instead of assuming it would be into the georchestra datadir.

pmauduit:
Also mentioning the introduction of a new webservice data-api side does not mean I was thinking it was a good idea :-)
I was concerned of what it could expose + being out of scope of the ogc api specs, so we are probably better off grabbing the configuration instead.

datafeeder/src/main/java/org/georchestra/datafeeder/autoconf/GeorchestraDatadirAutoConfiguration.java
```diff
@@ -20,19 +20,24 @@
...
 @Profile("georchestra")
-@PropertySource(value = { //
-        "file:${georchestra.datadir}/default.properties", //
-        "file:${georchestra.datadir}/datafeeder/datafeeder.properties" }, //
-        ignoreResourceNotFound = false)
+@EnableConfigurationProperties(PostgisSchemasConfiguration.class)
+@PropertySources({ @PropertySource(value = { "file:${georchestra.datadir}/default.properties", //
+        "file:${georchestra.datadir}/datafeeder/datafeeder.properties" }, ignoreResourceNotFound = false),
+        @PropertySource(value = {
+                "file:${georchestra.datadir}/data-api/application.yaml" }, ignoreResourceNotFound = true, factory = YamlPropertySourceFactory.class) })
```
pmauduit, line 40: I don't know if it would make sense to have it in its own dedicated class like `DataApiAutoConfiguration`, also with a system property dedicated to the data-api configuration path, instead of assuming it is into the georchestra datadir (there are no data-api configuration provided in the default georchestra datadir yet).

pmauduit: Why not even a dedicated profile ? like "data-api" ?

datafeeder/src/main/java/org/georchestra/datafeeder/autoconf/PostgisSchemasConfiguration.java
```diff
@@ -0,0 +1,134 @@
...
+import java.util.Map;
+import java.util.Optional;
+import java.util.Set;
+import java.util.function.Function;
+import java.util.stream.Collectors;
+
+@Data
+@EqualsAndHashCode(exclude = "includesBySchema")
+@ConfigurationProperties(prefix = "postgis.schemas")
+public class PostgisSchemasConfiguration implements InitializingBean {
```
pmauduit, line 25: Is it copy/pasted from the data-api codebase (https://github.com/georchestra/data-api/blob/main/src/services/ogc-features/src/main/java/com/camptocamp/opendata/ogc/features/autoconfigure/postgis/PostgisSchemasConfiguration.java) ? Then maybe we should add a comment to point the original file at least.

pmauduit:
> I had to add an YamlFactory to read data-api's configuration file

Indeed, not sure but if you go for a new profile, and set into your application.yaml something like:

```
---
spring.config.activate.on-profile: data-api
spring.config.import: ..., ${data-api.config.path}
```
then maybe you don't need the extra class.

f-necas:
It seems that we can't use .import cause of spring boot version. 

As we may globally improve DF, there's no need to bump version here.

f-necas:
@pmauduit for re-review the PR :)

datafeeder/src/main/java/org/georchestra/datafeeder/api/DataPublishingApiController.java
```diff
@@ -53,6 +55,8 @@ public class DataPublishingApiController implements DataPublishingApi {
     private @Autowired DataUploadService uploadService;
     private @Autowired DataPublishingResponseMapper mapper;
     private @Autowired DataFeederConfigurationProperties props;
+    private @Autowired GeorchestraNameNormalizer nameResolver;
+    private @Autowired(required = false) PostgisSchemasConfiguration postgisSchemasConfiguration;
```
pmauduit, line 59: there are some guidelines which indicate that we should not use Autowired, but since it looks like it is already the case here, I won't object (At least I'm pretty sure we had a talk about that with @groldan).

pmauduit (approved):

Issue #4343, opened 2024-09-19
datafeeder - Error while uploading - Does file contains valid geo data ? 


f-necas:
Sometimes datafeeder ends up with an error when publishing 

Error : Check that the file contains a valid geospatial dataset
In french : Le fichier transféré contient-il des données géographiques valides ?

`/geoserver/rest/workspaces/<workspace>/datastores/<store>/featuretypes`
STATUS:	400
MESSAGE:	Trying to create new feature type inside the store, but no attributes were specified

It goes inside this condition but shouldn't :
https://github.com/georchestra/geoserver/blob/2.25.x-georchestra/src/restconfig/src/main/java/org/geoserver/rest/catalog/FeatureTypeController.java#L225-L226

~~Could be related to (which often happen before this error) : #4342~~

EDIT : See 
- #4346

PR #4346, opened 2024-09-22
datafeeder - Don't override conf and warn if local and geoserver schema are desync

f-necas:
If `datafeeder.publishing.backend.local.schema=public` and `datafeeder.publishing.backend.geoserver.schema=<schema>` are unsync, it may crash beacause geoserver tries to get the data in another schema. It results of DF frontend showing the error, "Is this valid geo data ?", and backend showing an error on geoserver featuretypes which has no attributes.

Also create a new Hashmap to avoid erasing connectionParams.

Solves #4343

### DIFF
--- a/datafeeder/src/main/java/org/georchestra/datafeeder/service/publish/impl/GeorchestraDataBackendService.java
+++ b/datafeeder/src/main/java/org/georchestra/datafeeder/service/publish/impl/GeorchestraDataBackendService.java
@@ -24,10 +24,7 @@ import java.io.IOException;
 import java.sql.Connection;
 import java.sql.SQLException;
 import java.sql.Statement;
-import java.util.Arrays;
-import java.util.HashSet;
-import java.util.Map;
-import java.util.Set;
+import java.util.*;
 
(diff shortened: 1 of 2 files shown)


fvanderbiest:
A bit puzzled - if the app is not ready to have two parameters with different values, why don't we keep a single one ?

f-necas:
> A bit puzzled - if the app is not ready to have two parameters with different values, why don't we keep a single one ?

Because they are used in different parts of the app and it avoids to read backend.local in the geoserver part :
Local : https://github.com/georchestra/georchestra/blob/master/datafeeder/src/main/java/org/georchestra/datafeeder/service/publish/impl/GeorchestraDataBackendService.java#L179
Geoserver : https://github.com/georchestra/georchestra/blob/master/datafeeder/src/main/java/org/georchestra/datafeeder/service/publish/impl/GeorchestraOwsPublicationService.java#L214

jeanmi151 (approved):
Looks good to me, but i am not a java developer :p

PR #4392, opened 2025-01-29
fix: GS - bump GS version

pmauduit:
This PR ensures we integrate the following one:
https://github.com/georchestra/geoserver/pull/39

tests: compilation + decompiled the gs-wms module in the generated GS webapp, to make sure the added `EntityResolver` is indeed being used for entity resolution.

### DIFF
--- a/geoserver/pom.xml
+++ b/geoserver/pom.xml
@@ -30,6 +30,7 @@
     <module>geoserver-submodule/src/restconfig</module>
     <module>geoserver-submodule/src/gwc</module>
     <module>geoserver-submodule/src/web/core</module>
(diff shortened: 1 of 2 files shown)


landryb:
im not fully sure this is desired but isnt this partially backporting the header config file feature to 24.0 ?

pmauduit:
> isnt this partially backporting the header config file feature to 24.0 ?

there are indeed some modifications related to the header in the branch being used from the GS submodule, but the new header is already present in the 24.x serie, so when I PR-ed I presumed that the GS branch was just not uptodate.

pmauduit:
@f-necas is not available but told me privately he was ok with this PR.

PR #4398, opened 2025-02-04
Remove gn submodule from georchestra

f-necas:
Removing gn submodule to avoid realeasing a full georchestra for only gn releases


f-necas:
Close as no correct solution has been found

PR #4189, opened 2024-02-15
Issue 4168 - Console - RGPD "export my data" - Geonetwork on an other database

pierrejego:
This PR add a new datasource to connect to geonetwork database.

New fields have to be added in console.properties : 
# PostGreSQL database connection parameters to geonetwork
pgsqlGNHost
pgsqlGNPort
pgsqlGNDatabase
pgsqlGNUser
pgsqlGNPassword

PR will be done for those new params in datadir and ansible.

PR #4405, opened 2025-02-10
Add default values to pgsql GN

f-necas:
Following the PR #4189, it adds default values to pgslGN* datasource.

Avoid console not starting if valeus aren't set explicitly.


pmauduit (approved):
I am not aware of these `*GN*` variables, but https://github.com/georchestra/datadir/blob/master/console/console.properties#L362-L369 / LGTM

PR #4429, opened 2025-03-06
Change datafeeder output to match title and data-api output

f-necas:
Geoserver layer will be set on title instead of file name


fvanderbiest:
A long-awaited change !
Many thanks Florian.

fvanderbiest (approved):

PR #4435, opened 2025-03-12
console: Change logger limit for console dashboard page to improve loading time

f-necas:
Set logger limit to 100 instead of 100000

### DIFF
--- a/console/src/main/webapp/manager/app/components/logger/logger.es6
+++ b/console/src/main/webapp/manager/app/components/logger/logger.es6
@@ -46,7 +46,7 @@ class LoggerController {
     // manage query params to get user's or complete logs
     let typeQuery = 'Logs'
     const params = {
-      limit: 100000,
+      limit: 100,
       page: 0
     }
     if (this.user) {


pmauduit:
> Set logger limit to 100

500, relying on the modified code ;-)

pmauduit (approved):

landryb:
thanks for that, much welcome. our existing admin logs amounted to a whooping 7.87Mb json so far.. so it will definitely help loading times on the console :)

landryb:
after having deployed this to production, the 500 records limit make the admin logs request down to 100k,which gives us records of changes over the last two months. much faster loading time.

but there's a drawback: i haven't found a way in the logs tab of the console to make it refetch logs for older periods, changing the year selector or trying a custom period, the logs shown are still within the 'last 500' timeframe. so there's apparently no way to get access to older events.

f-necas:
Damn, I just looked deeper and it doesn't call other pages ! 

I think it has been done like that as a workaround to avoid calling pages from backend. Which is always the page 0 (the XHR call we see in Devtools)...

landryb:
sorry for throwing you in that rabbit hole ;)

PR #4450, opened 2025-03-27
security-proxy: ensure /whoami replies with application/json;charset=utf-8

### LINKED ISSUE #4449: /whoami replies with text/plain:charset=ISO-8859-1 content-type
quite unexpectedly:
```
$curl -s  -v  http://localhost:8180/whoami 2>&1 |grep Content-Type:
< Content-Type: text/plain;charset=ISO-8859-1
```
even if the [code](https://github.com/georchestra/georchestra/blob/master/security-proxy/src/main/java/org/georchestra/security/Proxy.java#L365) specificies `produces = MediaType.APPLICATION_JSON_UTF8_VALUE`

this leads to issues when rendering accentued chars in the header and the page has a `meta` with `charset=utf-8`:

Image

landryb:
fixes #4449 here, might be improvable because i dont know java.

### DIFF
--- a/security-proxy/src/main/java/org/georchestra/security/Proxy.java
+++ b/security-proxy/src/main/java/org/georchestra/security/Proxy.java
@@ -56,8 +56,10 @@ import org.json.JSONObject;
 import org.springframework.beans.factory.BeanInitializationException;
 import org.springframework.beans.factory.annotation.Autowired;
 import org.springframework.beans.factory.annotation.Qualifier;
+import org.springframework.http.HttpHeaders;
 import org.springframework.http.HttpMethod;
 import org.springframework.http.MediaType;
+import org.springframework.http.ResponseEntity;
 import org.springframework.security.cas.ServiceProperties;
 import org.springframework.security.core.Authentication;
 import org.springframework.security.core.context.SecurityContextHolder;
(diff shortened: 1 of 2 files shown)


security-proxy/src/main/java/org/georchestra/security/Proxy.java
```diff
@@ -364,12 +366,14 @@ public void handleDefaultRequest(HttpServletRequest request, HttpServletResponse
      */
     @GetMapping(value = "/whoami", produces = MediaType.APPLICATION_JSON_VALUE)
     @ResponseBody
-    public String whoami(HttpServletRequest request) throws DataServiceException {
+    public ResponseEntity<String> whoami(HttpServletRequest request) throws DataServiceException {
```
groldan, line 369: I think if you return the JSONObject directly instead of a string representation the encoding will be properly handled by GetMapping with produces = MediaType.APPLICATION_JSON_VALUE?

groldan (changes requested):

security-proxy/src/main/java/org/georchestra/security/Proxy.java
landryb: i've tried it (returning directly the JSONObject), and spring only returned 406 codes, whatever the accept header i manually set (and we cant change all the consumers of the API to set that header)

landryb:
fwiw, this works for me in production backported on 24.0, so i'd like to merge this. Not being a java expert, i'd like guidance on how to improve the existing PR, but other than that...

pmauduit (approved):
I'm still perplexed as to why we need to do so, and what is wrong with our attempt to use the framework here, but fair enough, as discussed on IRC.

security-proxy/src/main/java/org/georchestra/security/Proxy.java
landryb: > I think if you return the JSONObject directly instead of a string representation the encoding will be properly handled by GetMapping with produces = MediaType.APPLICATION_JSON_VALUE?

i know i've tried that, but i've read somewhere that `produces` didnt do what we expect (eg from my understanding it was more 'sanity check the type of the value' and **not** 'set the `content-type` header' -  in the end i mostly tried the examples from https://www.baeldung.com/spring-mvc-set-json-content-type.

looking at spring-projects/spring-framework#19050 there's a `@DefaultContentType` annotation ?

maybe that's due to the fact the S-P uses spring 4.3.22 which is somewhat old (2019) ?

anyway, the S-P is on life support, so maybe not worth spending too much time on this.

PR #4455, opened 2025-04-08
datafeeder - Some improvements on metadata publishing

pmauduit:
This PR aims to provide some improvements on datafeeder's metadata publication job:

* adds the ability to skip metadata publication on GeoNetwork
* impersonates (e.g. transfer ownership to) the user having actually pushed the dataset on the created metadata
* in case of role-based groups synchronization in GeoNetwork: adds the editing privileges to each groups the publisher belongs to

Adding a flag in the configuration so that the publication of the metadata becomes optional. If `datafeeder.publishing.geonetwork.publish-metadata` set to `false` (kept `true` by default), then the metadata is not published by the publication job.

With the current state of the DF code, when a user uploads a dataset, the user responsible for creating a metadata and publish it onto the catalogue is a bot account, making it impossible for the user having pushed the data to modify the metadata being generated from it afterwards. This PR also aims to impersonate the user, by giving the ownership on the metadata being created to the original user having uploaded the dataset.

Unrelated changes:

* updating README to reflect the fact that the docker-composition being used for integration testing is no longer managed by maven.
* updating docker-compose to use latest images, which makes more sense IMHO targeting the master branch.

Tests: mainly runtime, current testsuite adapted to reflect changes made onto the code.

### DIFF
--- a/datafeeder/src/main/java/org/georchestra/datafeeder/config/DataFeederConfigurationProperties.java
+++ b/datafeeder/src/main/java/org/georchestra/datafeeder/config/DataFeederConfigurationProperties.java
@@ -92,6 +92,12 @@ public @Data class DataFeederConfigurationProperties {
         private URI templateTransform;
         private String defaultResourceType = "dataset";
         private boolean publishMetadata = true;
+        // Sets the GeoNetwork group synchronization mode, defaults to "orgs".
+        // should be set to "roles" if a role-based synch is configured GeoNetwork-side.
+        //
+        // See:
+        // https://github.com/georchestra/datadir/blob/master/geonetwork/geonetwork.properties#L25-L32
+        private String syncMode = "orgs";
     }
 
     @Data
--- a/datafeeder/src/main/java/org/georchestra/datafeeder/config/DataFeederConfigurationProperties.java
+++ b/datafeeder/src/main/java/org/georchestra/datafeeder/config/DataFeederConfigurationProperties.java
@@ -91,6 +91,7 @@ public @Data class DataFeederConfigurationProperties {
         private URI templateRecord;
         private URI templateTransform;
         private String defaultResourceType = "dataset";
+        private boolean publishMetadata = true;
     }
 
     @Data
--- a/datafeeder/src/main/java/org/georchestra/datafeeder/service/geonetwork/DefaultGeoNetworkClient.java
+++ b/datafeeder/src/main/java/org/georchestra/datafeeder/service/geonetwork/DefaultGeoNetworkClient.java
@@ -20,6 +20,8 @@ package org.georchestra.datafeeder.service.geonetwork;
 
 import java.io.IOException;
 import java.net.URL;
+import java.util.ArrayList;
+import java.util.Arrays;
 import java.util.List;
 import java.util.Map;
 import java.util.Objects;
@@ -31,13 +33,17 @@ import org.fao.geonet.client.ApiClient;
 import org.fao.geonet.client.ApiException;
 import org.fao.geonet.client.GroupsApi;
 import org.fao.geonet.client.MeApi;
+import org.fao.geonet.client.OperationsApi;
 import org.fao.geonet.client.RecordsApi;
 import org.fao.geonet.client.UsersApi;
 import org.fao.geonet.client.model.Group;
+import org.fao.geonet.client.model.GroupOperations;
 import org.fao.geonet.client.model.InfoReport;
 import org.fao.geonet.client.model.MeResponse;
+import org.fao.geonet.client.model.SharingParameter;
 import org.fao.geonet.client.model.SimpleMetadataProcessingReport;
 import org.fao.geonet.client.model.User;
+import org.fao.geonet.client.model.UserGroup;
 import org.georchestra.datafeeder.model.UserInfo;
 import org.glassfish.jersey.client.ClientProperties;
 import org.springframework.http.HttpHeaders;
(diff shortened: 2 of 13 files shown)


pmauduit:
I left it into draft because I was afraid / concerned not to have added any tests, but looking at the coverage, it stays pretty decent.
Maybe I'll take some time to add one to test with the publsh-md flag set to false though.

pmauduit:
> Maybe I'll take some time to add one to test with the publsh-md flag set to false though.

done, added IT

datafeeder/src/main/java/org/georchestra/datafeeder/service/publish/impl/GeorchestraMetadataPublicationService.java
```diff
@@ -94,10 +97,13 @@ public void publish(@NonNull DatasetUploadState dataset, @NonNull UserInfo user)
         Supplier<String> record = templateMapper.apply(mdProps);
         Organization organization = user.getOrganization();
         // The metadata is inserted into the group which already exists due to
-        // GeoNetwork's LDAP sync (which currently garantees that a geonetwork group
+        // GeoNetwork's LDAP sync (which currently guarantees that a geonetwork group
         // exists for the publishing organization)
+        // TODO what if the sync is configured to synchronize against the roles, and not
```
fvanderbiest, line 102: oops that's exactly the case for CIRAD ...

fvanderbiest:
If I remember correctly, the datafeeder was originally made with the assumption that anyone (including unauthenticated users, given the adequate SP configuration) would be able to submit a dataset.

I understand this is not the case anymore with this change.

Are we all OK with this move ?

datafeeder/src/main/java/org/georchestra/datafeeder/service/publish/impl/GeorchestraMetadataPublicationService.java
pmauduit: I just modified the comment here, so we already are onto the standard behaviour. But I wonder how the DF can work in such a configuration.

pmauduit: This is the default behaviour (without the current PR) when GN is configured to sync with the roles:
Screenshot from 2025-05-05 17-55-55

pmauduit: Testing the current PR with roles-based sync + `publish-metadata` set to `false` give the following rights matrix on a MD generated by the DF:
Screenshot from 2025-05-06 09-40-26

which sounds compliant to me.

f-necas: I think: 
- DF should be aware of orgs/role sync config
- If roles sync is choosed, each group (role) of the user can edit the MD.
At least something like (user is editor in `my_role` and `another_one`):
image

f-necas:
Don't know what's the best way for DF to be aware of orgs/roles synchronization config. Do we duplicate the entry in DF (easiest way) or do we read gn's one (ugly), or do we move it to default.propoerties (huge change for minor thing).

I would say first option.

datafeeder/src/main/java/org/georchestra/datafeeder/config/DataFeederConfigurationProperties.java
```diff
@@ -92,6 +92,12 @@ public static class GeonetworkPublishingConfiguration extends ExternalApiConfigu
         private URI templateTransform;
         private String defaultResourceType = "dataset";
         private boolean publishMetadata = true;
+        // Sets the GeoNetwork group synchronization mode, defaults to "orgs".
+        // should be set to "roles" if a role-based synch is configured GeoNetwork-side.
+        //
+        // See:
+        // https://github.com/georchestra/datadir/blob/master/geonetwork/geonetwork.properties#L25-L32
+        private String syncMode = "orgs";
```
f-necas, line 100: Do we need to set a new value in `datafeeder.properties` ? https://github.com/georchestra/georchestra/blob/master/datafeeder/src/test/resources/datadir/datafeeder/datafeeder.properties#L81

pmauduit: we need to set it to "roles" (or whatever else, actually) if we want to diverge from the default (orgs) one.

pmauduit: You mean, for documentating purposes ?

datafeeder/src/main/java/org/georchestra/datafeeder/service/geonetwork/DefaultGeoNetworkClient.java
```diff
@@ -150,13 +170,65 @@ public GeoNetworkResponse putXmlRecord(@NonNull String metadataId, @NonNull Stri
...
+                if (impersonatedUser.isPresent()) {
+                    User usr = impersonatedUser.get();
+                    List<UserGroup> ugs = usersApi.retrieveUserGroups(usr.getId());
+                    List<GroupOperations> lgo = new ArrayList<>();
+                    ugs.stream().filter(ug -> ug.getGroup().getId() > 2). // skip "hardcoded" GN groups
+                            forEach(ug -> lgo.add(allowEditing(ug.getGroup().getId())));
+                    SharingParameter shareParams = new SharingParameter();
+                    shareParams.clear(false);
+                    shareParams.setPrivileges(lgo);
+                    api.share(metadataId, shareParams);
```
f-necas, line 204: We may need to set recordOwnership there to the first group we found: api.setRecordOwnership(metadataId, groupId.get(), impersonatedUser.get().getId(), true);

f-necas: Appart from that, it works well :)

pmauduit: then the groupName passed as argument to the method will become irrelevant

PR #4460, opened 2025-04-17
GWC - back to a fixed/released version  

### LINKED ISSUE #4459: geowebcache - rely on a stable / released version
Following some work on GWC upstream, we set the version to `1.26-SNAPSHOT`:
https://github.com/georchestra/georchestra/commit/66004be5116913b41e660bb04a66232eed37fcb6

Since a [1.27.0 has been released upstream](https://github.com/GeoWebCache/geowebcache/releases/tag/1.27.0), maybe we could set the version back to something fixed ?

pmauduit:
This should close #4459.

Also includes an unrelated commit to fix a maven warning.

### DIFF
--- a/geowebcache-webapp/pom.xml
+++ b/geowebcache-webapp/pom.xml
@@ -1,4 +1,4 @@
-<?xml version="1.0" encoding="ISO-8859-1"?>
+<?xml version="1.0" encoding="UTF-8"?>
 <project xmlns="http://maven.apache.org/POM/4.0.0" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
   xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 http://maven.apache.org/maven-v4_0_0.xsd">
   <modelVersion>4.0.0</modelVersion>
@@ -17,11 +17,11 @@
     <gwc.unpack.dir>${project.build.directory}/unpacked-webapp</gwc.unpack.dir>
     <geowebcache.war.excludes />
     <maven.test.skip>false</maven.test.skip>
-    <gwc.version>1.26-SNAPSHOT</gwc.version>
+    <gwc.version>1.27.0</gwc.version>
     <!-- override dependencies from parent pom to match the ones for the gwc version -->
-    <gt.version>32-SNAPSHOT</gt.version>
+    <gt.version>33.0</gt.version>
     <spring.version>5.3.39</spring.version>
-    <spring-security.version>5.7.12</spring-security.version>
+    <spring-security.version>5.7.13</spring-security.version>
     <hazelcast.version>5.3.8</hazelcast.version>
     <context.name>geowebcache</context.name>
     <packageDatadirScmVersion>master</packageDatadirScmVersion>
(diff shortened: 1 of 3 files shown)


landryb:
trying to deploy it, ive migrated my setup to the new header so `default.properties` has no value for `headerUrl`, webapp startup blows:
```
Caused by: java.lang.IllegalArgumentException: Could not resolve placeholder 'headerUrl' in value "${headerUrl}"
```
dunno if there should be a default value in https://github.com/georchestra/georchestra/blob/b5bc57824aaf87aca32d7cac04bc19289683c602/geowebcache-webapp/src/main/webapp/WEB-INF/geowebcache-georchestra.xml#L47, not sure this is the case in other webapps.

if i define `headerUrl`, then the webapp fails because georchestra-commons still looks for log4j1 ?
```
Caused by: java.lang.ClassNotFoundException: org.apache.log4j.PropertyConfigurator
    at org.apache.catalina.loader.WebappClassLoaderBase.loadClass(WebappClassLoaderBase.java:1308) ~[tomcat9-catalina-9.0.91+dfsg.jar:9.0.91]
    at org.apache.catalina.loader.WebappClassLoaderBase.loadClass(WebappClassLoaderBase.java:1120) ~[tomcat9-catalina-9.0.91+dfsg.jar:9.0.91]
    at org.georchestra.commons.configuration.GeorchestraConfiguration.<init>(GeorchestraConfiguration.java:75) ~[georchestra-commons-24.1-SNAPSHOT.jar:?]

```

landryb:
after manually readding log4j1 jar (and `apache-log4j-extras-1.1.jar` for `RollingFileAppender`) cf 413b0d1df3679e596e2ecf5308c5adb749675104 the webapp starts fine:
```
Welcome to GeoWebCache version 1.27.0, build 230de11bf247f7e6ac994c7383ba6b4958fddcdf/230de11bf247f7e6ac994c7383ba6b4958fddcdf
```

if we want to remove the log4j-extras dependency we have to adapt https://github.com/georchestra/datadir/blob/master/geowebcache/log4j/log4j.properties

pmauduit:
Looking at the code, I believe that the error arises because of the `georchestraConfiguration` bean defined here:

```
src/main/webapp/WEB-INF/geowebcache-georchestra.xml:  <bean id="georchestraConfiguration" class="org.georchestra.commons.configuration.GeorchestraConfiguration">
```

but apart from loading a log4j1 (which might not make sense depending on which lib the actual webapp relies on ?), it provides a way to return a JS file which was of interest when the extractorapp & mapfishapp webapps were still around:
https://github.com/georchestra/georchestra/blob/master/commons/src/main/java/org/georchestra/commons/configuration/GeorchestraConfiguration.java#L115-L140
maybe we could drop the bean definition completely.

landryb:
> pushed a commit which removes the bean, runtime-tested it does not seem to harm the webapp.

great, this produces a working gwc war/deb (which contains a georchestra-commons jar, expected ?), the only missing thing is some logging config.

If it moved to log4j2, maybe something should be done [here](https://github.com/georchestra/datadir/tree/master/geowebcache) for consistency ? according to https://geowebcache.osgeo.org/docs/current/troubleshooting/index.html#logging we should have a `log4j2.xml` file..

landryb:
noted that https://sourceforge.net/projects/geowebcache/files/geowebcache/1.27.1/ has 1.27.1 if that matters..

f-necas (approved):

PR #4468, opened 2025-05-07
console: display names in logger, delegations and org when availables

f-necas:
Displays correctly names on some pages
Instead of 
image
Now it's 
image


fvanderbiest:
LGTM, tkx !

PR #4494, opened 2025-05-19
datafeeder: record ownership wasn't set if sync is based on roles

f-necas:
`api.setRecordOwnership(metadataId, groupId.get(), impersonatedUser.get().getId(), true);` wasn't set if sync was based on roles because `groupId` was only set based on organization's name whcich doesn't work in roles context.
 
 - Set orgBasedSync check when we set `api.setRecordOwnership`  
 - If roles based sync, override groupId and groupName with first valid (not GN adminstration groups) group where user belongs to.

### DIFF
--- a/datafeeder/docker-compose.yml
+++ b/datafeeder/docker-compose.yml
@@ -82,7 +82,7 @@ services:
       - "28080:8080"
 
   elasticsearch:
-    image: docker.elastic.co/elasticsearch/elasticsearch:7.15.1
+    image: docker.elastic.co/elasticsearch/elasticsearch:7.17.21
     environment:
       discovery.type: single-node
       ES_JAVA_OPTS: "-Xms1g -Xmx1g"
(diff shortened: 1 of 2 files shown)


fvanderbiest:
Can you describe why it failed and what you did to make it work ?
Thanks :-)

f-necas:
mmh test fail but app works as expected ...

fvanderbiest:
OK, I understand that for CIRAD, as for many SDIs who chose to synchronise geOrchestra roles with GeoNetwork groups (this is not the default setting in geOrchestra, see https://github.com/georchestra/datadir/blob/17e750efa07ec9a66e10a93d3e5cee475336dc1f/geonetwork/geonetwork.properties#L25-L32), the key point is that the datafeeder-created metadata now belongs effectively to the user + a random GeoNetwork group he belongs to (as there is no way to decide which one is the most appropriate one).

Based on this, I have one additional question : is this user allowed to change the group manually afterwards ?

Issue #4345, opened 2024-09-22
console - erroneous title


fvanderbiest:
![image](https://github.com/user-attachments/assets/2ec68358-3775-4ff8-af01-014cf30ed994)
"Consentement" title should be renamed into "Collecte des données personnelles" in conformity with the GDPR.

PR #4500, opened 2025-06-04
console: password hints display in newPassword page

f-necas:
Also address issue #4345 

console wasn't disaplying password hints in newPassword page :
image

Password policy is not correctly displayed:
image

### DIFF
--- a/console/src/main/java/org/georchestra/console/ws/passwordrecovery/NewPasswordFormController.java
+++ b/console/src/main/java/org/georchestra/console/ws/passwordrecovery/NewPasswordFormController.java
@@ -96,6 +96,7 @@ public class NewPasswordFormController {
             formBean.setUid(uid);
 
             model.addAttribute(formBean);
+            model.addAttribute("pwdUtils", passwordUtils);
 
             return "newPasswordForm";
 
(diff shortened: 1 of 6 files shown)


console/src/main/java/org/georchestra/console/ws/passwordrecovery/NewPasswordFormController.java
```diff
@@ -96,6 +96,7 @@ public String setupForm(@RequestParam(required = false) String token, Model mode
             formBean.setUid(uid);
 
             model.addAttribute(formBean);
+            model.addAttribute("pwdUtils", passwordUtils);
```
pmauduit, line 99: what is ther purpose of adding this object to the model, if not used in the views afterwards ?

pmauduit (commented):
calling myself out for the german, spanish and nederland-ish translations :-) apart from that, I don't get the modifications in the Java code (see comment).

console/src/main/java/org/georchestra/console/ws/passwordrecovery/NewPasswordFormController.java
pmauduit: ok got it, it's already being referred in the view, but since the object was not there, the display was broken.

PR #4532, opened 2025-07-09
chore: ldap - removes the geoserver_privileged_user user (fixes #1073)

### LINKED ISSUE #1073: Avoid populating geoserver_privileged_user by default
For server side communications, we're sometimes using basic auth with a geoserver privileged user (eg: extractorapp). This is sub-optimal IMO.

Instead, could we reuse the user session (security-proxy cookie) that we would allow to go along with the proxied request ?

pmauduit:
The title is self-explanatory: see #1073 for the context.

Since the user is "protected", it won't show up in the console, and if not configured explicitely to use another password, will come with a default one, allowing people to connect to the platform with an `ADMINISTRATOR` account.

It seems that this user is no longer used, as geofence is not an external module anymore and integrated as an option to geoserver, and extractorapp also disappeared some years ago.

There are still some occurences of the user in the codebase:
1. still some references in the testsuite
2. It has been kept into the analytics configuration, as it sounds relevant to still filter it out from this module.

### DIFF
--- a/console/README.md
+++ b/console/README.md
@@ -140,12 +140,9 @@ Once created, set the following parameters:
 
 ### Protected Users
 
-Several user accounts can be protected against deletion or modification, with the `protectedUsersList` property, which holds a comma separated list of user accounts `uid`. These users also do not show up in the manager.
-
-By default, only `geoserver_privileged_user` is protected:
-```
-protectedUsersList=geoserver_privileged_user
-```
+Several user accounts can be protected against deletion or modification, with
+the `protectedUsersList` property, which holds a comma separated list of user
+accounts `uid`. These users also do not show up in the console.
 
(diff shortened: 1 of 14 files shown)


landryb:
wasnt it used in mapfishprint/mfapp (and maybe mapstore) to print 'protected' layers ?

pmauduit:
> and all [the places](https://github.com/search?q=repo%3Ageorchestra%2Fdatadir%20geoserver_privileged_user&type=code) that reference a 'privileged user'
 
I think it makes sense to keep the occurence in analytics (the old module), to ensure retrocompatibility (else the risk is to display the user in the charts/tables from the webapp, which won't harm though, but still).

f-necas (approved):

PR #4537, opened 2025-07-28
Update .gitmodules to update branches for 25.0

f-necas:
Create two different branches to avoid conflicts with non submodules builds

### DIFF
--- a/.gitmodules
+++ b/.gitmodules
@@ -1,8 +1,8 @@
 [submodule "geonetwork"]
 	path = geonetwork
 	url = https://github.com/georchestra/geonetwork.git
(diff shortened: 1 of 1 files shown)


landryb (approved):
i dont know if it matters much, but that triggered my OCD.. i'd rather have consistency with the existing branches on [geoserver](https://github.com/georchestra/geoserver/branches) and [geonetwork](https://github.com/georchestra/geonetwork/branches/), eg maybe `georchestra-gn4.4.8-25.0` and `2.27.1-georchestra-25.0` ? (eventually dropping the `georchestra` bit maybe..)

PR #4538, opened 2025-07-28
geoserver: remove geoserver from core

f-necas:


Makefile
```diff
@@ -91,7 +55,7 @@ deb-build-georchestra: war-build-georchestra build-deps deb-build-geoserver deb-
 # Base geOrchestra common modules
 build-deps:
 	mvn -Dmaven.test.failure.ignore clean install --non-recursive
-	mvn clean install -pl commons,ogc-server-statistics -Dmaven.javadoc.failOnError=false
+	mvn clean install -pl commons,ogc-server-statistics,security-proxy-spring-integration -Dmaven.javadoc.failOnError=false
```
landryb, line 58: is that change related to the PR ?

landryb (approved):
i suppose to merge along/altogether with georchestra/geoserver#41 ?

Makefile
f-necas: Not really but it was useful for another component, don't remember which one. Do you prefere that I remove it ?

landryb: those 'hidden deps' between apps are annoying, would be nice to know clearly which one needs which other module :) im definitely +1 on adding it, just wanted to make sure we knew 'why' ..

landryb:
i note that this leads to .. funnily named tags like https://github.com/georchestra/geoserver/releases/tag/2.27.1-georchestra-01 to push releases, which for my understanding are required by docker/k8s ?

but then, that tag was made on the `main` branch, which in the georchestra repo is based off `2.27.1-georchestra`. At that point, i'm completely lost in the branching process that is supposed to be used.

the upstream repo also uses `main`, but for their **main development** branch......

PR #4556, opened 2025-08-05
ogcss: remove ogc-server-stats from core

f-necas:


security-proxy/pom.xml
```diff
@@ -112,9 +112,9 @@
       <artifactId>spring-security-web</artifactId>
     </dependency>
     <dependency>
-      <groupId>${project.groupId}</groupId>
+      <groupId>org.georchestra</groupId>
       <artifactId>ogc-server-statistics</artifactId>
-      <version>${project.version}</version>
+      <version>//TODO SET VERSION WHEN OTHER PR IS MERGED AND JAR IS BUILT</version>
```
landryb, line 117: IM NOT SURE WHICH PR THIS REFERS TO BUT WHY ARE YOU YELLING ?

f-necas: For analytics I saw it in GHA https://github.com/georchestra/analytics-legacy/blob/main/.github/workflows/analytics.yml#L74-L80
For GWC, I don't know how it was done. 
For ogcss, I can add a github actions to push it ? (same fow gwc by the way)

PR #4563, opened 2025-08-12
After cleaning some repos from core, also clean dependencies

f-necas:
Cleans maven dependencies to use only necessary. 

Runtime tested console and security proxy. Seems to work fine

### DIFF
--- a/.github/workflows/console.yml
+++ b/.github/workflows/console.yml
@@ -15,6 +15,7 @@ on:
       - "commons/**"
       - "ldap-account-management/**"
       - "console/**"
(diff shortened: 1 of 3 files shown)


landryb:
out of curiosity, what rule/tool was used to drop those ?

f-necas:
Remove all dependencies, check which ones maven was arguing at. Readd them

landryb:
> Remove all dependencies, check which ones maven was arguing at. Readd them

hah, human intervention, i like that :)

landryb (approved):

PR #4616, opened 2025-12-18
feat: allow colon in roles for ldap and console

f-necas:
Required for PR in geonetwork https://github.com/georchestra/geonetwork/pull/350


pmauduit (approved):
I fear that some in the georchestra community would have been making use of ":" in their role names, but I think it'd deserve a mention in the doc somewhere. Maybe we don't even have something in the doc mentioning illegal characters in roles names ?

PR #4645, opened 2026-04-01
feat: allow to compile with 21 but use 17 runtime

f-necas:


landryb (approved):
we'll know shortly how it works on the demo :)

PR #4653, opened 2026-04-03
Clean POMs : use logback, remove RPM and rm docker datadir checkout

### LINKED ISSUE #2465: docker images - revisit the need to ship the datadir in images
When using the docker maven profile, a fraction of the datadir corresponding to the current webapp is copied to /etc/georchestra

This is a pain to maintain, and we never use the config that is shipped inside the image, since we usually bootstrap a full datadir from georchestra/datadir.

As a result I would be +1 on having images that do not ship the config datadir.

f-necas:
This PR does three things:

- Use logback instead of log4j and adapt loading of file in datadir. (embedded in pom with spring parent)
- Remove RPM package which was an old test.
- Remove checkout of datadir from docker-master bracnh in docker context only.


landryb:
shouldnt there be a PR for the datadir repo, providing a default sane `console/logback.xml` ?

f-necas:
> shouldnt there be a PR for the datadir repo, providing a default sane `console/logback.xml` ?

Indeed https://github.com/georchestra/datadir/pull/492 ! Thanks @landryb

commons/src/main/java/org/georchestra/commons/configuration/GeorchestraConfiguration.java
```diff
@@ -66,16 +72,19 @@ public GeorchestraConfiguration(String context) {
...
-            File log4jXml = new File(contextDataDir, "log4j" + File.separator + "log4j.xml");
-            String log4jConfigurationFile = null;
-            if (log4jProperties.exists()) {
-                log4jConfigurationFile = log4jProperties.getAbsolutePath();
-                PropertyConfigurator.configure(log4jConfigurationFile);
-            } else if (log4jXml.exists()) {
-                log4jConfigurationFile = log4jXml.getAbsolutePath();
-                DOMConfigurator.configure(log4jConfigurationFile);
+            // logback configuration
+            File logbackXml = new File(contextDataDir, "logback" + File.separator + "logback.xml");
```
pmauduit: something more "modern"-Java comes to my mind:
```
jshell> java.nio.file.Paths.get("/etc/georchestra/or/whatever", "gateway", "logback", "logback.xml").toFile()
$3 ==> /etc/georchestra/or/whatever/gateway/logback/logback.xml
```

pmauduit (approved):
Just a remark on the logback configuration file which can be calculated using the java.nio api instead of concatenating strings, but apart from that, looks good

PR #4673, opened 2026-06-01
fix: rework LDAP startup logic

pmauduit:
The `/var/run/slapd` directory is not provided by the package, I am not sure it was the case before, but we have to ensure the directory exists before trying to launch slapd. moving the mkdir -p/chown logic into the early beginning of the 00-init script, and executing regardless of whether the LDAP has already been initialized or not.

Tests: directly onto a test infra (with an existing setup), and running the docker image by hand locally to make sure the bootstrap process has not been broken by the change.


f-necas (approved):
Not tested but LGTM

